PR #53, opened 2018-09-01
Update Step2_ConvertingLabels2DataFrame.py

dilawar:
Fixes for Python3.4. It fails because python3.4 raises `OSError` and not `FileNotFoundError`. `FileNotFoundError` is a subclass of `OSError`. Relevant https://stackoverflow.com/a/28633573/1805129 

Probably a good idea to replace all instance of `FileNotFoundError` with `OSError`.


AlexEMG:
The FileNotFoundError exception is actually used at various spots throughout the repo. Did you check for comparability with p3.4 elsewhere?

dilawar:
NO. In my fork, I've replaced all of them by OSError. So I can't be sure. In any case, FileNotFoundError is subclass of OSError so it would not break the code.

⁣Sent from handheld device. Pardon the typos and brevity.

PR #94, opened 2018-11-06
Close video reader after selecting frames

WernTanj:
On Windows, a `clip.reader.close()` is necessary to avoid errors such as `OSError: [WinError 6] The handle is invalid`.

See https://stackoverflow.com/a/45393619

### DIFF
--- a/Generating_a_Training_Set/Step1_SelectRandomFrames_fromVideos.py
+++ b/Generating_a_Training_Set/Step1_SelectRandomFrames_fromVideos.py
@@ -113,6 +113,7 @@ def SelectFrames(videopath,filename,x1,x2,y1,y2,cropping, videotype,start,stop,T
                     io.imsave(os.path.join(basefolder,folder,"img" + str(index).zfill(indexlength) + ".png"),image)
                 except FileNotFoundError:
                     print("Frame # ", index, " does not exist.")
+            clip.reader.close()
 
 if __name__ == "__main__":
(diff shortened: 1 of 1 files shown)


AlexEMG:
Hi @wernle ,

Thanks for raising this issue. I had come across a similar problem (that is not an issue in Linux), and pushed it yesterday for the Analysis-tools/AnalyzeVideos.py script. There, I added 

 clip.close()
 del clip

based on this issue of moviepy: https://github.com/Zulko/moviepy/issues/518 (can slow down on Windows ...)

Anyway, I was wondering if for your patch, clip.close() would not be better, as the code is not using the clip.reader instance (which does not work for cropping), but rather clipresized.get_frame(index * 1. / clipresized.fps)?

Best,
Alex

AlexEMG:
I noticed that both in Step1 and frameselection tools there were open clips. Closed those in the latest push. Thank you very much for raising this. Please, let me know if it works for you!

WernTanj:
Hi Alex,

Thank you for the fast reply. I’ll check if the updated Step1 works asap next week and will try the new UI as well!

All the best,
Tanja

On 6 Nov 2018, at 17:24, Alexander Mathis <notifications@github.com<mailto:notifications@github.com>> wrote:


I noticed that both in Step1 and frameselection tools there were open clips. Closed those in the latest push. Thank you very much for raising this. Please, let me know if it works for you!

PR #102, opened 2018-11-11
Rotation and lighting (gamma) data augmentation

lambdaloop:
This is some prototype code for rotation and lighting (gamma) data augmentation. 
This has been tested and works well in the previous deeplabcut version (v1.11) but as yet untested in v2.0 . Including it here in case someone is interested and could test it earlier than me.

### DIFF
--- a/conda-environments/dlc-windowsGPU.yaml
+++ b/conda-environments/dlc-windowsGPU.yaml
@@ -1,7 +1,7 @@
 # dlc-windowsGPU.yaml 
 # 
 # DeepLabCut environment for use on FMI computers
-# FIRST: INSTALL CORRECT DRIVER for GPU, see https://github.com/AlexEMG/DeepLabCut/blob/master/docs/images/cuda_driver.png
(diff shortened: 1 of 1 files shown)


lambdaloop:
In order to actually activate this, `pose_cfg.yaml` needs to be edited in order to add rotation and gamma augmentation parameters (default is no augmentation, to mimic previous behavior). 
Here are some example options which I have been using:
```yaml
rotation_jitter_lo: -100
rotation_jitter_up: 100
gamma_jitter_lo: 0.85
gamma_jitter_up: 1.15
```

AlexEMG:
Just as a note to everybody else: I discussed this with Pierre, it will be integrated in a few days.

MMathisLab:
perhaps update with your benchmarking on this, @AlexEMG Alex?

lambdaloop:
Oh I see what happened! I updated the master branch in my copy of the repo. 

If you'd still like to see the changes, they are in the data-aug branch:
https://github.com/lambdaloop/DeepLabCut/tree/data-aug

MMathisLab:
(also, sorry this was not in 2.0.4, as we think there is an error, but will discuss offline with you!)

PR #114, opened 2018-11-14
fix mpl rectangle bbox

arnefmeyer:
Use width/height instead of absolute coordinates when creating the MPL cropping preview rectangle


AlexEMG:
What is the advantage of this?

arnefmeyer:
This is what the constructor expects:
https://matplotlib.org/api/_as_gen/matplotlib.patches.Rectangle.html

The current master branch is using the corner coordinates instead of width/height.

PR #117, opened 2018-11-15
check_labels: Split windows file paths correctly

imagejan:
The `MakeLabeledPlots` method contained a line that assumed Unix file separators (`"/"`) to split paths. This commit fixes the issue by using `os.sep`.

/cc @WernTanj


AlexEMG (approved):
Thanks!

PR #153, opened 2018-12-18
Fixed some typos in docs

ghost:
I also wanted to fix the typo in 

> max_snapshots_to_keep: int, or None. Sets how many snapshots are kept, i.e. states of the trained network. Every **savinginteration** many times a snapshot is stored, however only the last max_snapshots_to_keep many are kept! If you change this to None, then all are kept.

found in ``docs/functionDetails.md`` and ``deeplabcut/pose_estimation_tensorflow/training.py``, but I'm not sure what "savinginteration" is supposed to be exactly.

### DIFF
--- a/docs/functionDetails.md
+++ b/docs/functionDetails.md
@@ -10,7 +10,7 @@ Optional arguments specify the working directory, where the project directory wi
 
            deeplabcut.create_project(`Name of the project',`Name of the experimenter', [`Full path of video 1',`Full path of video2',`Full path of video3'], working_directory=`Full path of the working directory',copy_videos=True/False) 
           
- (TIP: you can also place ``config_path`` in front of ``deeplabcut.create_project`` to create a vriable that holds the path to the config.yaml file, i.e. ``config_path=deeplabcut.create_project(...)``)
+ (TIP: you can also place ``config_path`` in front of ``deeplabcut.create_project`` to create a variable that holds the path to the config.yaml file, i.e. ``config_path=deeplabcut.create_project(...)``)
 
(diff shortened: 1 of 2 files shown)


AlexEMG:
_savinginteration_ is a variable that you can set in the pose_config file. For the default values and description see: https://github.com/AlexEMG/DeepLabCut/blob/master/deeplabcut/pose_cfg.yaml

ghost:
> _savinginteration_ is a variable that you can set in the pose_config file. For the default values and description see: https://github.com/AlexEMG/DeepLabCut/blob/master/deeplabcut/pose_cfg.yaml

Besides ``docs/functionDetails.md`` and ``deeplabcut/pose_estimation_tensorflow/training.py``, I can't find the string "savinginteration" using silver searcher anywhere in the repository or the package after being installed by pip,  let alone in pose_cfg.yaml. Perhaps it used to be there but was removed for some reason?

MMathisLab (approved):

MMathisLab (approved):

AlexEMG:
Yes, sorry: ```save_iters``` is the name in the pose_config files -- as stated: 

CRITICAL POINT: It is recommended to train for thousands of iterations until the loss plateaus (typically around 200,000). The variables display_iters and save_iters in the pose_cfg.yaml file allows the user to alter how often the loss is displayed and how often the weights are stored.

saveiters: this variable is actually set in pose_config.yaml. However, you can overwrite it with this hack. Don't use this regularly, just if you are too lazy to dig out 
the pose_config.yaml file for the corresponding project. If None, the value from there is used, otherwise it is overwritten! Default: None

https://github.com/AlexEMG/DeepLabCut/blob/9d9cf5b922dc993da3aad6e9a4693e2173b1e7ae/docs/functionDetails.md

PR #162, opened 2019-01-03
Fix docstrings

ghost:
This fixes a lot of docstrings to keep them clean, consistent, and comprehensive.

It does not completely fix the docstrings of cli.py, and it does not touch the docs themselves.


AlexEMG:
Thanks for your input! I will look at this, once I have some idle time.

AlexEMG (dismissed):

MMathisLab (approved):

MMathisLab:
okay, I am fine with the changes now. I don't love "str" as it is not clear to novice programmers that it means string.

ghost:
That's true, although I like `str` for accuracy.

ghost:
Perhaps we could keep `str` for docstrings to stay consistent with python docstring conventions (i.e. numpy, Google, reST, etc.) which would keep them compatible with documentation generation programs (e.g. Sphinx) and other tools in the python ecosystem. Meanwhile the user docs could be more novice friendly. I'm thinking that a novice that does not know what `str` is probably does not know about docstrings and how to view them either.

MMathisLab:
I think actually a lot of users check the docstrings, as we sugest that for every function in DLC if they type `?` afterwards, i.e. `deeplabcut.create_new_project?` they can get help :)

ghost:
Ah, fair point.

ghost:
What about cli.py? It still has `str`, was that intentional?

ghost:
Can we please get this merged before further conflicts arise?

MMathisLab (commented):
added back information under analyzie_videos

MMathisLab:
sorry, @UnicodeAlt255 - we had so many function changes, but I can work through these merge issues... I think 2.0.4 is our best package yet though!

PR #168, opened 2019-01-10
Use current version of matplotlib

ghost:
This allows the use of current versions of matplotlib rather than being stuck on 2.2.2.
I did not do extensive testing but it seems to not have changed anything.


AlexEMG:
Changed in 2.0.5. Thanks!

PR #174, opened 2019-01-15
Update make_labeled_video.py

Mot0ka64:
Speed up CreateVideo function using temporary variables. (about x50 faster in my environment)


AlexEMG:
Great & simple idea! Thanks for pushing this; I tested it and got similar improvements. Switched the code to numpy (which increased the speed a bit more) and also added it for the slow video creation code.

PR #215, opened 2019-02-18
Fixes bug in extract_outlier_algorithms()

michaelfsp:
Fixes bug where extract_outlier_algorithms() fails to compute filtered timeseries when outlieralgorithm='fitting' and comparisonbodypart='all'


AlexEMG:
Thanks!

PR #221, opened 2019-02-24
small GUI updates

jbonaiuto:
maintain field of view when moving to next or previous frame
only call self.figure.canvas.draw() when necessary
convert from openCV BGR to matplotlib RGB


MMathisLab (approved):

MMathisLab:
Hi @jbonaiuto we are happy to accept the PR related to the GUI, but the other edits we need more time on; can you make a PR with just the GUI edits?

PR #245, opened 2019-03-18
Provide ability to test on videos that weren't trained on

katierupp:
Added the option to test on a different dataset than the dataset that was trained on.  Utilized the Shuffles parameter to perform leave-one-out cross-validation on the folders within the training dataset. Added the suffix to allow multiple training datasets to be created within the same project folder. The parameters are set to the defaults to be backwards compatible.

### DIFF
--- a/deeplabcut/generate_training_dataset/trainingsetmanipulation.py
+++ b/deeplabcut/generate_training_dataset/trainingsetmanipulation.py
@@ -306,6 +306,34 @@ def SplitTrials(trialindex, trainFraction=0.8):
         trainIndexes = shuffle[:trainsetsize]
         return (trainIndexes, testIndexes)
 
+def divideTestAndTrain(divide_folders_lists, shuffle, fileNames): 
+
+    trainFiles = []
+    testFiles = []
+    for ii in range(len(divide_folders_lists)): 
+        if ii == shuffle:
+            testFiles.extend(divide_folders_lists[ii])
+        else: 
+            trainFiles.extend(divide_folders_lists[ii])
+    
+    trainIndexes = []
+    testIndexes = []      
+    for ii in range(len(fileNames)): 
+        path = fileNames[ii]
+        filePath = os.path.split(path)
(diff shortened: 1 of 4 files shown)


Idavr:
I am still learning DLC, but I was under the impression that after training a network on a dataset, you were supposed to be able to use that network for future analysis on other videos without having to retrain the network for each new video? Have I misunderstood?

AlexEMG:
@katierupp thank you for contributing this! This is quite useful, and I just pushed a version that we had used internally. I hope it allows all the capabilities that you want. While the suffix variable is a great idea, I also think it adds a bit complexity. In the push one can use just different "shuffles". Hope that makes sense. 

**Let me give some brief examples:** to create a leave-one-folder-out model:
``trainIndexes, testIndexes=deeplabcut.mergeandsplit(config,trainindex=0,uniform=False)``
returns the indices for the first video folder (as defined in config file) as testIndexes and all others as trainIndexes. You can then create the training set by calling (e.g. defining it as Shuffle 3):
``deeplabcut.create_training_dataset(config,Shuffles=3],trainIndexes=trainIndexes,testIndexes=testIndexes)``
    
To freeze a (uniform) split:
``trainIndexes, testIndexes=deeplabcut.mergeandsplit(config,trainindex=0,uniform=True)``
You can then create two model instances that have the identical trainingset. Thereby you can assess the role of various parameters on the performance of DLC.
    
``deeplabcut.create_training_dataset(config,Shuffles=[0],trainIndexes=trainIndexes,testIndexes=testIndexes)
deeplabcut.create_training_dataset(config,Shuffles=[1],trainIndexes=trainIndexes,testIndexes=testIndexes)``

AlexEMG:
@Idavr you can certainly use any trained model on other videos. If trained with enough diversity it should generalize well to other individuals etc. This is also our recommended use case. 

This contribution allows one to test whether a model generalizes to unseen data (video).

Idavr:
@AlexEMG Thank you for specifying!

PR #250, opened 2019-03-27
small GUI changes

jbonaiuto:
just GUI changes this time :) I think I got it right, now:

maintain field of view when moving to next or previous frame
only call self.figure.canvas.draw() when necessary (tested on windows)
convert from openCV BGR to matplotlib RGB in a single line (rather than reading and then converting)


MMathisLab:
sorry for the delay in testing, but I finally had some time. Seems there are some issues. (1) I can no longer delete a point if I accidentally place it, (2) the zoom is still engaged as before, but your PR mentions it automatically turns off? (3) and it is quite laborious now if I want to zoom on another frame differently, I have to click twice, i.e. home, then zoom again. It would be better if it did not automatically stay zoomed. If that is the case, one should crop the input images! :) (4) If you click on the image you get a "ghost" of the zoom box now, such that you get stuck zooming... Here is a video of the errors: https://youtu.be/wLres2yT_8E

jbonaiuto:
Ah sorry - the functionality to automatically turn off zoom got wiped out by my last merge. I've put it back in and tested point deletion and zoom ghosting. Seems to work for me, but let me know if there are still problems. For our images it was more of a pain to have to zoom again after clicking next, but I can see how for other images it would be more effort to click home and then zoom again. I've added a "lock view" checkbox that turns this functionality on and off - that should accommodate both usage patterns.

MMathisLab:
this is great!

MMathisLab:
Hi @jbonaiuto seems this doesn’t work in pythonw on macOS actually. The zoom no longer works and then crashes the GUI- so you have the ability to test/fix? @meet10may could you also check this? I don’t want to put on pypi as is!

MMathisLab:
Hi @jbonaiuto we mad some fixes but couldn't keep the zoom off; the new version will be up soon, so perhaps you can give us your expertise in this - seems to not be stable in MacOS :(

jbonaiuto:
Sorry for the late reply. I've finally tracked down a mac that I can test
on, so I should be able to tackle this by the end of the week.

PR #280, opened 2019-05-05
More robust downloading pre-trained models

jchutrue:
Eliminate need to call download script, and (hopefully) never have to worry again about correct permissions

### DIFF
--- a/deeplabcut/generate_training_dataset/trainingsetmanipulation.py
+++ b/deeplabcut/generate_training_dataset/trainingsetmanipulation.py
@@ -527,23 +527,23 @@ def create_training_dataset(config,num_shuffles=1,Shuffles=None,windows2linux=Fa
     Data = Data[scorer] #extract labeled data
 
     #set model type. we will allow more in the future.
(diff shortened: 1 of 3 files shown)


AlexEMG:
Thanks! I will have a look at this soon (currently occupied). Also see https://github.com/AlexEMG/DeepLabCut/issues/288

MMathisLab:
(we haven't forgotten about this, just have been traveling). However, it could be useful to think about additional models, such as this (below), which can be used as well:
```
curl -L -O https://datasets.d2.mpi-inf.mpg.de/deepercut-models-tensorflow/mpii-single-resnet-101.data-00000-of-00001 
   curl -L -O https://datasets.d2.mpi-inf.mpg.de/deepercut-models-tensorflow/mpii-single-resnet-101.meta
   curl -L -O https://datasets.d2.mpi-inf.mpg.de/deepercut-models-tensorflow/mpii-single-resnet-101.index
```

AlexEMG:
Updated it a bit by turning it into functions, tested on Ubuntu + windows. Thanks @jchutrue !

Issue #285, opened 2019-05-10
path should not be PosixPath


afunTW:
**Your Operating system and DeepLabCut version**

Ubuntu 16.04 LTS, DeepLabCut2.0.6.2

**If applicable, please also complete the following information:**

Work on server
 
**Describe the bug**
https://github.com/AlexEMG/DeepLabCut/blob/master/deeplabcut/utils/auxiliaryfunctions.py#L73
Declare the path by `pathlib.Path()`, but still use the `os.path.exists()`

Recommend to use `path.exists()`

**To Reproduce**

1. Create project
2. extract frames

MMathisLab:
could you give us the error statement that you see from this line, thanks!

afunTW:
At first, I use the `pipenv` with python 3.5 and the issue occurred when I invoke the function `deeplabcut.extract_frame`.

Then I modified the code, and reinstall the module with `python3 setup.py install`
```
# path = Path(configname)
path = str(Path(configname).resolve())
```

However, the other issue #105  triggered, so I upgrade the python version to 3.6 and all the problem are gone included this issue.

So, the issue is available to close now.
But I still recommend to unified the style, either `os.path` or `pathlib.Path`

Related PR: https://github.com/AlexEMG/DeepLabCut/pull/286

MMathisLab:
Awe yes, we required python 3.6. Thanks for clarification!

PR #286, opened 2019-05-10
modified: fixed the bug of PosixPath

afunTW:
Solve the Issue #285


AlexEMG:
This was resolved (relevant for python 3.5). DLC should be run with python 3.6.

PR #293, opened 2019-05-15
fix auto-zoom off for Mac

jbonaiuto:
So it seems that on mac, matplotlib likes to throw extra, unecessary xlim_changed and ylim_changed events. The onZoom function was registered to listen to these events so this was causing zoom to turn off prematurely. I now save the prezoom xlim and ylim values and when onZoom is called, compare the new values with the saved ones to determine if the zoom really has been completed. Tested on Linux, Windows, and Mac :)


MMathisLab:
tested!

darshanmehta17:
I'm still facing the auto-zoom off issue on my Mac. I noticed that this was particularly happening when I had a marker on the border of the canvas.

MMathisLab:
@jeylau can you check this out. 

from gitter: 
> 
@MMathisLab I'm still facing the auto-zoom off issue on my Mac. I tried to compile the project again with some debug lines and noticed that the fix pushed in https://github.com/AlexEMG/DeepLabCut/pull/293 isn't really helping. The updateZoomPan function is being called by the onZoom function prematurely. I also noticed that this was happening more often when I'm zooming around the boundaries of the image canvas. Do you know of any fix for this issue?

PR #309, opened 2019-06-04
Update make_labeled_video.py for cropping

AmandineCornil:
Instead of cropping video, use raw images and translate cropped coordinates of labels to raw coordinates.


AlexEMG:
Thanks for the PR. I added the possibility to also plot it on the cropped video, i.e.:

Now the user can crop during analysis and then:
- by default ```deeplabcut.create_labeled_video``` will plot the detected bodyparts in the full-sized frame at the correct location (by shifting them)
- plot them on the cropped video itself. This can be done with:
```deeplabcut.create_labeled_video(path_config_file,[video],displaycropped=True)```

PR #318, opened 2019-06-10
Limb hotkey

tvajtay:


tvajtay:
This can be closed I guess, @sebo313 has a better solution.

MMathisLab:
I think I can accept both PRs actually.... let me test across platforms (ubuntu, mac, windows!)

PR #319, opened 2019-06-10
GUI hotkeys UP/DOWN

sebo361:
In order to switch between labels more quickly, i added keyboard UP and DOWN buttons. DOWN arrow advances to the next label in the list and UP arrow goes back to the previous label. Hope it helps :)


MMathisLab:
works really well! Only minor issue is that when you arrow down all the way to the end of the bodypart list, you can't go back up - otherwise UP/DOWN works! Also, when one clicks on the buttons, future up/down is disabled.

PR #321, opened 2019-06-12
Add option to return multiple maximums per bodypart

lambdaloop:
This pull request adds an (optional) option for number of outputs per bodypart.
That is, instead of returning a single maximum per frame, this allows for returning the top N maxes.
This is useful for further processing to pick the best points based on temporal or geometrical constraints at a later stage in the pipeline.

All the code is vectorized, and falls back onto the current functions in the case of num_outputs=1 .
Furthermore, I found that the new code runs at about the same speed as before, even in the case of num_outputs=20 . It slows down a bit as you increase num_outputs, but still runs at 85% of the speed of num_outputs=1.


MMathisLab:
Not sure I understand this, as you can have an arbitrary batch size already, and it's set in the config file. This seems a revision to our previous code?

image

MMathisLab:
Oh, I see you un-did that part of the PR.

lambdaloop:
Yes, sorry for the confusion. I edited that part when I was testing and forgot to undo it in the initial commit. (I do think that commit would be useful to allow arbitrary batch sizes when creating a model with `setup_pose_prediction`, but that's a conversation for a different pull request.)

MMathisLab:
sure; the reason batch size is 1 for models is that users can have different sized images that way ;)

lambdaloop:
I've updated the code to be compatible with DeepLabCut v2.0.7, let me know if there are any issues.

AlexEMG:
Finally testing this...

lambdaloop:
Hey @AlexEMG, any updates on this?

AlexEMG:
It works ok, however, it is not integrated with anything else in the package so that users can use the multiple peaks. Do you have anything in mind?

lambdaloop:
It does output into the hdf/csv files so that users may integrate this into their pipelines.

I am working on using these to recover from bad tracking in the case of multiple cameras, within anipose, by distinguishing which point is correct.

I think it could also be used for users to prototype algorithms for multiple animal tracking. This feature makes it much easier to apply the "extract local peaks" heuristic you mentioned before.

AlexEMG:
I understand that - and I think it works well! I was just wondering if it would make more sense to have it more integrated in DLC?

lambdaloop:
Ah I see, what are you thinking in terms of integration?

AlexEMG:
Perhaps at least displaying in the video? And processing such that spatially neighboring predictions across time are linked in column?

lambdaloop:
I agree this would be a nice addition, but I don't have time to do this at this time. 

I think the functionality would be already be useful without more integration though. Perhaps other users could add it as needed?

AlexEMG:
Fair enough.

bcb225:
Hi Alex, thank you for keep updating your project.

In our lab we uses DLC for detecting mouse and it works very well.

Is there any updates on labeling and detecting multiple animal?

I read some issues that DLC will get that feature in near future.

Let me know if there will be an update and when it will be available to us.

Thank you again for your commitment.

lambdaloop:
Hello again, I've pulled the latest updates of DeepLabCut, so that there are no merge conflicts.

AlexEMG:
Finally integrated and released! Thanks for your contribution.

PR #324, opened 2019-06-15
Option for deterministic training + ResNet-152 added

sebo361:
Added option for running 100% deterministic training on CPU. When training on a GPU, cuDNN still brings in stochastic values, but the standard deviation of losses is really low as shown in my quick evaluation below :)

Set "deterministic: True/False" (default is False) in pose_cfg.yaml

CPUvsGPU_loss


sebo361:
For indexing more than one folder in mergeandsplit() I added the option test_video_name=[]. Now you can list video folder names in order to return their indices as testIndexes and all others as trainIndexes.

Examples
    --------
    To create a leave-folder-out model:
    >>> trainIndexes, testIndexes=deeplabcut.mergeandsplit(config ['mouse1','mouse3'],uniform=False)
    returns the indices for the video 'mouse1' and 'mouse3' as testIndexes and all others as trainIndexes.
    You can then create the training set by calling (e.g. defining it as Shuffle 3):
    >>> deeplabcut.create_training_dataset(config,Shuffles=[3], trainIndexes=trainIndexes, testIndexes=testIndexes)

sebo361:
Hi @AlexEMG, thank you so much for helping to implement this option for deterministic training and even adding a test script! In terms of reproducibility that's an awesome option. I have trained the same data set (with same config settings) on the GPU to compare deterministic vs. non-deterministic performance. When training deterministically, train and test pixel error is always much higher as you can see in the chart. So the steps of randomly shuffling and re-scaling images have quite a huge impact on performance :)

image

AnneleenDewulf:
Deterministic training does not seem to work for me. I froze a uniform split via the code provided by @sebo361 (but with uniform = True) and I have put deterministic = True in the pose_cfg.yaml. Yet, the creation of my training set is deterministic while the training of the network remains non-deterministic. Any suggestions? I am using the deeplabcut 2.3.0.

PR #340, opened 2019-06-27
WIP: Add conda dev environement, pinning only versions required by deeplabcut

gdevenyi:
Add a conda "dev" environment which only pins the versions of packages specified in the setup.py of deeplabcut.

In addition, migrate almost all pip packages to conda based packages.


gdevenyi:
These could also now be renamed as they're by-no-means ubuntu restricted. This is the beauty of using conda packages everywhere.

MMathisLab:
closed to work off #341

PR #341, opened 2019-06-27
WIP: Modernize conda environment

gdevenyi:
Remake conda environment using the dev environment in #340 and then re-exporting to have specific versions.

This fixes the tons of packages install by pip *and* the fact that the old environment was installing things and then upgrading them with pip.

### DIFF
--- a/conda-environments/dlc-ubuntu-GPU.yaml
+++ b/conda-environments/dlc-ubuntu-GPU.yaml
@@ -1,105 +1,173 @@
-#conda env. as used on Jun 18th 2019 for Woods Hole course on new Ubuntu 18.04
-name: dlc
+name: null
 channels:
   - conda-forge
   - defaults
 dependencies:
   - _tflow_select=2.1.0=gpu
-  - blas=1.0=openblas
+  - absl-py=0.7.1=py36_0
+  - asn1crypto=0.24.0=py36_1003
+  - astor=0.7.1=py_0
+  - atk=2.32.0=haf93ef1_0
+  - backcall=0.1.0=py_0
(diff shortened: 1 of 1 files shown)


gdevenyi:
These could also now be renamed as they're by-no-means ubuntu restricted. This is the beauty of using conda packages everywhere.

gdevenyi:
So, technically speaking, you only need *one* conda environment. These file should work everywhere.

I have no Windows or OSX to test.

MMathisLab:
Yeah, I will leave the others as options still, as for some reason we did hit some cross-platform issues last year.

DenisPolygalov:
Getting rid of the pip-based packages is only half of the solution. Another half is to keep strictly single (default) conda package distribution channel, i.e. get rid of the 'conda-forge'. The TensorFlow nowadays is also distributed within 'conda install', so don't need to install system-side stuff like cudatoolkit. At least on Windows. The only package that is absent in default conda channel is 'easydict'. After replacing 'ruamel.yaml' in the utils/auxiliaryfunctions.py with 'ruamel_yaml' DeepLabCut can be simply installed like this:
> conda create -n dlc
> activate dlc
> conda install tensorflow-gpu, matplotlib, wxpython, opencv, pandas, tqdm, yaml, PyYaml
> conda install ruamel_yaml, scikit-image, scikit-learn, statsmodels, click, pytables
> pip install easydict
Without using BS like Docker, pip or enforcing users to install older version of packages as dependencies.

MMathisLab:
Well, unfortunately that’s not quite the full story as dlc has several dependencies that need to be version fixed... also Docker is not BS ;)

PR #349, opened 2019-07-02
Added Skeleton Analyzer to Post Processing

FedeClaudi:
Added a function for skeleton analysis. 

- Similar arguments to `analyze_videos`, mainly config file path and list of video files
- it loads the `.h5` file with the predictions for one video, so need to have passed the video through `analyze_videos` first
- it extracts the length and orientation of each bone (as specified in config file) at each frame
- Saves the data as a multi index pandas dataframe (saved to a .h5 file) where each bone has name "bp1_bp2" and has columns Length, Orientation, Likelihood. 
- For the likelihood at each frame, the lowest of the likelihoods of the two body parts is taken.


MMathisLab:
Awesome, thanks so much :)

AlexEMG:
I guess you need to add it to init to become available.

Yaraat92:
This seems great! 
How do I make it work on my program? Is there something to download or install?

PR #405, opened 2019-08-23
Update predict_videos.py

ambareeshsrja16:
Solves " NameError: name 'TF' is not defined`" when deeplabcut.analyze_time_lapse_frames is called.

Refer: https://github.com/AlexEMG/DeepLabCut/issues/402 (first issue)


ambareeshsrja16:
OS: Ubuntu18.04 with Anaconda Env
DeepLabCut Version: 2.0.8
Browser: Chrome

Adding test script output. 
[testscript_output.txt](https://github.com/AlexEMG/DeepLabCut/files/3536700/testscript_output.txt)
The test script executes to the end.

Note:
A setting I have to make on my system is:
```
     import tensorflow as tf
     gpu_options = tf.GPUOptions(per_process_gpu_memory_fraction=0.5)
     sess = tf.Session(config=tf.ConfigProto(gpu_options=gpu_options))
```
which is because of a GPU issue (unrelated to the PR), which also might explain "Tensorflow enqueue operation cancelled" errors which pop-up in the output (which are harmless https://stackoverflow.com/questions/38678371/tensorflow-enqueue-operation-was-cancelled)

PR #410, opened 2019-09-04
Frame extraction class 

gwappa:
## Overview

1. Tried to fix an issue when loading `tensorflow` in my environment.
2. Tried to fix an issue during extraction of frames from videos whose FPS are not set properly.

## Environments

`tensoflow` falied to load on my conda environment (Windows + CUDA + Miniconda). I believe that there were some conflicts between installed packages as I created the environment.

Below are things that I did to solve the issue:

1. installing each package from scratch, without version specifications.
2. because the specs on `setup.py` resulted in re-download of the packages, I brutely removed the version specification on it, too (in hindsight, the lower bounds at least must have been left).
3. since `scipy.misc` does not have the `imread` and `imresize` functions, I created the `imageop` submodule and re-implement the proxy functions.

It seems to me that the installation of `tensorflow-gpu` is the only tricky part. If this is the case, I believe that, except for that package, it would be better to specify the versions by only using lower bounds (i.e. `>=` instead of `~=`).

The `imageop.py` submodule is found under the `utils` module.

I set up the `environments.yaml` file corresponding to my attempts. It looks like this "up-to-date" version runs DLC 2.0.8 on my other environment running Ubuntu 18.04 + CUDA + Anaconda.

## Reading frames from videos

I found that, if the grabber software did not include the FPS info into the video, `extract_frames` and other functions throw an exception due to zero-division. As having FPS-based information on the log still seemed useful, I tried to resolve the issue by adding a different approach of reading frames from videos.

In doing so:

1. I tried to make the code modular so that all the frame extraction-related code are restricted inside what I call as the "frame picker" class.
2. Basic video-related functions (i.e. related to `extract_frames` and `predict_videos`) were re-implemented based on it. The driver of the frame picker (e.g. `opencv` or `moviepy`) can be specified by the `videodriver` keyword argument of the function, and it defaults to `opencv`.
3. I added a `scikit-video` based frame picker. I chose it because it is merely a wrapper to `ffmpeg`, and reads raw frames one after another i.e. it does not really care about the FPS.

The frame-picker code is located in `utils/frame_pickers.py`.

A potential problem is that the user have to have installed `ffmpeg` and `ffprobe` on his/her environment (unlike `moviepy`). It may be a better solution to tweak the OpenCV-based frame picker code so that, if the FPS of zero (or less than e.g. `1e-3`) is detected, the frame picker automatically sets it to one.

Some functions related to outlier pickup and time-lapse prediction have not been touched so far (since I have not had the need personally), but I believe it would be not so hard to migrate in the future.

I hope these changes increase the usability...


AlexEMG:
Dear @gwappa, thanks for your PR -- I like most of the ideas you propose. However, they the PR will need to be expanded to "outlier pickup and time-lapse prediction" before they can be accepted. In the next weeks, I unfortunately have no time to do this, but I'll try to do that at some point.

caniko:
> 2. Tried to fix an issue during extraction of frames from videos whose FPS are not set properly.

Same issue that I [attempt](https://github.com/caniko2/DeepLabCut/blob/414faefe13a5b6d02cc779f4ea13d5cd4bc10a39/deeplabcut/generate_training_dataset/frame_extraction.py#L140) to resolve in #390

gwappa:
@caniko2 which of your commits are you referring to? I don't see where you try to solve it.

caniko:
> @caniko2 which of your commits are you referring to? I don't see where you try to solve it.

[Here](https://github.com/AlexEMG/DeepLabCut/pull/390/commits/2511df1cd0a27112ba729f64d027f02ea93e72bc), I haven't read through your fix, but I think it is more thorough than mine. I outright skip the video with the corrupt metadata. You seem to make an attempt to handle it, am I right?

AlexEMG:
I see - so you updated only the framepicker class now. I think it would be great to change 
deeplabcut/utils/imageop.py to deeplabcut/utils/auxfun_video.py?

gwappa:
as I get more and more familiarized with the code, I don't think my changes are worthwhile any more.
`frame_pickers` may be the only nice one if you would want to use it, but I don't think I can manage and test all of them.

You can reject this PR if you want. I will try smaller PR's from next time.

PR #457, opened 2019-10-16
RF: do not hardcode versions, specify minimal compatible for tables and imageio

yarikoptic:
Commits which introduced those hardcoded versions (e.g. ca93b3e7a69c674abb31bfaa812cb38940a5d598 for tables) provided no explanation, so I decided to be brave and bold to suggest this PR.

Without these changes I have failed to install DeepLabCut (in a virtualenv on a Debian Linux, with or without isolation from the system installed modules) because of e.g.
an issue in tables described here: https://github.com/fink/fink-distributions/issues/317#issuecomment-446937431

In general hardcoding versions, if really desired should be done via
requirements.txt or pipenv.  For more reasoning see e.g.
https://caremad.io/posts/2013/07/setup-vs-requirement/


AlexEMG:
Thanks for raising this -- I suppose we should have a requirements.txt (feel free to PR that :)) file then... 

As for the versions, especially for pytables many users have tons of issues when not using 3.4.3 (on various systems): https://github.com/AlexEMG/DeepLabCut/issues?utf8=%E2%9C%93&q=pytables

This is why we fix it.

yarikoptic:
> As for the versions, especially for pytables many users have tons of issues when not using 3.4.3

yeah, for any complex enough system, it is not unlikely to run into issues with newer versions of dependencies, and then there is a desire to just "fall back".  Our approach (e.g. in DataLad) is to try to keep the finger on "what is coming", and report them upstream asap to get them properly resolved, or then accounting for difference in behavior within our tool. E.g., (not without a pride) I can say that just a few days back we caused new releases of Python to come out free of a [regression](https://bugs.python.org/issue38449).  Yes -- such approach might cause more "short term" work, but IMHO it is also more productive in the long run (reduction in duplication, improved stability of the entire ecosystem, etc).

For "guaranteed to work" user deployments -- I largely rely on distributions (debian/conda) where we run tests during package build, and actually at times (with our debian hats on) validating uploads of upstream packages to (neuro)debian on datalad first.  There (in Debian and conda) versions are virtually never hard limited.

With those approaches we so far avoided hardcoding versions in requirements.txt or pipenv.  But for guaranteed deployment, pipenv probably would indeed be the way to go -- can be tuned per OS etc.  

Note also that within setup.py you can declare "not good" versions (e.g. `>= good.one,!=later.buggy`)

AlexEMG:
I think we go for guaranteed deployment.. (we want to avoid getting too many issues/emails/etc)

yarikoptic:
ironically, absence of such a guarantee (failed for me) is what triggered this PR ;-)

yarikoptic:
Oh, I only now spotted that there is no CI (Travis, circle, GitHub actions) of any kind!? That could explain some portion of issues arising. Even checking automagically that things still install/deploy would be of great help. I personally can't envision any code development without at least some ci these days

MMathisLab:
Sure - we are just a small team of neuroscientists (n=2 part-time developers), and new to software deployment, so learning on the fly ;). We aim to do this in the coming months. Thanks for the advice. But, we do have[ extensive test scripts](https://github.com/AlexEMG/DeepLabCut/tree/master/examples) to check all default cases and installation. The issue is we need to support macOS, linux, and Windows, and often the packages required to not align ...

yarikoptic:
I used Travis and a bit circle ci. Started to learn/use GitHub actions now. If you would like I could kick start one of those for you to help. If you decide to go with new GitHub actions, you will need to apply for it.

Here is eg my setup for dandi only a few days old: https://github.com/dandi/dandi-cli/blob/master/.github/workflows/test.yml which is quite minimalistic and ready to digest

Let me know if you need help on that, and let's continue in some dedicated issue/pr.

Cheers

PR #458, opened 2019-10-20
Use 'allow_growth' for training

1over137:
Not using this causes cudnn errors due to a tensorflow bug. If this is too hacky, please try to make it better.
From my tests, this change alone is enough to avoid problems for evaluation and testing.

### DIFF
--- a/deeplabcut/pose_estimation_tensorflow/train.py
+++ b/deeplabcut/pose_estimation_tensorflow/train.py
@@ -16,6 +16,8 @@ import threading
 import argparse
 from pathlib import Path
 import tensorflow as tf
+config = tf.ConfigProto()
+config.gpu_options.allow_growth = True
 vers = (tf.__version__).split('.')
 if int(vers[0])==1 and int(vers[1])>12:
     TF=tf.compat.v1
@@ -128,7 +130,7 @@ def train(config_yaml,displayiters,saveiters,maxiters,max_to_keep=5,keepdeconvwe
     restorer = TF.train.Saver(variables_to_restore)
     saver = TF.train.Saver(max_to_keep=max_to_keep) # selects how many snapshots are stored, see https://github.com/AlexEMG/DeepLabCut/issues/8#issuecomment-387404835
 
-    sess = TF.Session()
+    sess = TF.Session(config=config)
     coord, thread = start_preloading(sess, enqueue_op, dataset, placeholders)
     train_writer = TF.summary.FileWriter(cfg.log_dir, sess.graph)
     learning_rate, train_op = get_optimizer(total_loss, cfg)


AlexEMG:
Merged. For now you can install with: 

`pip install git+https://github.com/AlexEMG/DeepLabCut
`

I added a flag for train_networ:
`In [2]: deeplabcut.train_network?
deeplabcut.train_network(config, shuffle=1, trainingsetindex=0, max_snapshots_to_keep=5, displayiters=None, saveiters=None, maxiters=None, **allow_growth=False**, gputouse=None, autotune=False, keepdeconvweights=True)
`
with description:
```
allow_groth: bool, default false.
    For some smaller GPUs the memory issues happen. If true, the memory allocator does not pre-allocate the entire specified
    GPU memory region, instead starting small and growing as needed. See issue: https://forum.image.sc/t/how-to-stop-running-out-of-vram/30551/2
```


Just run: **`deeplabcut.train_network(path_config_file,allow_growth=True)`**

1over137:
> Merged. For now you can install with:
> 
> `pip install git+https://github.com/AlexEMG/DeepLabCut `
> 
> I added a flag for train_networ:
> `In [2]: deeplabcut.train_network? deeplabcut.train_network(config, shuffle=1, trainingsetindex=0, max_snapshots_to_keep=5, displayiters=None, saveiters=None, maxiters=None, **allow_growth=False**, gputouse=None, autotune=False, keepdeconvweights=True) `
> with description:
> 
> ```
> allow_groth: bool, default false.
>     For some smaller GPUs the memory issues happen. If true, the memory allocator does not pre-allocate the entire specified
>     GPU memory region, instead starting small and growing as needed. See issue: https://forum.image.sc/t/how-to-stop-running-out-of-vram/30551/2
> ```
> 
> Just run: **`deeplabcut.train_network(path_config_file,allow_growth=True)`**

Actually, later in my tests with larger frames, prediction also crashes. Please add that one too :)

AlexEMG:
Btw: you can also set this via an environment variable:

import os
os.environ['TF_FORCE_GPU_ALLOW_GROWTH'] = 'true'

or in the terminal before you start training/predicting
export TF_FORCE_GPU_ALLOW_GROWTH=True

shelhamer:
I encountered this same issue with tf 1.13.1 and a GTX 2080. It seems that many do encounter such an issue, but find resolution with the `allow_growth` setting, as mentioned in https://github.com/tensorflow/tensorflow/issues/24496#issuecomment-455265295. I recognized this symptom from another tf headache, but it might be worth a note in the demo notebooks in case it's someone's first time.

For instance, the [first invocation of training](https://github.com/AlexEMG/DeepLabCut/blob/master/examples/Demo_labeledexample_MouseReaching.ipynb?short_path=425dc0f#L149) in the reaching demo could either pass `allow_growth=True` or make a mention of trying it should there be a cuDNN error. I thought I'd suggest this since it's likely a starting point for users, but if it's too much of a corner case I can appreciate leaving it out and keeping the demo more streamlined if you prefer.

I was able to find this issue after all, so perhaps it isn't so dire.

zcq0730:
> Merged. For now you can install with:
> 
> `pip install git+https://github.com/AlexEMG/DeepLabCut `
> 
> I added a flag for train_networ:
> `In [2]: deeplabcut.train_network? deeplabcut.train_network(config, shuffle=1, trainingsetindex=0, max_snapshots_to_keep=5, displayiters=None, saveiters=None, maxiters=None, **allow_growth=False**, gputouse=None, autotune=False, keepdeconvweights=True) `
> with description:
> 
> ```
> allow_groth: bool, default false.
>     For some smaller GPUs the memory issues happen. If true, the memory allocator does not pre-allocate the entire specified
>     GPU memory region, instead starting small and growing as needed. See issue: https://forum.image.sc/t/how-to-stop-running-out-of-vram/30551/2
> ```
> 
> Just run: **`deeplabcut.train_network(path_config_file,allow_growth=True)`**

Could you please add this option to the GUI? Or could you please tell me how to set the default value as TRUE?  Thanks a lot for your help!

backyardbiomech:
> import os
> os.environ['TF_FORCE_GPU_ALLOW_GROWTH'] = 'true'

I can confirm that adding this at the start of the import cell in a custom jupyterlab notebook fixes this problem on Windows 10.

Issue #475, opened 2019-11-01
Create labeled video uses wrong FPS hence wrong duration


auesro:
**Your Operating system and DeepLabCut version**
Linux Mint 19.1, Anaconda, DeepLabCut 2.1.1

**Describe the problem**
The fps of the video is wrongly calculated (or read) as 1000 fps. So even if the real number of frames is correctly meassured: 18038, when create_labeled_video calculates duration as nframes/fps it gets a duration of 18.4 seconds. 
`Starting %  /media/augustoer/P aescalante/Behavoir/6-12-2017 ['/media/augustoer/P aescalante/Behavoir/6-12-2017/18349(Right)18351(Left).wmv']
Loading  /media/augustoer/P aescalante/Behavoir/6-12-2017/18349(Right)18351(Left).wmv and data.
604783
Duration of video [s]:  18.04 , recorded with  1000.0 fps!
Overall # of frames:  18038 with cropped frame dimensions:  640 360
Generating frames and creating video.
100%|████████████████████████████████████| 18038/18038 [02:47<00:00, 107.77it/s]`

The video is created with 18.4 seconds of duration even if the real video (recorded at 30 fps) is 10 minutes long.

I tried setting outputframerate=30 but the result was the same.

Any idea how to solve this? 

Thanks!

A

MMathisLab:
It's likely how openCV is reading the` .wmv` file; can you run create_labled_video and include `save_frames=True` - this uses a different video loader (and produces higher quality videos). 

You can also change the codec, and/or I have no experience with .wmv files, but we have no issues with mp4 or avi

```
deeplabcut.create_labeled_video?

Signature: deeplabcut.create_labeled_video(config, videos, videotype='avi', shuffle=1, trainingsetindex=0, filtered=False, save_frames=False, Frames2plot=None, delete=False, displayedbodyparts='all', codec='mp4v', outputframerate=None, destfolder=None, draw_skeleton=False, trailpoints=0, displaycropped=False)
```

auesro:
OK, I am giving it a try but saving frames makes the creation of the video like 50 times slower (it will take 1:40 hours to make a 10 minutes video). I will check if I can find out why these files are so missunderstood by opencv.
Thanks, Mackenzie,
A

MMathisLab:
awe sure, but does it print 1000 FPS still ?

auesro:
It actually does, and the end result is the same:

```
[...]

PR #476, opened 2019-11-02
Update video_processor.py

auesro:
Following issue #475 , I modified slightly `video_processor.py` .
Picking up FPS correctly from the original video is essential to obtain a labeled video of correct duration. Both openCV and ffmpeg implementations depend on OpenCV to read this tag from the header in the original video file.
At least one type of file (that with fourcc code `844516695` ) is read by OpenCV as to have 1000 fps (being original 30 fps), which gives rise to a labeled video 33.33 times shorter than the original.
If other file types were found to have the same problem, the `if` condition could be easily expanded to include them.

### DIFF
--- a/deeplabcut/utils/video_processor.py
+++ b/deeplabcut/utils/video_processor.py
@@ -125,6 +125,11 @@ class VideoProcessorCV(VideoProcessor):
         self.w = int(self.vid.get(cv2.CAP_PROP_FRAME_WIDTH))
         self.h = int(self.vid.get(cv2.CAP_PROP_FRAME_HEIGHT))
         all_frames = int(self.vid.get(cv2.CAP_PROP_FRAME_COUNT))
+        infourcc = int(self.vid.get(cv2.CAP_PROP_FOURCC)) #read fourcc code from original video
+        if infourcc == 844516695: #This fourcc code is not read properly/does not contain FPS information
+            self.FPS = self.FPS #Set FPS to init 30 fps
+        else:
+            self.FPS = self.vid.get(cv2.CAP_PROP_FPS)
         self.FPS = self.vid.get(cv2.CAP_PROP_FPS)
         self.nc = 3
         if self.nframes == -1 or self.nframes>all_frames:


AlexEMG:
Do you have any reason to believe that codec 844516695 is always misread?

auesro:
Actually no, that's the only metadata I could find to tell apart my non-accurately readable videos from those that work flawlessly.
Indeed, I tried something: I re-encoded a perfectly OpenCV-readable file to the WMV2 codec (thats what the OpenCV fourcc number 844516695 means) with FFMPEG and OpenCV was able to read the metadata without issue so I can only think that the Logitech acquisition software is not good at encoding videos or does not fill the metadata correctly... but so far I am unable to find another piece of information in the metadata informative about these weird videos...

So for now, I would hold off and not apply this PR. But maybe leave a message somewhere about the possibility for some videos not to work properly in OpenCV even though you dont recieve any warning.

MMathisLab:
I added a note to the troubleshooting tips wiki: https://github.com/AlexEMG/DeepLabCut/wiki/Troubleshooting-Tips

catubc:
Hi. Are there any plans to reconcile this branch?
Thanks 
Catubc

AlexEMG:
I think this specific fix, is not generally useful for users. Does 844516695 happen to be your fourcc number?

auesro:
@catubc could you check if your video fourcc code is the same one I was reporting?

MMathisLab:
we did add a more robust frame reader for some issues actually; https://github.com/DeepLabCut/DeepLabCut/blob/e5c3384897b39e3e33ca440ffef9a7830bb02a70/deeplabcut/pose_estimation_tensorflow/predict_multianimal.py#L34 

You can set `robust_nframes=True` and test it out.

AlexEMG:
Indeed, sometimes the metadata is incorrect with respect to the actual data. Then one solution, as linked by @MMathisLab is to first iterate over the frames (to count how many there are) and then use that. For the create labeled video steps one can also pass the target frame rate (if I am not mistaken)

auesro:
@AlexEMG you can pass the argument ```outputframerate``` but I tried [that](https://github.com/DeepLabCut/DeepLabCut/issues/475#issue-516292117) and the result was the same, wrong FPS, hence wrong duration labeled video. As far as I remember, there were multiple places in the [code ](https://github.com/DeepLabCut/DeepLabCut/issues/475#issuecomment-549157610)(from training till creating the labeled videos) where FPS was read through OpenCV, so practically overwritting your input framerate. 
But I must say I havent tried the latest version, and didnt know about the ```robust_nframes```.

catubc:
Hey everyone, so for me the issue was a 2-step issue:

1.  Original .wmv files were too high resolution 1280 x 640. So I used ffmpeg to reduce to 640 x 320 and save as .mp4 files.

2.  Then trained DLC on 640 x 320.

3.  Then labeled on 640 x 320.  But as I mentioned the pipeline reads the converted data as having 1000.0s FPS.

The solution that worked for me was to explicitly specify frame rate during the ffmpeg conversion process using -r 15 (in my case the vids had 15Hz).

`ffmpeg -i input.wmv -c:v libx264 -crf 23 -q:a 100 -vf scale=640:360 -r 15 output.mp4`

So now DLC reports the correct frame rates.

MMathisLab:
@auesro no worries! - `robust_nframes` was just added last week -  the list of actual updates would have been so long for the release note (it was 141 files changed and ~30K lines ... whelp!) Hence why, in part, we beta released it to be sure we didn't screw up everyone in one fell swoop :smile:

auesro:
Dear @MMathisLab , 
I am back to this. I was quickly getting back to speed again.
If I followed everything correctly [here](https://github.com/DeepLabCut/DeepLabCut/blob/e5c3384897b39e3e33ca440ffef9a7830bb02a70/deeplabcut/pose_estimation_tensorflow/predict_videos.py#L36) I think I found an unexpected behavior:
The option `robust_nframes` is not used for non multi-animal datasets even though you can explicit it in the call to the `analyze_videos` function, right? Because when dealing with non multi-animal datasets the code will go [here](https://github.com/DeepLabCut/DeepLabCut/blob/e5c3384897b39e3e33ca440ffef9a7830bb02a70/deeplabcut/pose_estimation_tensorflow/predict_videos.py#L303) and then opencv will be used for counting the number of frames, fps and duration, correct?
If thats the case, it would be great to be able to use the robust mode in non multi-animal datasets.

PR #484, opened 2019-11-11
Allow running module directly

jackbrookes:
Users who just want access to the GUI can run the module with `python -m deeplabcut` without the unnecessary steps of invoking an `ipython` session.

This can be built on with python's `argparse` and command line arguments.


AlexEMG:
Thanks! I suppose one should also check for the environment variable (i.e. is GUI support active):if `os.environ.get('DLClight', default=False) == 'True':`

AlexEMG:
Ok, I have added this feature. Thanks for your input. However, I opted for a `__main__.py` file instead. https://stackoverflow.com/questions/4042905/what-is-main-py

Issue #493, opened 2019-11-17
Ipython freezes after (-215:Assertion failed) !ssize.empty() in function 'resize'


ThejanW:
**Your Operating system and DeepLabCut version**

Please state your operating system, env, and which version of DeepLabCut you are using.
Example: Ubuntu 16.04 LTS, with an Anaconda Env, & DeepLabCut1.x or 2.x.

Please complete the following information about your system:

OS: Ubuntu 18.04
DeepLabCut Version: 2.1.3
Browser: chrome
TensorFlow Version: tensorflow-gpu==1.14
CUDA 10
GPU: GeForce RTX 2080 Ti

**Describe the problem**

I am trying to train a model with a ResNet-50 backend. I have 1000+ annotations. Both training and validation functions were working pretty well. But after few runs, an error started to appear at the same iteration index. But the training session doesn't stop. However, It eventually freezes my ipython session, advising me to force quit the session. I tried running this multiple times, it freezes at the exact same iteration too. Perhaps, one of my images are corrupted as it is having trouble with cv2.resize. But the training continues even after the error. 
Why does it freeze only after few more iterations from the occurrence of the error?

**How to Reproduce the problem**
1. Run DLC GUI
2. Start a training session with a ResNet-50 while keeping all the other settings as defaults

**Screenshots**
Here's the error
![image](https://user-images.githubusercontent.com/17251474/69016103-0676ab80-09c1-11ea-822a-27bf363e1a2b.png)

After some time, it freezes the session
![image](https://user-images.githubusercontent.com/17251474/69016112-1f7f5c80-09c1-11ea-9c3f-f9b3d415a200.png)

![image](https://user-images.githubusercontent.com/17251474/69016293-2e670e80-09c3-11ea-8b19-7f9e8aca46c9.png)

**Additional context**
I'm pretty sure this is a data specific problem. I just need some help figuring out the origination of this.  

Thanks.

AlexEMG:
Thanks @ThejanW for raising this -- this might be related to other issues: https://github.com/AlexEMG/DeepLabCut/issues/463 and https://github.com/AlexEMG/DeepLabCut/issues/468 that recently popped up. 

Can you make a catch for the cv2.resize and figure out why it fails? See e.g. https://github.com/aleju/imgaug/issues/205

[...]

PR #501, opened 2019-11-22
Fix for negative coordinates in the dataset

ThejanW:
Fix for issue #493 

Apparently, few of my annotations are having negative coordinates. I've changed trainingsetmanipulation.py to not to add such examples to the dataset.

Ubuntu 18.04
tensorflow-gpu==1.14
CUDA 10
GeForce RTX 2080 Ti

How to test: 

- Please download the .csv and .h5 files from this [link](https://drive.google.com/drive/folders/1t69jf0hl2u-odIvS5ImxZff-gN6hBCMf?usp=sharing)
- You can use those with **examples/Reaching-Mackenzie-2018-08-30**. This .csv has few negative coordinates to reproduce this behavior.   
- I have added few prints for debugging purposes. Will remove them once you test this.

[testscript.log](https://github.com/AlexEMG/DeepLabCut/files/3877415/testscript.log)

### DIFF
--- a/deeplabcut/generate_training_dataset/trainingsetmanipulation.py
+++ b/deeplabcut/generate_training_dataset/trainingsetmanipulation.py
@@ -641,19 +641,18 @@ def create_training_dataset(config,num_shuffles=1,Shuffles=None,windows2linux=Fa
                 indexjoints=0
                 joints=np.zeros((len(bodyparts),3))*np.nan
                 for bpindex,bodypart in enumerate(bodyparts):
-                    if Data[bodypart]['x'][jj]<np.shape(im)[1] and Data[bodypart]['y'][jj]<np.shape(im)[0]: #are labels in image?
+                    # check whether the labels are positive and inside the img
+                    x_pos_n_inside = 0 < Data[bodypart]['x'][jj] < np.shape(im)[1]
+                    y_pos_n_inside = 0 < Data[bodypart]['y'][jj] < np.shape(im)[0]
+                    if x_pos_n_inside and y_pos_n_inside:
                         joints[indexjoints,0]=int(bpindex)
                         joints[indexjoints,1]=Data[bodypart]['x'][jj]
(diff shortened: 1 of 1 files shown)


ThejanW:
Here's a sample output of the prints.

image

AlexEMG:
I think both the x & y coordinate should be removed. Also, I think it would be better to fix the issue in the loader itself. I.e. check for corrupt frames (post augmentation).

ThejanW:
Hi @AlexEMG. Both x and y get removed with this,
```
 joints = joints[np.where(
                    np.prod(np.isfinite(joints),
                            1))[0], :]
```

ThejanW:
A case where only x is negative,
image

A case where only y is negative,
image

A case where both x and y are negative,
image

Do you suggest editing these source files to not to load corrupted frames?
https://github.com/AlexEMG/DeepLabCut/blob/master/deeplabcut/pose_estimation_tensorflow/dataset/pose_defaultdataset.py
https://github.com/AlexEMG/DeepLabCut/blob/master/deeplabcut/pose_estimation_tensorflow/dataset/pose_dataset_deterministic.py
https://github.com/AlexEMG/DeepLabCut/blob/master/deeplabcut/pose_estimation_tensorflow/dataset/pose_dataset_imgaug.py
https://github.com/AlexEMG/DeepLabCut/blob/master/deeplabcut/pose_estimation_tensorflow/dataset/pose_dataset_tensorpack.py

AlexEMG (changes requested):
In Line 644 we check if data points are smaller than image size. (if Data[bodypart]['x'][jj]<np.shape(im)[1] and Data[bodypart]['y'][jj]<np.shape(im)[0]:) 

I think one could add the ">=0" case there as well.

ThejanW:
@AlexEMG Thanks for your input. I've made the suggested changes. Please review and merge.

AlexEMG (approved):

PR #520, opened 2019-12-06
fall back to config.yaml for num_outputs

lambdaloop:
This is a simple fix to go back to the old behavior for "num_outputs".
That is, it reads from the num_outputs in the `test/pose_cfg.yaml` as is current behavior, then if it doesn't find it there, reads from the `config.yaml` file. 

I find it much easier to modify the `config.yaml` file than the `test/pose_cfg.yaml` for each network

### DIFF
--- a/deeplabcut/pose_estimation_tensorflow/predict_videos.py
+++ b/deeplabcut/pose_estimation_tensorflow/predict_videos.py
@@ -164,7 +164,7 @@ def analyze_videos(config,videos, videotype='avi', shuffle=1, trainingsetindex=0
     dlc_cfg['init_weights'] = os.path.join(modelfolder , 'train', Snapshots[snapshotindex])
     trainingsiterations = (dlc_cfg['init_weights'].split(os.sep)[-1]).split('-')[-1]
     # Update number of output and batchsize
-    dlc_cfg['num_outputs'] = dlc_cfg.get('num_outputs', 1)
+    dlc_cfg['num_outputs'] = dlc_cfg.get('num_outputs', cfg.get('num_outputs', 1))
 
     if batchsize==None:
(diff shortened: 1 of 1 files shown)


AlexEMG:
Can you perhaps change it so that only falls back to the config.yaml file, if the value exists in there.

lambdaloop:
To clarify, do you want to completely ignore the values in `test/pose_cfg.yaml` then?
Or just prioritize the values in `config.yaml` over `test/pose_cfg.yaml` ?

AlexEMG:
No preference, I suppose. (I.e. which one overwrites)

lambdaloop:
Alright, I've made it default to look in config.yaml first, and only if it's not present to look in the test/pose_cfg.yaml . If it's not present in either, it defaults to 1.

AlexEMG:
Great, can you please also check the docs for this feature.

lambdaloop:
I'm not sure what you mean exactly. 
The only place the num_outputs feature is mentioned is in `docs/functionDetails.md`:
> As of 2.0.8+: you can extract multiple bodyparts, although there is no support for plotting or further analysis at this time; i.e. if you want to extract 3 snouts (as in [Figure 4 of Mathis et al, 2018 Nature Neuroscience](https://www.nature.com/articles/s41593-018-0209-y/figures/4)), you can edit the config.yaml file to contain ``num_outputs=3``. Then, when you run ``deeplabcut.analyze_videos`` it will extract the top three points (i.e. the 3 x, y, and likelihoods) and save this to the .h5 output file.

This pull request actually brings the latest code in line with the current documentation (enabling once again reading num_outputs from config.yaml).

AlexEMG (approved):

PR #551, opened 2020-01-16
Update conversioncode.py

7heRedFox:
I needed to export the h5 files to csv again and spotted a mistake in the name checking that prevented the function to achieve its goal. Here is a fix that worked for me (and should work for everyone)


AlexEMG:
Great catch - with the change of the scorer name this should be updated. It should also be made backwards compatible though.

Issue #557, opened 2020-01-21
Zero Division error when Analyzing video


byronrogers73:
Using Google Colab.
DeepLabCut Version 2.1.5.2
Chrome

**Describe the problem**

run: deeplabcut.analyze_videos(path_config_file,videofile_path)

Error returned:
```

Using snapshot-100000 for model /content/drive/My Drive/HorseBio-BR-2020-01-15/dlc-models/iteration-0/HorseBioJan15-trainset95shuffle1
Initializing ResNet
INFO:tensorflow:Restoring parameters from /content/drive/My Drive/HorseBio-BR-2020-01-15/dlc-models/iteration-0/HorseBioJan15-trainset95shuffle1/train/snapshot-100000
INFO:tensorflow:Restoring parameters from /content/drive/My Drive/HorseBio-BR-2020-01-15/dlc-models/iteration-0/HorseBioJan15-trainset95shuffle1/train/snapshot-100000
Analyzing all the videos in the directory
Starting to analyze %  2017Affidavit.mp4
Loading  2017Affidavit.mp4
---------------------------------------------------------------------------
ZeroDivisionError                         Traceback (most recent call last)
<ipython-input-20-131012eec99f> in <module>()
----> 1 deeplabcut.analyze_videos(path_config_file,videofile_path)

1 frames
/usr/local/lib/python3.6/dist-packages/deeplabcut/pose_estimation_tensorflow/predict_videos.py in analyze_videos(config, videos, videotype, shuffle, trainingsetindex, gputouse, save_as_csv, destfolder, batchsize, cropping, get_nframesfrommetadata, TFGPUinference, dynamic)
    214         #looping over videos
    215         for video in Videos:
--> 216             DLCscorer=AnalyzeVideo(video,DLCscorer,DLCscorerlegacy,trainFraction,cfg,dlc_cfg,sess,inputs, outputs,pdindex,save_as_csv, destfolder,TFGPUinference,dynamic)
    217 
    218         os.chdir(str(start_path))

/usr/local/lib/python3.6/dist-packages/deeplabcut/pose_estimation_tensorflow/predict_videos.py in AnalyzeVideo(video, DLCscorer, DLCscorerlegacy, trainFraction, cfg, dlc_cfg, sess, inputs, outputs, pdindex, save_as_csv, destfolder, TFGPUinference, dynamic)
[...]

PR #558, opened 2020-01-21
Check video validity early on.

jeylau:
Preliminary check to protect users against #557.


AlexEMG:
When I tested this, a "wrong path" is insufficient to create #557. (one gets the message: "No video/s found. Please check your path!")

However, if I pass a link to a corrupt video, then one gets indeed #557. Thus, I think your suggestion is good, but I will adapt the user feedback.

Issue #560, opened 2020-01-22
'bool' object is not iterable (in main GUI, Analyze video with cropping the video)


Shachafbe:
Windows 10, anaconda 3, GPU, DLC 2.1.5.2

After using DLC 2.0 for a few months (thank you for this tool!), I just updated to the new version of DLC. 
When using the new GUI, Analyze video **with cropping the video** gives the message:
"'bool' object is not iterable"
![dlc](https://user-images.githubusercontent.com/53897118/72875561-31e97f80-3cfd-11ea-9dd0-702e8e17d0aa.png)

Without cropping the analyzation now runs very slow. I must have deleted something with updating.. not sure what. 

Thank you so much for your amazing work, 
can't wait for the multiple animals' update!!!!

jeylau:
Good catch! Rather than True/False, the GUI should rather ask the user for the cropping coordinates. Otherwise, there is no way to get them from the config when analyzing new videos. We will push a fix within the next couple of hours.

PR #564, opened 2020-01-22
Define crop coordinates in GUI when analyzing videos.

jeylau:
The GUI was passing True/False rather than a list of coordinates to 'cropping' in analyze_videos.
#560 is now fixed, and the documentation has been slightly improved to avoid confusion ('cropping' has also been renamed to 'crop', consistent with all other functions expecting a list rather than a boolean).

### DIFF
--- a/deeplabcut/create_project/human_dataset.py
+++ b/deeplabcut/create_project/human_dataset.py
@@ -136,7 +136,7 @@ def create_pretrained_human_project(project,experimenter,videos,working_director
 
     if analyzevideo==True:
         # Analyze the videos
-        deeplabcut.analyze_videos(cfg,[video_dir],videotype,save_as_csv=True)
+        deeplabcut.analyze_videos(cfg, [video_dir], videotype, save_as_csv=True)
     if createlabeledvideo==True:
         deeplabcut.create_labeled_video(cfg,[video_dir],videotype, draw_skeleton=True)
         deeplabcut.plot_trajectories(cfg, [video_dir], videotype)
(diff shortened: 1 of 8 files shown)


MMathisLab:
The idea with cropping = True was that it should read the values from the config.yaml file though, but now I have to enter this again into the GUI? @jeylau 

and cropping and crop were separated as they have different names in config.yaml, albeit this is perhaps confusing.

MMathisLab:
also this means it's not backwards compatable with older projects, as cropping is used in the config.yaml
```

# Cropping Parameters (for analysis and outlier frame detection)
cropping: false
#if cropping is true for analysis, then set the values here:
x1: 0
x2: 640
y1: 277
y2: 624
```

MMathisLab (changes requested):
crop should be reverted to cropping, and the default behavior is that if cropping=true then the config.yaml file is read and those coordinates are used.

jeylau:
I have removed the 'cropping' panel in the GUI, so we now only rely on whether it is True/False in the config.yaml. I kept the argument as 'crop' in `analyze_videos()` to indicate we expect coordinates (which is also consistent with 'crop' in video_sets) vs 'cropping' (which implies a boolean).

MMathisLab:
PR good for me now! Thanks @jeylau - @AlexEMG if you're fine, please merge.

MMathisLab:
@alexemg please check and merge.

AlexEMG:
Now it's great!

PR #569, opened 2020-01-24
Addition of a Pose Prediction Plugin System to DLC

isaacrobinson2000:
## Pull Request Summary:

This pull request adds support for a plugin system for pose prediction algorithms. A pose prediction algorithm takes raw probability maps generated by the DeepLabCut neural network for each body part and returns the predicted locations for the body parts in each frame. It adds a flag to the `deeplabcut.analyze_videos` method called "predictor" which allows the user to select the prediction plugins currently installed on their DeepLabCut instance by name. To create a new plugin, one can simply write a class extending the `deeplabcut.pose_estimation_tensorflow.nnet.processing.Predictor` class, implement all the abstract methods, and place the new file in the `deeplabcut/pose_estimation_tensorflow/nnet/predictors/` folder (or any sub-folders of this folder). From here the plugin will be dynamically loaded into DeepLabCut and ready for usage. Note this system maintains full backward compatibility with the old "analyze_videos" method because the plugin selected when the user doesn't pick one ("argmax") uses the same code the original post-processing methods used. Some helper methods were also added for listing the current plugins and their configurable settings(`list_predictor_plugins`, `get_predictor_settings`, and `test_predictor_plugin`). Settings are configured in the config.yaml by placing the settings under a "predictors" header followed by another header matching the plugin's name, as demonstrated below:

```yaml
predictors:
     plotterargmax:
          # Settings for "plotterargmax" plugin go in here...
     argmax:
          # Settings for "argmax" plugin go here... 
```

As part of the plugin system, some helpful frame extraction classes (`TrackingData` and `Pose` classes) and a testing API are included. 

### Purpose of the Pull Request:
The purpose of this pull request is to provide an easy way of rapidly modifying and creating new pose prediction algorithms, which can currently only be done by manually adding code to the `analyze_videos` method. I initially developed the plugin system while implementing a method (now a plugin) for multi-animal tracking. That plugin is not included in this pull request, as it is not yet ready for release. This new API also offers an easy way of extracting probability maps to other formats for programs to use outside of DeepLabCut. 

### Included Plugins:

There are three plugins included in the pull request, "argmax", "plotterargmax", and "fast_plotterargmax". "argmax" is the default plugin and predicts poses in the same exact way as DeepLabCut currently does using the new API. The "plotterargmax" plugin predicts frames in an identical manner as "argmax" but also writes the probability frames as a 2D or 3D color map video using matplotlib. The "fast_plotterargmax" plugin is a much faster version of "plotterargmax" which doesn't have 3D support. I have attached some videos below that were generated by the "plotterargmax" plugin using different settings.

#### "plotterargmax" Output:
----------------------------------------------------------------------------------
```yaml
"3d_projection": True
"use_log_scale": True
```
_________________________________________________
output
_________________________________________________
```yaml
"3d_projection": False
"use_log_scale": False
```
________________________________________________
output1
_________________________________________________
```yaml
"3d_projection": True
"use_log_scale": False
```
_______________________________________________
output2
_________________________________________________

### Documentation:

All methods, classes, and APIs added for the plugin system are heavily documented via python doc strings. There is a user guide/tutorial in the "docs" folder ("docs/PluginOverview.md")

### Other Minor Additions:
 - Addition of a new multi output format which plots multiple predictions for a single body part as separate body parts (on `analyze_videos` method). Turned on by setting the "multi_output_format" argument to "separate_bodyparts"
 - `create_labeled_video` patched to plot all body parts when output is in the multi pose "separate_bodyparts" format.
 - Argument "num_outputs" added to to `analyze_videos` which overrides the config.yaml setting when set.

### An Example Plugin:

Attached below is a simple example plugin which extracts the probability maps to a text format for later processing by other programs (To use it, it needs to be renamed to exampleplugin.py). It showcases all of the features of the new plugin system (testing, settings, etc). The current user guide provides a very simple plugin that doesn't take advantage of all the plugin API features, and therefore it might be desirable to include this detailed example in more expansive documentation. 

[exampleplugin.txt](https://github.com/AlexEMG/DeepLabCut/files/4110344/exampleplugin.txt)


MMathisLab:
Thanks @isaacrobinson2000 and @traviswheeler for this PR - we will definitely review and get back to you in the next weeks! Thanks very much for contributing to DLC!!

traviswheeler:
Checking in - is there anything @isaacrobinson2000 or I can do to ease the burden for folks reviewing the pull request?

AlexEMG:
We are in the process - and have been for a quite while -- to roll to DLC 2.2 (our multianimal version), which changes the inference code substantially. For that reason, before this happens, I don't think it makes sense to include this PR.

AlexEMG:
Dear @traviswheeler and @isaacrobinson2000! Thanks for this contribution. We finally released part of 2.2 as a beta version. It would be wonderful if we could have a discussion on how to combine your contribution now?

traviswheeler:
Congrats on getting the 2.2 beta out the door - the changes are exciting! @isaacrobinson2000 will work to resolve conflicts between the PR and your new changes, then we should chat - shouldn't be long.  How would you like to "have a discussion"?  If you mean "actually talk to each other", we can arrange via the e-mail address associated with my account.

Issue #580, opened 2020-02-07
Videos not copying when creating project with GUI


backyardbiomech:
**Describe the bug**
At least on macOS Catalina, when creating a new project, selecting one or multiple videos and choosing to copy videos leads to a project without copied videos, just linked videos. Copying works as expected in the same env on the command line.

**To Reproduce**
1. Launch gui
2. enter project name and experimenter
3. Load a video or several
3. check the "do you want to copy the videos" box
4. Check the Yes radio button
5. Click ok

**Expected behavior**
Full video files in the videos folder, and proper paths in config.yaml

**Desktop (please complete the following information about your system):**
 - OS: macOS 10.15.2
 - DeepLabCut Version 2.1.5

**Additional context**
lines 104-105

Is this it on lines 104-105 of create_new_project.py? Yes and no are reversed?

```
        self.yes = wx.RadioButton( self, -1, "No", style = wx.RB_GROUP)
        self.no = wx.RadioButton( self, -1, "Yes")
```

PR #583, opened 2020-02-09
fixed GUI copy vids not copying

backyardbiomech:
Fixes issue  #580 

GUI copy videos radio buttons were not bound to the function that set self.copy, and when bound it was passing the radio button label ("Yes", "No") rather than True/False.

The GUI checkbox was also not Enabling/Disabling the radio buttons completely. 

This pull request fixes the binding and enabling, and changes the booleans in new.py to also accept "Yes" and "No".


MMathisLab:
This conflicted with #584 (seems we wrote the patch at the same time, apologies!)

PR #645, opened 2020-04-02
Fix dependency hell issue with moviepy/NumPy.

### LINKED ISSUE #644: If NumPy is locked to v. 1.16, moviepy should be locked to 1.0.1 or less to avoid dependency hell.
**Describe the bug**
Within _setup.py_ [here](https://github.com/AlexEMG/DeepLabCut/blob/master/setup.py#L30), NumPy is locked to version 1.16.4.  The latest release of moviepy (1.0.2), however, requires a version of NumPy greater than or equal to 1.17.3 (see [here](https://github.com/Zulko/moviepy/blob/master/setup.py#L75)).  This causes a dependency hell issue.  moviepy 1.0.1 does not mandate a specific version of numpy (see [here](https://github.com/Zulko/moviepy/blob/v1.0.1/setup.py#L70)).  Proposed fix would be to lock moviepy at version <= 1.0.1 in _setup.py_.

**To Reproduce**

```
(foo) [jsp2205@ax08 ~]$ pip install deeplabcut
Collecting deeplabcut
...
moviepy 1.0.2 has requirement numpy>=1.17.3; python_version != "2.7", but you'll have numpy 1.16.4 which is incompatible.
Installing collected packages: proglog, numpy, imageio-ffmpeg, moviepy, intel-openmp, tabulate, msgpack-numpy, termcolor, tensorpack, opencv-python, Shapely, imgaug, ruamel.yaml.clib, ruamel.yaml, easydict, deeplabcut, mock
```

jpellman:
Minor code change that should fix https://github.com/AlexEMG/DeepLabCut/issues/644 (issue where different mutually exclusive numpy versions are requested by deeplabcut and the latest version of moviepy (1.0.2).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,7 @@ setuptools.setup(
     install_requires=['certifi','chardet','click','easydict',
                       'h5py~=2.7','intel-openmp','imgaug',
                       'ipython','ipython-genutils',
-                      'matplotlib==3.0.3','moviepy','numpy==1.16.4','opencv-python~=3.4',
+                      'matplotlib==3.0.3','moviepy<=1.0.1','numpy==1.16.4','opencv-python~=3.4',
                       'pandas','patsy','python-dateutil','pyyaml>=5.1','requests',
                       'ruamel.yaml~=0.15','setuptools','scikit-image','scikit-learn',
                       'scipy','six','statsmodels','tables',


AlexEMG:
Fair enough, note that moviepy is anyway only used, if there a video doesn't play well with OpenCV (which is very rare).

Issue #646, opened 2020-04-03
"specify the shuffle" using the GUI unable shuffle 0


Shachafbe:
**Describe the bug**
After using the compare models (on 3 models:resnet_50, resnet_101,mobilenet_v2_1.0 ), the shuffle given to resnet_50 was zero.
In the next tracking steps, as in analyze videos or extract outlier frames, when I changed the shuffle to zero it immediately changed it to 1 as the minimum possible number is 1. 

**To Reproduce**
1. Use the GUI
2. Create a training set with 3 models to compare. 
3. Try to analyze video with shuffle 0 using the GUI

**Desktop (please complete the following information about your system):**
 - OS: windows 10
 - DeepLabCut Version '2.1.6.2'

Many thanks in advance,
Shachaf.

MMathisLab:
it does seem in Windows it defaults to "0" whereas in MacOS and Linux it starts at 1. We can change the GUI such that you can analyze videos with shuffle=0; n the future we should modify this behavior so that it is the same across OS. @jeylau - thanks for the note, @Shachafbe.

AlexEMG:
Fixed with #647

PR #647, opened 2020-04-03
Shuffle 0 patch1

MMathisLab:
fix for #646 - if shuffle 0 was used to create training set, allow for 0 to be used in subsequent GUI options.


AlexEMG (changes requested):
Changed one more range to 0 as well. Does that make sense?

AlexEMG (approved):

PR #649, opened 2020-04-08
Export module

gkane26:
### changes:
- adds export.py to deeplabcut/pose_estimation_tensorflow
  - includes export_model function, along with some helpers
    - exports pose_cfg.yaml, checkpoint, and pb files for a specified model
- adds import for export_model function (from deeplabcut import export_model)
- adds export_model to cli.py for command line interface to export_model function

### To Do
- currently copies pose_cfg.yaml file for the exported model. Maybe add additional information or change to deploy_cfg or export_cfg?


MMathisLab:
perhaps it should specify which snapshot to package, i.e. minimally grab the snapshot from the config.yaml?

gkane26:
just added snapshotindex as a parameter to the export_model function and updated export_model in the helper function docs. If provided it will overwrite the snapshotindex in config.yaml. If not provided it will use the specified snapshotindex. If snapshotindex == 'all', it uses snapshotindex = -1

MMathisLab (approved):
verified functionality.

Issue #670, opened 2020-05-01
Issue with repeating refining labels


hosuk88:
**Describe the bug**
The first frame's label is not matched when repeating the refining labels.
It seems the same issue as #353 and #374 

**To Reproduce**
1. Run deeplabcut.refine_labels(path_config_file)
2. In the GUI, load data, refine all frames, save, and click yes for repeating.
3. In the second data, refine the labels in the first frame, and go to the next frame. Go back to the first frame, and the labels were messed up and not at the manually refined positions.

**Expected behavior**
The previous data's last frame labels are not cleared, still remained, and affect to the next data's 1st frame. This issue consecutively happens - second data's last frame affects the third data's first frame, etc - during keep clicking yes for the repeating refining. Maybe the GUI issue.

**Screenshots**
![image](https://user-images.githubusercontent.com/24601048/80803260-9c188880-8b66-11ea-82a1-11dbac465edf.png)

**Desktop (please complete the following information about your system):**
 - OS: Windows 10
 - DeepLabCut Version: 2.1.6.2
 - Browser: both in chrome & edge

**Additional context**
It did not happen when I refine data separately (close the GUI and re-run deeplabcut.refine_labels),  not repeated in the same GUI.

MMathisLab:
Thanks for this detailed report! We will look into it ASAP

MMathisLab:
Hi @hosuk88 - I have made a PR (#671), if you'd like to checkout that branch and double check it fixed the behavior you describe, that would be great.

MMathisLab:
#671 should fix this, thanks!

AlexEMG:
Thanks for your documentation, this was very useful for reproducing the error!

hosuk88:
Thanks a lot! I'll try this ASAP.
Also, found something in the committed file. Should change this
remove_data() function as following "delete_data()"?
[image: image.png]

<http://www.avg.com/email-signature?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
Virus-free.
www.avg.com
<http://www.avg.com/email-signature?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
<#DAB4FAD8-2DD7-40BB-A1B8-4E2AA1F9FDF2>

Issue #581, opened 2020-02-08
Wrong indexing of label positions 


AdoNunes:
**The bug**
Previously I commented glitches in the label positions when relabeling or labeling new frames [https://github.com/AlexEMG/DeepLabCut/issues/574#issuecomment-582221262](url)
I tracked down the bug. In my case, I tried to create a project in Windows but I had problems when creating a symbolic link or copying over the videos. To save time, I created the project and labeled frames in Unix. Then I trained the model, and then refining labels or labeling frames, I would often see the labels of another frame into the wrong frame. As commented in #574 the frame number would even go to negative indexes. 

I did not dive into the code, but the reason is: if I have 5 images in a video labeled folder, the script reads the 5th image and calls the label file (h5/csv) and looks for the 5th row (+3 header rows), without taking into account the file name that is present in the first column and same raw. In addition, the path in the name is hardcoded with system-dependent slashes ('/', '\'). 

This bug leads to many unexpected behaviors. In my case, labels were done in linux, re-labeling in Windows. The label file is sorted in descending order, '/' precedes '\', so all the linux created images were before all the windows ones, leading to unmatched label-frames. If an image is deleted, the code fails to get the right label positions. 

This unexpected behavior has previously caused issues, and there are some patch functions such as `convertannotationdata_fromwindows2unixstyle` or `dropannotationfileentriesduetodeletedimages`. However, these do not address the problematic code. In my case, `dropannotationfileentriesduetodeletedimages` doesn't work as it just flips the slash to the opposite side without taking into account the O.S.

**Solution**
Don't hard code the row index of the frame label positions, or match the first col. name with the file name being read, or when creating the label file, sort the images names by taking the base name, not the hardcoded path with forw./back. slashes.

**To Reproduce**
Create two labeled frames in linux and two in windows. The label positions will be wrong. 

**Fix for unsorted rows in the CollectedData.csv files**
I provide the code I used to change all the paths in the first column of the collectedData.csv. Then it 

This sorts the rows without changing the names in the first column:
```

import pandas as pd
from glob import glob
import numpy as np
from pathlib import Path

path_root = "D:/projects/Ataxia/FingerTapping-Adonay-2020-01-29/labeled-data/"

path_labels = glob(path_root + '*Finger_Tapping')
csv_first_cols = ['scorer', 'bodyparts', 'coords']
nfirst =  len(csv_first_cols)

[...]

Issue #206, opened 2019-02-12
delete accidentally placed label


mschart:
Hi there,

I'm using the GUI for labeling and I sometimes had the problem to place a point too quickly, hence I'd like to delete it (as that point is not seen in the given frame). Now I need to restart or keep that error in there and hope for the best. The option to delete an accidentally picked point would be nice.

Cheers

AlexEMG:
Hey!

The newest version >2.0.4 can do that (pip install deeplabcut==2.0.4.1). You need to use the middle mouse key. The new GUI also allows intermediate saving, moving all points at any point, going back and forth ... It is really convenient - let me know how it works for you!

Cheers,
Alexander

P.S. now one can also add (missed) labels at any later point: https://github.com/AlexEMG/DeepLabCut/issues/147
P.P.S: before 2.0.4 the solution for deleting was to use the refinement GUI (and adjust the original labels). This method could not add points though, now adding, removing & changing are supported.

qsimeon:
what's the equivalent of "middle mouse button" on a macbook though?

catubc:
I am also looking for the same.

backyardbiomech:
Look for bettertouchtool (https://folivora.ai/buy). It's very affordable and does much more, and does allows middle click with trackpad or magic mouse
.

MMathisLab:
@catubc @qsimeon - in a near release it will also have a hot key: i.e the "delete" button.... now that I have to work from home on my macbook, I also am annoyed by this ;).

catubc:
Hey guys, I think there's some mac settings for doing this, here's a youtube video:

https://www.youtube.com/watch?v=VIr9R1P2wys

PR #671, opened 2020-05-02
Fix for bug #670, #581 ... 

MMathisLab:
Per the excellent bug report in #670 there is indeed a bug if one (1) open refinement GUI, save, and ops to open another folder, go forward (next) then go back (prev). The code was not properly resetting between this save and re-launch mode. Now confirmed to be working as desired (fix for #670, #581):

ezgif com-video-to-gif (1)

This PR also includes a delete hotkey in the GUI, and labeling GUI (from @jeylau! <3). So those of you without middle wheel clicking ability can delete data if needed. Simply click on the point, then use your "delete" See #206. 

Given the nature of this bug, I would suggest a roll-up to 2.1.8 ? @AlexEMG


AlexEMG:
Tested the hotkey: on my keyboard in Ubuntu it called by "backspace"

PR #688, opened 2020-05-16
Fast image loading/writing from labeled data

AlexEMG:


AlexEMG:
Solves https://github.com/AlexEMG/DeepLabCut/issues/652

PR #689, opened 2020-05-16
Bodypart ordering

AlexEMG:
Related to https://github.com/AlexEMG/DeepLabCut/issues/684 and https://forum.image.sc/t/label-mismatch-between-plotted-labeled-frames-and-plotted-evaluation-results/37777


AlexEMG:
Note: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.concat.html Sort non-concatenation axis if it is not already aligned when join is ‘outer’. This has no effect when join='inner', which already preserves the order of the non-concatenation axis.

Changed in pandas version 1.0.0: Changed to not sort by default. [Users should use at least pandas 1.0.0]

PR #709, opened 2020-05-27
Save individual maps rather than a grid of them

jeylau:


MMathisLab:
I think this is worth finishing, i.e. so each image can get a map, vs. subplots :)

jeylau:
@MMathisLab, let me know if the output looks better to you 😃!

MMathisLab:
looks very nice! The `scoremaps` and `LocRef` looks good; `PAFs`
 seem to be all separate images; maybe same image but different colors vs. 50 images if 50 PAFs? i.e. only 1 is shown as in this screencap

Screen Shot 2020-07-15 at 7 25 04 PM

jeylau:
I tried plotting all graphs on the same image, but they overlap each other (which makes it unreadable even with different colors). I could add a keyword to let the user select either that behavior or the frame-by-frame fields, if that sounds good. Anyway, don't you think this function makes sense for only a subset of bodyparts?

AlexEMG (approved):
Perfectamente!

PR #714, opened 2020-05-29
Cleaner imports

AlexEMG:


AlexEMG:
Passes linux/windows/osx tests; but 3D script @jeylau

MMathisLab:
looks great to me. @AlexEMG I think you found a few liltte GUI things, and once those are added PR ✅

PR #717, opened 2020-05-30
Better SkeletonBuilder

jeylau:
Skeleton is built from the config if existent (rather than from scratch systematically).
The necessity of having a fully labeled animal has been relaxed so that the Builder will always show up?

### DIFF
--- a/deeplabcut/utils/auxfun_multianimal.py
+++ b/deeplabcut/utils/auxfun_multianimal.py
@@ -47,8 +47,12 @@ def getpafgraph(cfg, printnames=True):
 
     bodypartnames = multianimalbodyparts + uniquebodyparts
     lookupdict = {bodypartnames[j]: j for j in range(len(bodypartnames))}
-    # print(lookupdict)
 
+    if cfg['skelton'] is None:
+        cfg['skelton']=[]
+
+    # CHECKS if each bpt is connected to at least one other bpt
+    # TODO: check that there is a path leading from each (multi)bpt to each other (multi)bpt!
     connected = set()
     partaffinityfield_graph = []
     for link in cfg["skeleton"]:
@@ -64,7 +68,8 @@ def getpafgraph(cfg, printnames=True):
     unconnected = set(range(len(multianimalbodyparts))).difference(connected)
     if unconnected:
         raise ValueError(f'Unconnected {", ".join(multianimalbodyparts[i] for i  in unconnected)}. '
-                         f'For multi-animal projects, all bodyparts should be connected. '
+                         f'For multi-animal projects, all multianimalbodyparts should be connected. '
+                         f'Ideally there should be at least one (multinode) path from each multianimalbodyparts to each other multianimalbodyparts. '
                          f'Please verify the skeleton in the config.yaml.')
 
     if printnames:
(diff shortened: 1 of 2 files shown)


MMathisLab:
If my config.yaml is:

```
individuals:
- individual1
- individual2
- individual3
uniquebodyparts:
- ice1
- ice2
- ice3
multianimalbodyparts:
- bodypart1
- bodypart2
skeleton:
bodyparts: MULTI!
```
I label frames, and then run build skeleton:

```
  File "/Users/mwmathis/Documents/DeepLabCut/deeplabcut/gui/label_frames.py", line 139, in build_skeleton
    skeleton.SkeletonBuilder(self.config)
  File "/Users/mwmathis/Documents/DeepLabCut/deeplabcut/utils/skeleton.py", line 58, in __init__
    for bone in self.cfg['skeleton']:
TypeError: 'NoneType' object is not iterable
  File "/Users/mwmathis/Documents/DeepLabCut/deeplabcut/gui/label_frames.py", line 139, in build_skeleton
    skeleton.SkeletonBuilder(self.config)
  File "/Users/mwmathis/Documents/DeepLabCut/deeplabcut/utils/skeleton.py", line 58, in __init__
    for bone in self.cfg['skeleton']:
TypeError: 'NoneType' object is not iterableTraceback (most recent call last):
  File "/Users/mwmathis/Documents/DeepLabCut/deeplabcut/gui/label_frames.py", line 139, in build_skeleton
    skeleton.SkeletonBuilder(self.config)
  File "/Users/mwmathis/Documents/DeepLabCut/deeplabcut/utils/skeleton.py", line 58, in __init__
    for bone in self.cfg['skeleton']:
TypeError: 'NoneType' object is not iterable
```

MMathisLab:
I must connect the skeleton first in the config.yaml, which is not the ideal behavior. 

Screen Shot 2020-05-30 at 3 44 37 PM

jeylau:
Aw ok, I thought our default was an empty list in the config
skeleton: []
I'll account for both cases.

MMathisLab:
I guess this is an edge case if someone deletes the "fake" skeleton that is put in by default, that is true ;)

AlexEMG:
Yes, def. should be []. But better to cover both.

jeylau:
Let me just also add the test for the paf graph and this will look good :)

MMathisLab (approved):
works great, thanks @jeylau !!!

jeylau:
Please let me know if you feel the error is documented enough!

MMathisLab (commented):
`bodypartnames = multianimalbodyparts + uniquebodyparts`

currently the SkeletonBuilder does not allow one to visualize and connect the unique bodyparts, which upon further conisderation, perhaps it should do; i.e. now it only shows the `multianimalbodyparts` but indeed the unique points "can" be connected, while the `multianimal` must be, correct @AlexEMG ?

so, here it's check for only multianimal bodyparts, but should we warn if any bodyparts (as defined in the code sniplet above) are not-connected?

MMathisLab:
also small thing, but if it's `skeleton:` in the config.yaml, it doesn't print the warning, it just doesn't work ;). if it's `skeleton: [ ]` works nicely :)

jeylau:
> also small thing, but if it's `skeleton:` in the config.yaml, it doesn't print the warning, it just doesn't work ;). if it's `skeleton: [ ]` works nicely :)

https://github.com/DeepLabCut/DeepLabCut/pull/717/commits/02e10fd114e3ae988836ff0e4b3791e263529e24 should handle that case, I've tested it :)

AlexEMG (approved):
Great!

PR #534, opened 2019-12-21
New feature:  real-time multiple object tracking based DLC mode

ehsainit:
pjimage

### Goal & workflow 
The goal is to extend DLC to support real-time multiple object tracking. For simplicity and efficiency, a traditional tracking by detection approach was introduced.

The DLC network detects a scale of high probabilities representing animals or look-alike objects, which allows modeling the score-map as a function with multiple local maxima.

graph 

Afterwards, candidates _detections_ will be collected from the score-maps, all of the candidates will be assigned as new _births_ ,  each having a unique identity, [Kalman filter](https://de.wikipedia.org/wiki/Kalman-Filter) instance initialized with the first _detection_ , a constant signifying the amount of skipped frames because of no available _detection_ and a unique color. After that, the _detection_-_track_-association problem is solved using the Hungarian algorithm. Only motion-information ([Mahalanobis distance](https://en.wikipedia.org/wiki/Mahalanobis_distance)) is used.
Next, the _tracks_ will be maintained and updated using the matched detection, only if their motion metric value is reasonable (e.g.  not too high), otherwise, they will be re-assigned. _Detections_ with no existing _tracks_ will be assigned as _births_. 
Finally the matched _tracks_ Kalman filter instances will be updated using their correspondent's _detection_.  To deal with the unmatched _tracks_, we keep track using the _a priori_ information that their Kalman filter instances hold. If they remain unmatched up tp certain number of frames they will be assigned as _deaths_.

p4

**Note**: This was done in the course of my studies as a Bachelor's project in the [Straw Lab](https://strawlab.org/) at the University of Freiburg. The Straw Lab is interested in real-time flies tracking , so the cost metric & other stuff were specifically chosen with the regard to the input.

### Glossary:
- **_Detection_** : an instance of the class Detection, contains coordinates , height and width of a local maximum extracted from the score-map
- **_Track_** : an instance of the class Track, contains information about an associated _detection_ 's coordinates  or predicted coordinate, associated color.
-  **_Birth_** : new _track_ instance, i.e. new animal or object detected  by the NN and was not associated with any existing _tracks_
-  **_Death_** : an existing _track_ instance with a skipped frame number greater than a threshold, i.e. the animal or the object no longer detected. Therefore deleted.
 
### Steps:

At time step _t_, for a processed frame: 

1. DLC inference returns a score-map which contain probability information about the processed frame
2. Maximum and minimum filters are applied to extract the **n** objects with probability greater than a threshold (i.e. local maxima) & assigned as _detections_
[...]

PR #736, opened 2020-06-10
New feature: TrackByDetection - real-time multiple objects tracker 

ehsainit:
### Goal & workflow 
The goal is to extend DLC to support real-time multiple instances-of-one-object tracking. For simplicity and efficiency, a traditional tracking by detection approach was introduced.

## How does it work ?
Candidates _detections_ will be collected from the score-maps, all of the candidates will be assigned as new _births_ ,  each having a unique identity, [Kalman filter](https://de.wikipedia.org/wiki/Kalman-Filter) instance initialized with the first _detection_ , a constant signifying the amount of skipped frames because of no available _detection_ and a unique color. Afterwards, the _detection_-_track_-association problem is solved using the Hungarian algorithm. Only motion-information ([Mahalanobis distance](https://en.wikipedia.org/wiki/Mahalanobis_distance)) is used.
Next, the _tracks_ will be maintained and updated using the matched detection, only if their motion metric value is reasonable (e.g.  not too high), otherwise, they will be re-assigned. _Detections_ with no existing _tracks_ will be assigned as _births_. 
Finally the matched _tracks_ Kalman filter instances will be updated using their correspondent's _detection_.  To deal with the unmatched _tracks_, we keep track using the _a priori_ information that their Kalman filter instances hold. If they remain unmatched up tp certain number of frames they will be assigned as _deaths_.

p4

**Note**: This was done in the course of my studies as a Bachelor's project in the [Straw Lab](https://strawlab.org/) at the University of Freiburg. The Straw Lab is interested in real-time flies tracking , so the cost metric & other stuff were specifically chosen with the regard to the input.

### Steps:

At time step _t_, for a processed frame: 

1. DLC inference returns a score-map which contain probability information about the processed frame
2. Maximum and minimum filters are applied to extract the **n** objects with probability greater than a threshold (i.e. local maxima) & assigned as _detections_
3. Create _tracks_ if the frame is the first or if no _tracks_ exists  
4. Data association: associate new _detections_ with existing _tracks_, according to a cost metric, remove _deaths_, assign new _births_ etc. 
5. Feed relative-to-object-size bounding boxes of _matches_, predict coordinates for _unmatches_

 ### Evaluation Results
This feature was evaluated using 1, 3 and 10 flies offline due to lack of time (back then), 
Videos proprieties : 
- FPS : 25
- Type :  mp4 
- Length : 20s- 180s
- GPU: ToDo 

Here are some of the results:

pre-one-fly-y-trej
**Trajectories plots** of tracking process on one fly. The 4 subplots diagrams demonstrate the accuracy performs of Kalman filter prediction against the DLC1 detectors .while the top two subplots illustrate the detectors performance over 40 second video, the other two subplots shows how accurate the Kalman filtering was.

cmp-one-fly-x-trej
cmp-one-fly-y-trej

**This figure** demonstrate a trajectories comparison between the  predictions (orange line)  and  the  DLC1  detection (blue line) over 20 seconds video of 3 flies.  Here we can see the x, y trajectories of one fly, 

annafunction2

**Illustration** of the prediction performance of the Kalman filter upon missing measurements over a 10 frames sequence.

p1
Here we can see the time complexity of the framework. the x-axis represent the time in second and the y-axis the number of tracked objects. 
 ### In Action
ezgif-5-b5fbc677896a

ezgif-5-734c87e2408f

ezgif com-video-to-gif


deeplabcut/pose_estimation_tensorflow/predict_videos.py
```diff
@@ -1344,11 +1345,18 @@ def convert_detections2tracklets(
             vname = Path(video).stem
             dataname = os.path.join(videofolder, vname + DLCscorer + ".h5")
             data, metadata = auxfun_multianimal.LoadFullMultiAnimalData(dataname)
-            method = "sk" if track_method == "skeleton" else "bx"
+            if track_method == 'skeleton':
```
MMathisLab: does not conform to black style (you can run black, i.e. `pip instal black`)

deeplabcut/pose_estimation_tensorflow/lib/single_object_tracker.py
```diff
@@ -0,0 +1,116 @@
+"""
+DeepLabCut 2.2 Toolbox (deeplabcut.org)
+© A. & M. Mathis Labs
```
astraw: I am not a lawyer, but I believe this copyright statement may be problematic.

As far as I know, @ehsainit did this work (started as part of a student project in my lab) in Germany. Under German law, he retains copyright any code he wrote. Therefore, even if he wanted to, he could not legally assign copyright to anyone else. (It may be possible under certain circumstances that typically involve a strong contract negotiated prior to employment that it is possible for companies to own the copyright of work produced by their employees in Germany.) I think OSS projects often say "(C) Copyright 2020, Deep Lab Cut authors" or similar to deal with this issue.

AlexEMG: Thanks for raising this point as well as your suggestion! 

We are currently thinking about how to address this (e.g. [WIP here](https://github.com/DeepLabCut/DeepLabCut/pull/746), your thoughts would be very welcome). My goals are to a) acknowledge all contributors b) keep DLC open-source and in accordance to its license, which means that [contributors should waive their copyright](https://opensource.guide/legal/) so that others can use, modify, and share the work.

astraw: I think "waiving" copyright is not the correct phrasing. Copyright is the legal means by which open source is enforced. Keeping the code copyrighted (even by many people) is how the license can be enforced. Waiving copyright seems counter to that goal (and, as far as I understand it, not possible under German law).

I think you want the contributor (copyright holder) to license the code under the DLC license. I didn't read your Individual Contributor License Agreement v2.0 in detail, but I think this is exactly what it does. My suggestion is just to change the header from "© A. & M. Mathis Labs" to "© DeepLabCut authors" (or similar) and then have an authors file that all contributors get added to and acknowledges their work. (You may already have this file, I didn't check.) They hold their respective copyrights but grant the provisions specified in the CLA which indeed keep DLC open source in accordance with its license.

Take all this with a grain of salt, however. I am not a lawyer.

AlexEMG: Yes, I also think that Contributor License Agreement v2.0 does just that. 

I like the suggestion regarding "© DeepLabCut authors". We do have an [authors file, in which we highlight individual contributions](https://github.com/DeepLabCut/DeepLabCut/blob/master/AUTHORS) that should be expanded.

astraw: Looks good to me, thanks!

MMathisLab (changes requested):
Hi @ehsainit  an additional check point would be good to throw an error is it's not an appropriate tracker, i.e. if someone runs this on say two individual mice, this does not work, i.e. 

```python

Analyzing /Users/mwmathis/Desktop/two_white_mice_052820-SN-2020-05-28/videos/xxx_200000.h5
3811it [00:06, 624.46it/s]
  File "/Users/mwmathis/Documents/DeepLabCut/deeplabcut/gui/analyze_videos.py", line 417, in convert2_tracklets
    track_method=self.trackertypes.GetValue(),
  File "/Users/mwmathis/Documents/DeepLabCut/deeplabcut/pose_estimation_tensorflow/predict_videos.py", line 1464, in convert_detections2tracklets
    trackers = mot_tracker.track(bb)
  File "/Users/mwmathis/Documents/DeepLabCut/deeplabcut/pose_estimation_tensorflow/lib/single_object_tracker.py", line 54, in track
    self.tracks[i].skipped_frames = 0
IndexError: list index out of range
```

I would suggest a warning immediately upon parsing the pickle file, i.e. throw an error that say "this is for single objects/animals only" i.e. 1 individual in the config.yaml (or such)

ehsainit:
@MMathisLab  oh yeah right ! i forgot about that thanks for the note :)
Could you also share the data you used to produce this error so that i confirm that the error was due to _different objects_ and not something else ?

MMathisLab:
will send you an email, thanks @ehsainit !

ehsainit:
Hey @MMathisLab,

It was a small bug in code that I fixed now - the tracker did actually with the videos you sent me. 
I think it could actually be used to track multiple parts of multiple animals. 

Could you test again and give me some feedback ? Thanks !

MMathisLab (approved):

MMathisLab:
Hi @ehsainit so indeed it works to track multiple parts (and actually as it's code now must be multiple parts), but the "issue" is that each point is treated as a single object, i.e. the animals are not assembled, so that is why your tracker is highly similar to ours, but it does not do the "assembly" of the animal. Of course I think having a single tracker is very valuable, i.e. for instances where someone wants to track a "single point" per object, i.e. 1 point per mouse. This could be beneficial for fast tracking, etc. 

Screen Shot 2020-06-23 at 12 21 08 PM

Right now though, I guess the images you post above, aside for the mouse - which is with 2.2 - the single objects are not made with 2.2 (i.e. https://github.com/DeepLabCut/DeepLabCut/pull/736#issue-432531692), i.e it is made with pre-2.2 networks, etc (as related to this PR: https://github.com/DeepLabCut/DeepLabCut/pull/534). As currently this crashes. So, we will work on this part such that 1 point per multi-animal is accepted.

ehsainit:
Hi @MMathisLab, 
great! yes the tracker does track multiple "objects", so it does not care about the type of an object (*instances of each object*) hence the name multiple-objects-tracker. As you know I am a computer science student so I do not know if it would be helpful to track different points on animals separately - if it is not we have also the choice of limiting the tracking process for one object-type. 

Indeed the flies were tracked using DLC1 and a different method to extract loc-maxima from the score map. What crashes currently ? let me know if I can help.

MMathisLab:
Hi @ehsainit sorry for the long pause - right now the issue is that the PAFs assume a skeleton, but if I want to use this as a SO detector, this needs to be adapted in the code. I have some working code, but have yet to debug why the NN is crashing. 

in general the other trackers work on assembled animals, i.e. the general notion is that you (1) predict keypoints, (2) group together appropriate bps to assemble individual animals, and (3) track these individuals; without this you loose the animal ID. Thus, in your case - which is indeed a good case to have ;) - (1) prediction pose (2) track points (i.e. skipping assembly). This is really useful for tracking as you show above, i.e. 1 keypoint per animal/fly/mouse. 

here is some modified code you can try out, which allows you to create the training set efficiently, but still crashing at the NN step. i.e. use a multi-animal video, label only 1 point per `multianimalbodypart` but with >1 `individual` then mod with this, and start training:

```python 

def getpafgraph(cfg, printnames=True):
    """ Auxiliary function that turns skeleton (list of connected bodypart pairs)
        into a list of corresponding indices (with regard to the stacked multianimal/uniquebodyparts)

        Convention: multianimalbodyparts go first!
    """
    individuals, uniquebodyparts, multianimalbodyparts = extractindividualsandbodyparts(
        cfg
    )
    # Attention this order has to be consistent (for training set creation, training, inference etc.)

    bodypartnames = multianimalbodyparts + uniquebodyparts
    lookupdict = {bodypartnames[j]: j for j in range(len(bodypartnames))}
    if len(cfg["multianimalbodyparts"])==1:
        partaffinityfield_graph = []
    else:
        if cfg["skeleton"] is None:
            cfg["skeleton"] = []

        # CHECKS if each bpt is connected to at least one other bpt
        # TODO: check that there is a path leading from each (multi)bpt to each other (multi)bpt!
        connected = set()
        partaffinityfield_graph = []
        for link in cfg["skeleton"]:
            if link[0] in bodypartnames and link[1] in bodypartnames:
                bp1 = int(lookupdict[link[0]])
                bp2 = int(lookupdict[link[1]])
                connected.add(bp1)
                connected.add(bp2)
                partaffinityfield_graph.append([bp1, bp2])
            else:
                print("Attention, parts do not exist!", link)

        unconnected = set(range(len(multianimalbodyparts))).difference(connected)
        if unconnected:
            raise ValueError(
                f'Unconnected {", ".join(multianimalbodyparts[i] for i  in unconnected)}. '
                f"For multi-animal projects, all multianimalbodyparts should be connected. "
                f"Ideally there should be at least one (multinode) path from each multianimalbodyparts to each other multianimalbodyparts. "
                f"Please verify the skeleton in the config.yaml."
            )

    if printnames:
        graph2names(cfg, partaffinityfield_graph)

    return partaffinityfield_graph
```

MMathisLab:
Hi @ehsainit Hope you are well! Let me know if you have any questions, etc!

monajalal:
@ehsainit Hi Tabet, interesting work :) by any chance could you please share the module that applies Kalman Filter to DeepLabCut predictions with jalal@bu.edu ?

ehsainit:
Hey @MMathisLab,
Sorry for taking so long to resolve this issue.

Regarding the NN crush upon setting `partaffinityfield_graph = []`, i was successfully able to fix it by setting `partaffinityfield_graph` in :

https://github.com/ehsainit/DeepLabCut/blob/6d6831046a3babbe8262e9c41609766a4971a85b/deeplabcut/generate_training_dataset/multiple_individuals_trainingsetmanipulation.py#L128

to _false_ (if `partaffinityfield_graph` is empty i.e if there is no need for PAFs ) - ( in case of single object SO tracking - one bp per indiv ).

The training is still in process right now (I am using my CPU to train) ...  it is probably going to take few days :'), no NN crushes up until now !
Screenshot from 2020-11-01 11-56-20

MMathisLab:
Hi All! as you can see there have been a lot of code changes; but we do still want to support SO tracking, so it's slated for 2.2.1 ;). If you want to merge the new code and do so, great - but I also understand it's a big time commitment, so we are happy to do so (and that would be fast for us) in another PR.

ehsainit:
Hey Back !
Due to the limited time i got right now, i would appreciate it if you could do it in new PR :)
I would also be grateful if you can assign me in the new PR though

Best wishes,

PR #748, opened 2020-06-13
Parallelizes creation of labeled videos

smu160:
Hi,

Thank you for making DeepLabCut available. I was processing large batches of videos, and I was curious to see if it would be possible to parallelize this functionality. I used python's multiprocessing package to get a working solution. I've been able to process large batches of videos significantly faster (from running overnight to running in several minutes). Running `create_labeled_videos()` on a single video at a time still works as it did prior to this change.

- Parallelizes create_labeled_videos() using multiprocessing, i.e.,
  each video in a given directory of videos is processed using a
  separate process.

- Tiny formatting changes in accordance with pylint suggestions

- Removed superfluous docstring arg in create_labeled_videos()

All changes and tests were run on a machine running:
OS: Ubuntu 20.04 LTS; kernel version: 5.4.0-37-generic

I ran `testscript.py` after applying these changes, and the test was successful (run-time: 1:20 minutes). 

Thank you!

### DIFF
--- a/deeplabcut/utils/make_labeled_video.py
+++ b/deeplabcut/utils/make_labeled_video.py
@@ -19,6 +19,8 @@ import os
 ####################################################
 import os.path
 from pathlib import Path
+from functools import partial
+from multiprocessing import Pool
 
 import matplotlib.colors as mcolors
 import matplotlib.pyplot as plt
@@ -68,7 +70,7 @@ def CreateVideo(
     displaycropped,
     color_by,
 ):
-    """ Creating individual frames with labeled body parts and making a video"""
+    """Creating individual frames with labeled body parts and making a video"""
     bpts = Dataframe.columns.get_level_values("bodyparts")
     all_bpts = bpts.values[::3]
     if draw_skeleton:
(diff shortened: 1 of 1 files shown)


jeylau:
Hi @smu160! Thank you for the contribution, that's a great addition 😃! It looks good to me, but could you run `black` instead of `pylint` on these files?

smu160:
Hi @jeylau! All done. Thank you!

jeylau (approved):

jeylau:
Looks great to me!

AlexEMG:
Finally speed testing this today! Sorry for the delay

smu160:
@AlexEMG No worries! I actually found an edge case that I didn't account for in my original implementation. Namely, the processes would just process a single video, and then complete. This is a problem in the case that the # of videos to process exceeds the # of cpus on a particular machine (e.g., 24 core machine with 40 videos to process would leave 16 videos unprocessed when running `create_labeled_video()`). 

I also cleaned up the implementation a bit using the multiprocessing Pool API. It gets rid of all the producer/consumer details, and a bunch of other headaches that can occur from the previous implementation. 

I'll push that up so you can review it.

AlexEMG:
Ok, yes that is a good catch!

AlexEMG:
Ok, tested it & find with my 6 CPU machine about fourfold speed improvements! Works great, thanks!

AlexEMG (approved):

PR #751, opened 2020-06-15
doc upgrades

MMathisLab:


AlexEMG:
Looking ok, this includes https://github.com/DeepLabCut/DeepLabCut/pull/736 which should be merged first (otherwise docs deviate)

AlexEMG:
Please DO NOT have spaces in the names of bodyparts, uniquebodyparts, individuals, etc

>> is this true. I mean not ideal, but fine?

PR #783, opened 2020-07-04
Fix typos in multianimal tracking in UseOverviewGuide

neslihanedes:
Boolean flag was reversed for single animal tracking and missing underscores for the function name.

### DIFF
--- a/docs/UseOverviewGuide.md
+++ b/docs/UseOverviewGuide.md
@@ -78,7 +78,7 @@ We highly recommend using 2.2 first in the Project Manager GUI ([Option 3](docs/
 
 ##### *What scenario do you have?*
 
-- **I have single animal videos, but I want to use DLC2.2:**
+- **I have single animal videos, but I want to use the advanced tracking features & updated network capabilities introduced (for multi-animal projects) in DLC2.2:**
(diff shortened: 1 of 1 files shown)


AlexEMG (commented):
It appears that your fix is incorrect, if you want to use DLC 2.2 even with one animal, then you need to pass multianimal=True when you create the project. 

**I have single animal videos, but I want to use DLC2.2:**
     - quick start: when you `create_new_project` just set the flag `multianimal=False`.

Perhaps we can clarify this point. I.e. 

I have single animal videos, but I want to use DLC2.2, because it allows learning limb connects, has more powerful trackers etc.

neslihanedes:
Thanks for the clarification, I was indeed misunderstanding the flag.

So `multianimal=True` means creating a multi-animal project, which itself can also be used for single animal tracking, but then provides the advanced tracking features.

I think this sentence is the most confusing from the docs:
`I have single animal videos, but I want to use DLC2.2`

Actually no matter the provided config, one would use DLC2.2 after all.

Maybe a better phrasing could be:
`I have single animal videos, but I want to use the advanced tracking features introduced in DLC2.2`

WDYT?

I'd be happy to adapt my PR accordingly, but also feel free to close it.

MMathisLab (approved):

PR #793, opened 2020-07-09
Ponies in the zoo ... 

MMathisLab:
This network is trained on data first presented here: https://arxiv.org/abs/1909.11229 (but with added performance, as the upcoming preprint update will reflect).


AlexEMG:
Thanks https://github.com/tbiasi! More details at https://sites.google.com/view/udlworkshop2020/accepted-papers and https://arxiv.org/pdf/1909.11229.pdf (please cite the latter paper when using it!)

PR #802, opened 2020-07-11
Model Zoo & project creation prep for live use

sneakers-the-rat:
a number of small changes to anticipate folks wanting to make projects from pretrained weights/model zoo but without having any video files in the project -- eg. a live case

tried to be minimally impactful and follow naming/etc conventions best i could.

* `new.create_new_project` - allow projects with no videos at time of creation - did this in such a way as to be minimally impactful to existing code -- just don't raise error rather than adding new checks/calling syntax.
* `new.create_new_project` - previously: if a project was new, function returns path to its configuration file. if a project already existed, a warning was printed and `None` was returned. changed so behavior is the same if a project already exists -- return path to config file. this keeps the modelzoo function from choking.
* modelzoo - added an `exportmodel` flag to export a model at the time of download/project/creation. can be called with a boolean or a dict, in which case passed on to `export_model` as kwargs
* modelzoo - don't redownload files if they're already there unless `force` arg is `True`

### DIFF
--- a/deeplabcut/create_project/modelzoo.py
+++ b/deeplabcut/create_project/modelzoo.py
@@ -109,7 +109,8 @@ def create_pretrained_project(
     filtered=True,
     createlabeledvideo=True,
     trainFraction=None,
-    exportmodel: dict = {}
+    exportmodel: [bool, dict] = False,
+    force: bool = False
 ):
     """
     Creates a new project directory, sub-directories and a basic configuration file.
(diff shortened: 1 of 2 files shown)


MMathisLab:
maybe instead we should just expose this otherwise hidden fxn, vs. from the create project steps?
i.e something like `deeplabcut.download_model('full_human', <pathtowhereyouwantItToGo>)`

```python
from deeplabcut.utils import auxiliaryfunctions
deeplabcut.utils.auxfun_models.DownloadModel('full_human', '/content/')
```

PR #810, opened 2020-07-18
Updated refine tracklets playback behavior, hot keys, and help text

backyardbiomech:
## Summary
All in refine tracklets GUI
+ Could not previously get back to 1x forward after fast-forwarding or reversing. No 1x reverse. This PR changes playback behavior to allow 1x reverse, and allows toggling through speed (to return to 1x). Prints "speed" to terminal to show user what speed is playing.
+ In previous PR set "p" key as alternative play/pause key. Turns out that is a built-in matplotlib hotkey to enable/disable pan/zoom tool. Switched play/pause to "space".
+ updated GUI help text

## Updates
+ in `tracklets.py` changed playback behavior. Previously, 1x forward required `speed=''`. If in fast-forward or reverse and opposite direction set, `speed` was set back to blank, but then set to fast-forward/reverse two lines later, so there was now way to get back to 1x forward. Nor was there a 1x reverse playback. Now, `speed` can be `'F'` for 1x forward, `'R'` for 1x reverse. FF/FR keyboard shortcuts add up to 5 F or R, at which point it toggles back to 1x.
  + for example, at load, hitting 'tab' or 'space' starts playing at 1x forward (`speed = 'F'`). Hitting `alt+->` repeatedly toggles from 1x to 2x, 4x, 6x, 8x, 1x, 2x, etc.
  + At any forward speed, hitting `alt+<-` sets speed to 1x reverse.
  + and similar for reverse playback
  + variable `speed` is printed to cmd line (e.g. as `F` or `RR`) so user can see playback speed. 
+ I added `'p'` previously since 'tab' is not captured properly on windows. 'P' is a built-in matplotlib hotkey to toggle the pan/zoom tool on and off, so I switched the play/pause hotkey to 'space'
+ I updated the above in the help text for the GUI, and added the 'P' for pan/zoom.

## Testing

+ All tested on Windows 10


jeylau (approved):

jeylau:
Works smoothly on MacOS too! Thanks @haliaetus13 for the contribution :)

MMathisLab:
we should add these hotkeys, etc into the docs @jeylau 

also, hitting space bar does not stop it for me in MacOS; it continuously plays/reverses

PR #826, opened 2020-07-30
Much (much) faster tracklet parser

jeylau:
...and minor improvements in tracklet association.
The memory usage is now much lower, preventing OOM.


AlexEMG (approved):
Alors

PR #832, opened 2020-08-02
Full macaque

MMathisLab:
adding in the 🐒  full_macaque network, from https://www.biorxiv.org/content/10.1101/2020.07.30.229989v2 !


AlexEMG (approved):
Great!

PR #841, opened 2020-08-05
Creation of Johansson-like, keypoint-only videos

jeylau:
Very fast video creation triggered by `keypoints_only=True` in `create_label_videos()`.
The option to draw the skeleton remains valid.

### DIFF
--- a/deeplabcut/utils/make_labeled_video.py
+++ b/deeplabcut/utils/make_labeled_video.py
@@ -605,9 +605,9 @@ def proc_video(
                     cfg["pcutoff"],
                     cfg["dotsize"],
                     cfg["alphavalue"],
-                    skeleton_color,
-                    color_by,
-                    cfg["colormap"]
+                    skeleton_color=skeleton_color,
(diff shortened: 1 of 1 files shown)


MMathisLab:
Tested on colab modelzoo:

!pip install git+https://github.com/DeepLabCut/DeepLabCut.git@johansson

with horsesideview: 

nice :) love it!

ezgif com-video-to-gif (3)

MMathisLab (approved):

MMathisLab:
if we want to be really extra, the background could be black (or transparent) to really match the Johansson style ;)

jeylau:
> if we want to be really extra, the background could be black (or transparent) to really match the Johansson style ;)

The option is there now: the default is black, but any color can be passed (using the function that is not wrapped in create_labeled_video) 😄

MMathisLab:
nice :)
ezgif com-video-to-gif (4)

AlexEMG (approved):
Allez!

PR #856, opened 2020-08-16
Labeling data no longer causes small keypoint shifts

### LINKED ISSUE #854: small shifting issue with labeling frames
Hi all, 
I have noticed for a time now that there is a very small (but yet annoying) shifting problem with the GUI labeling option. 
Each time I'm labeling some frames, if I go back to some previous frames which I already labeled I see that some of the placements had changed. This is a very small shifting, but yet important for the accuracy of the training. 

(my operating system and dlc version:) Windows 10, Anaconda 3 Env (GPU) & DeepLabCut 2.2b6

Attached are some screenshots of the same labeled frame with the noticeable shifting:

image
image

jeylau:
Previously, the final keypoint coordinates were set to the last position of the mouse cursor, therefore causing a discrepancy between the location the point was dragged to and what underlying data was stored (unless the keypoint was displaced exactly from its center...). Correctly storing the circle center now addresses that issue.
 
Fixes #854.

### DIFF
--- a/deeplabcut/generate_training_dataset/auxfun_drag_label.py
+++ b/deeplabcut/generate_training_dataset/auxfun_drag_label.py
@@ -134,7 +134,7 @@ class DraggablePoint:
             self.point.set_animated(False)
             self.background = None
             self.point.figure.canvas.draw()
-            self.final_point = (event.xdata, event.ydata, self.bodyParts)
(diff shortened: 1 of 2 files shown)


AlexEMG (approved):
Looks good, didn't test.

PR #859, opened 2020-08-20
Duplicate keypoints from the previous frame

### LINKED ISSUE #849: Copy labels from previous image
Hi, we label a lot of images, and on some of them the markers barely move or don't move at all during some portions of movements, e.g. shoulder during finger manipulation. For that, I have written a small routine that checks several previous frames and copies the marker position from the previous image if the marker has not moved much.
This was inserted into generate_training_dataset/labeling_toolbox.py:572
```
        # copy over labels
        is_this_labeled = any([not np.isnan(self.dataFrame[self.scorer][bp]['x'].values[self.iter])
                               for bpindex, bp in enumerate(self.bodyparts)]) 
        num_prev_frames = 3
        pixel_distance_threshold = 10
        if self.iter > 1 and not is_this_labeled:
            start_copy_frame = max(1, self.iter-num_prev_frames)
            prev_labeled_bps = []
            prev_labeled_bpindexes = []
            for bpindex, bp in enumerate(self.bodyparts):
                if not np.isnan(self.dataFrame[self.scorer][bp]['x'].values[self.iter-1]):
                    prev_labeled_bps.append(bp)
                    prev_labeled_bpindexes.append(bpindex)
            for bpindex, bp in zip(prev_labeled_bpindexes, prev_labeled_bps):
                xs = []
                ys = []
                for frame_index in range(start_copy_frame, self.iter):
                    xs.append(self.dataFrame[self.scorer][bp]['x'].values[frame_index])
                    ys.append(self.dataFrame[self.scorer][bp]['y'].values[frame_index])
                if not any(np.isnan(x) for x in xs) and (np.std(xs)**2+np.std(ys)**2)**0.5 < pixel_distance_threshold:
                        self.rdb.SetSelection(bpindex)
                        event2 = wx._core.CommandEvent()
                        event2.xdata = self.dataFrame[self.scorer][bp]['x'].values[self.iter-1]
                        event2.ydata = self.dataFrame[self.scorer][bp]['y'].values[self.iter-1]
                        event2.button = 3
                        self.onClick(event2)
```
It has several parameters that should be available to the user, like:
`num_prev_frames` - number of frames to be checked for marker movement;
`pixel_distance_threshold` - allowed 2D deviation of the marker position for the marker to be considered "not moving".

It calls OnClick event to place the markers, which makes them appear gradually on the screen, and some users liked that as a visual indication of the procedure.

I am not sure of the best way to integrate it into DLC config, so I am just leaving it here as a suggestion.

jeylau:
When labeling frames, hitting Ctrl+C (or Command+C on MacOS) now duplicates the keypoints from the previous frame. With multi-animal DLC, only the keypoints of the animal currently selected are affected.

Closes #849.


AlexEMG:
Would also be great if @nishbo could have a look!

nishbo:
This looks great, glad my suggestion worked for you! I think it would be nice to have an option for "smart" duplication of the labels, that measures their variability and copies only stable ones, e.g. on Ctrl+Shift+C. With default parameters that can be overwritten in config.
Also, I was surprised that users enjoyed labels appearing gradually, because I had to use OnClick call. They said they could take time to evaluate the correctness of each one as it appeared, but that is more of a flavoring, so it is up to you.

AlexEMG:
I tested the code for single animal GUI! It's great. I think ready to commit, but I put a reference in the help of the GUI to the docs and suggest we also make table of hot keys for the labeling GUI. I started, see **docs/functionDetails.md** 

HOT KEYS IN THE Labeling GUI:
- Ctrl + C: Copy labels from previous frame.

AlexEMG:
What other hotkeys should be added?

jeylau:
> What other hotkeys should be added?

None besides the arrow keys (left/right: frame navigation; up/down: prev/next keypoint) and delete, but these are obvious :)

MMathisLab:
>>What other hotkeys should be added?

>>None besides the arrow keys (left/right: frame navigation; up/down: prev/next keypoint) and delete, but these are obvious :)

- wait, but those are already in the main GUI, no?

MMathisLab:
maDLC:
- note, cannot copy from current frame to prior frame. 
- (Frames must be labeled in order; odd bugs occur otherwise).
- ONLY ANIMAL # 1 is copied:
ezgif com-video-to-gif (5)

- odd box inside GUI; unstable behavior of the matplotlib legend/keys ^ see video above (but also seems to be on current master for me as well). 

>>> matplotlib.__version__
'3.3.0'

jeylau:
> > > What other hotkeys should be added?
> 
> > > None besides the arrow keys (left/right: frame navigation; up/down: prev/next keypoint) and delete, but these are obvious :)
> 
> * wait, but those are already in the main GUI, no?

Yes yes, I just wasn't sure they were in the docs already.

jeylau:
> maDLC:
> 
> * note, cannot copy from current frame to prior frame.
> * (Frames must be labeled in order; odd bugs occur otherwise).
> * ONLY ANIMAL # 1 is copied:
>   ezgif com-video-to-gif (5)
> * odd box inside GUI; unstable behavior of the matplotlib legend/keys ^ see video above (but also seems to be on current master for me as well).
> 
> > > > matplotlib.**version**
> > > > '3.3.0'

For multi-animal scenarios, only the current selected animal gets copied. You'd need to change the radio button to animal2 to copy the second mouse. I thought this was better than copying the whole scene, but I can change this if you think that's not intuitive :)

MMathisLab:
okay makes sense! I think the HELP button should clarify this; I added!

MMathisLab (approved):

AlexEMG (approved):

PR #870, opened 2020-08-25
Added delete button in labelling_toolbox screen

anandcpanchal:
- Added functionality to delete frame during labeling
image
- On use, Deletes image from the labeled-data folder and marks all NaN in self.dataFrame
- During save data method execution, ignores data frame rows with all NaN entries.
- Suggested to use the 'Save' button at the end of labeling

Kindly ignore it if it's irrelevant!
Thanks


deeplabcut/generate_training_dataset/labeling_toolbox.py
```diff
@@ -867,14 +873,51 @@ def saveEachImage(self):
...
+	    """
+	    Reset data for each image
+	    """
+	    for idx, bp in enumerate(self.updatedCoords):
+	        self.dataFrame.loc[self.relativeimagenames[self.iter]][
+	            self.scorer, bp[0][-2], "x"
+	        ] = None
+	        self.dataFrame.loc[self.relativeimagenames[self.iter]][
+	            self.scorer, bp[0][-2], "y"
+	        ] = None
```
jeylau: Please fix the indentation.

deeplabcut/generate_training_dataset/labeling_toolbox.py
```diff
@@ -867,14 +873,51 @@ def saveEachImage(self):
...
+    
+
+    def deleteImage(self, event):
+    	image_path = os.path.join( self.currentDirectory, self.dataFrame.index[self.iter])
```
jeylau: I would use self.dataFrame.loc[self.relativeimagenames[self.iter]] like you did above rather than from the DataFrame index directly. This protects us against cases (we recently had) of images named in a way other than imgXXX, which resulted in the ascending order being different in the pandas index and the list of image names.

deeplabcut/generate_training_dataset/labeling_toolbox.py
```diff
@@ -867,14 +873,51 @@ def saveEachImage(self):
...
         """
+
+        # Remove previous save
+        csv_path = os.path.join(self.dir, "CollectedData_" + self.scorer + ".csv") 
+        if os.path.exists( csv_path):
+        	os.remove( csv_path)
+
+        hdf_path = os.path.join(self.dir, "CollectedData_" + self.scorer + ".h5")
+        if os.path.exists( hdf_path):
+        	os.remove( hdf_path)
```
jeylau: Is that necessary, given that data are overwritten by default? I'm worried that if something fails—although I admit this shouldn't happen—between that line and to_csv()/to_hdf(), the user is left with no data.

deeplabcut/generate_training_dataset/labeling_toolbox.py.orig
```diff
@@ -0,0 +1,923 @@
+"""
```
jeylau: You can delete this one 😊.

jeylau (changes requested):
Thanks a lot for the contributionI have just made a couple of comments. Any chance you implement the button for the multi-animal labeling GUI too? 😄

anandcpanchal:
> Thanks a lot for the contributionI have just made a couple of comments. Any chance you implement the button for the multi-animal labeling GUI too?

Yes, Sure!

MMathisLab:
I think it should be clear it deletes the frame, not all that has been labeled (like a reset). Can it read "Delete Frame" ?

deeplabcut/generate_training_dataset/labeling_toolbox.py
anandcpanchal: Have replaced it with backup file creation for extra safety, eventually can add restore function as well

anandcpanchal:
I have found (and resolved for my use case at least) one more thing which hinders label visualization in cross-platform case.
e.g. My data labeled in windows environment can not be reworked in Linux ( due to path separators difference '\' in windows and '/' in Linux), I have fixed it by adding some code lines, should I push that as well?

PR #905, opened 2020-09-13
Standardised references to config_path

PolarBean:
super minor commit which just standardises all references to the config file to allow for easier copy/pasting


MMathisLab:
thanks for the contribution!

PR #923, opened 2020-09-24
Update maDLC_AdvUserGuide.md

backyardbiomech:
One thing I have been confused about since moving to maDLC has been how to extract outlier frames from analyzed video (as was typical in the single animal workflow) to use the refine labels toolbox. I would convert to tracklets, but need better detection before the refine tracklets GUI would work. For example, I could run the refine tracklets GUI, but some points would be non existent so I couldn't refine them. I was stuck just manually extracting frames without the efficiency and power of the extract outliers functions.  I couldn't find anything in the documentation, and I believe others have had the same problem based on posts on gitter and image.sc. I've tried to add basic instructions (including a GUI workaround for less experienced users, and a mention of the command line call for advanced users).

### DIFF
--- a/docs/maDLC_AdvUserGuide.md
+++ b/docs/maDLC_AdvUserGuide.md
@@ -459,6 +459,10 @@ Please note that you do **not** get the .h5/csv file you might be used to gettin
 
 If this does not look good, we recommend extracting and labeling more frames (even from more videos). Try to label close interactions of animals for best performance. Once you label more, you can create a new training set and train. 
 
+You can either:
+1. extract more frames from existing or new videos and label as when initially building the training data set, or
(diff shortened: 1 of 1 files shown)


MMathisLab:
Hi @backyardbiomech great point! 

We had it on our internal #todo list to make a proper function for this, ie. dlc.extract_outliers_from_pickle (or such), but agree we should prioritize this (cc @jeylau @AlexEMG). In the meantime, this doc upgrade is great, thanks so much! 

Namely, we should have 2 quality checks here:

Group 18

MMathisLab (approved):

jeylau:
Looks good to me! I would add that one can use that function https://github.com/DeepLabCut/DeepLabCut/blob/b1f8f413ec8f5ffdb522624b62a02b5980472f66/deeplabcut/utils/make_labeled_video.py#L915 to evaluate tracking quality before refining the tracklets.

MMathisLab:
@jeylau in the docs I put this directly above, is the other call better to ask people to use?

```python
scorername = deeplabcut.analyze_videos(config_path,['/fullpath/project/videos/testVideo.mp4'], videotype='.mp4')	
deeplabcut.create_video_with_all_detections(config_path, ['/fullpath/project/videos/testVideo.mp4'], scorername)	
```

jeylau:
@MMathisLab, no no that's great, I really like `deeplabcut.create_video_with_all_detections` as it is very fast, and so should be run systematically :)

PR #1034, opened 2020-12-13
Circleci project setup -- simple testscript.py

MMathisLab:
this PR adds a simple "testscript_cli.py" -- tests create project, extract frames (kmeans), create training set, train, eval., export. There are some time limits on circle cli I am working to resolve, but by in large this "as is" is a good addition. Every merge at least runs the testscript!


AlexEMG (approved):
Thanks @MMathisLab

PR #1061, opened 2021-01-05
Patch camera triangulation in multi-animal projects

jeylau:


AlexEMG (approved):

MMathisLab:
why do we drop destfolder? This would break anipose @jeylau @AlexEMG

jeylau:
@MMathisLab because I had the argument marked as unused.

PR #1096, opened 2021-01-28
Use keyword argument for `inferenceutils.assemble_individuals`

sugyan:
When executing `deeplabcut.convert_detections2tracklets()` with `printintermediate=True`, it raises `KeyError`, since it called `inferenceutils.assemble_individuals()` with 11 arguments and `printintermediate` is the 11th argment but `inferenceutils.assemble_individuals` now receives 12 arguments and `print_intermediate` as the 12th argment!
I fixed calling arguments of `inferenceutils.assemble_individuals` using *keyword argument*


jeylau (approved):

jeylau:
Thanks @sugyan for the bug fix!

Issue #1124, opened 2021-02-23
pip install downloads many versions of opencv (with 3.8 python)


svenkreiss:
After running `pip install deeplabcut`:

<img width="578" alt="image" src="https://user-images.githubusercontent.com/2775078/108826140-3c431080-75c4-11eb-8d2e-83451b864c13.png">

Versions:
* pip 21.0.1
* python 3.8.6

svenkreiss:
Same with the latest beta: `pip install deeplabcut==2.2b8`

<img width="583" alt="image" src="https://user-images.githubusercontent.com/2775078/108826807-1833ff00-75c5-11eb-9b6a-d8aa1ad9649d.png">

MMathisLab:
Hi @svenkreiss yes, this is with 3.8; if you use 3.7 (and the DLC conda files) it will force you to use 3.7, and then this does not happen :)

MMathisLab:
Hey @svenkreiss I should add, right now tensorflow 1.x is not supported in 3.8 (is my understanding at least(, and currently we require TF 1.x; we do have a dlc-core branch that is for 2.x support, but without any GUI support, etc: https://github.com/DeepLabCut/DeepLabCut-core

https://www.tensorflow.org/install/pip

![Screen Shot 2021-02-25 at 3 09 02 PM](https://user-images.githubusercontent.com/28102185/109165144-83bfcd00-777b-11eb-83de-5f32089d5ec3.png)

svenkreiss:
Thanks for the extra info. I'll wait until Python 3.8 is supported.

MMathisLab:
^ yeah, I was hoping for 3.8 TF1 support; it was initially promised in Dec, I have low hopes now... 

We are currently working on bumping up to TF2 (as is in the core), but some models are not supported, so that limits enthusiasm for a full roll over to TF2.

MMathisLab:
Hi @svenkreiss `pip install deeplabcut-core` now supports TF2, so 3.8 is more doable ;). Hope that helps!

PR #1126, opened 2021-02-26
3.8* support / dependency clean up [WIP]

MMathisLab:
a few dependencies are drop as they were dep of dep; note moviepy is also dropped, but this might be required (seems not required to run `testscipt.py` and `testscript_multianimal.py` which pass successfully). 

- note, this also includes updating the conda files to TF 1.15.5 vs. 1.13.1; then numpy 1.17.3 is supported/no warnings.
- this has only been tested on MacOS; I would ask @AlexEMG to test on ubuntu and @jeylau on windows before committing. 
- also only tested on CPU MacOS; GPU support should be tested w/TF 1.15.5
- note 3.7 is still required, given the TF limitations given in issues #1124, but at least now DLC package does build in 3.8; meaning we are getting away from pip dependency issues :)


MMathisLab:
@jeylau this is failing now: https://github.com/DeepLabCut/DeepLabCut/runs/1995719243#step:6:14

jeylau:
Yep, we have to install deeplabcut. But I'm looking for ways around it.

MMathisLab:
okat @jeylau I added back moviepy and this is passing. I didn't go through all your changes (it's a lot ;P); can you confirm this runs all the test scripts? (and run black)

MMathisLab:
@jeylau TF 1.x is not supported in 3.8 python ;/. we we can only test the installation of DLC w/o TF https://www.tensorflow.org/install/pip#system-requirements

MMathisLab:
we should run black ;) when I ran it on your edits it required 40 files to be changed btw!

jeylau:
Aw I forgot! :/ That's something I wanted to add to the workflow as well...

PR #1203, opened 2021-04-22
Release GPU memory after task completion with numba.cuda

alexcwsmith:
This PR takes advantage of numba.cuda.close() to release GPU memory upon completion of training, evaluation, or analysis of videos. Very minimal additions, no additional dependencies, no docstring changes.

Tested on platform:
Ubuntu 18.04
Nvidia driver 450.102.04
cudatoolkit 10.2.89 (installed via anaconda)
cudnn 7.6.5 (installed via anaconda)

I do not get an error when i run numba.cuda.close() in a console with no open cuda processes, so I don't think this should cause any issue for those running on a CPU, but I suppose that is something to check.


alexcwsmith:
Not sure why the above checks failed for all python versions <3.8, looks like numba couldn't get the cuda driver for lower versions. FWIW I'm running 3.7 and do not have an issue.

alexcwsmith:
Ah I just read those checks again and it looks like this indeed would cause an issue for CPU-only users if there is no CUDA installed. I will make a new optional boolean argument in these functions in these functions that defaults to False and only runs uses numba to release GPU memory if set to True, and update the docstrings. Having the GPU memory get released when these functions finish without the user having to kill the entire console/process would be _really_ nice for those of us who manage workstations that several people/labs use.

alexcwsmith:
Looks like the 2nd commit (bc907e9) did the trick. I'm not having any issues at all running in an ipython console in the Spyder IDE, but a colleague mentioned that in a jupyter notebook his kernel died when he started dlc.analyze_videos() right after evaluation with term_gpu=True. 

From looking at the docs of numba.cuda.close() it might be more surprising that I am not having that issue in Spyder, since it does seem that a single process/console may not be able to start new cuda threads after using cuda.close() within that process. For my own situation that is still far preferable to functions like evaluate and analyze_videos holding onto GPU memory until the user kills them, but I'll leave it up to you whether you agree that is a benefit. I think the kernel dying issue can be resolved with numba but I won't have a chance to look into that much more for a while.

MMathisLab:
hey @alexcwsmith thanks! Could you delineate why you needed this addition? we have not experienced memory issues/not closing the TF session properly such that it would hold memory.

backyardbiomech:
This would be a big help for me. I frequently run into problems with multiple users. For example, a student can start a training session that takes all of the gpu memory and let’s it run overnight. It finished by morning, but my student isn’t at the computer. When I log in on my user account in the morning and try to analyze videos, I get memory allocation errors. The only fix is to wait for my student, or to start killing processes. But I can’t actually confirm that the training session is completed (though I’ve found that gpu fan speed is a decent proxy), so killing processes is a gamble.

This isn’t a problem within a single Python instance - the memory simply gets reallocated, but is never released according to task manager. But with multiple Python instances either in a single user account or different user accounts, the memory isn’t reallocated, creating this headache.

alexcwsmith:
Hey @MMathisLab of course! We run into issues very similar to what @backyardbiomech experiences. My headache isn't so much with training, but the ops that don't have the allow_growth parameter (evaluation/analysis) claim all GPU memory (we now bypass this by setting os.environ variable to force TF to allow growth), and don't take nearly as long, but after completion the GPU memory is held as long as the kernel is alive. So yeah I have to start killing processes, which is mostly just a nuisance since I don't want to give sudo privileges to the students etc that remote into our workstation.

MMathisLab:
okay, makes sense! Thanks! We are doing to do a big push in the next 24 hours, then can work this in/test afterwards!

alexcwsmith:
Cool thank you! The few people I've showed and talked to all had similar feedback that it would be super helpful, but I think it's pretty specific to those of us who have multiple users on one workstation. The issue we've run into now is that after using numba.cuda.close(), for example at the end of evaluation, trying to start analysis on the same kernel will result in the kernel dying. I have been out of town all week (writing this on iOS) so haven't checked, but I think that can be solved with numba.cuda.select_device(gputouse) at the beginning of the following function.

MMathisLab (approved):
given the interest from others, I think this is a good mod; I have not tested it though; so it should still be tested on a GPU-system in a scripted manner, namely, @alexcwsmith could you run the examples/testscript.py and testscript_multianimal.py on your system and attached logs/terminal output?

jeylau:
Agreed! This looks very useful and all tests seem to pass. I'm concerned about adding some more arguments though; are there cases where emptying memory is not desirable? Wouldn't a `try/except numba.cuda.cudadrv.error.CudaSupportError` be preferable?

alexcwsmith:
Hey @jeylau I will definitely defer to y'alls expertise on this, I'm not super experienced with Numba. I have been working my way through the docs trying to figure out what the best way to do this is, and had not gotten as deep as ```numba.cuda.cudadrv.error.CudaSupportError```, so thank you very much for that! From my quick look at those docs, I think that does sound like the better way to go. This week is a surgery week, but I will get back to this and try to implement that in the next few days. And thanks @MMathisLab, I was low-key worried this was going to be shot down as a terrible idea, so I'm glad there is general interest from others. I'll run the test scripts as soon as I get a chance and post the outputs.

MasterOnCard (commented):
Look to Tap

hmaarrfk:
I added similar code in our workflow only to remove it quickly.

We were doing
1. Inference on GPU using DLC + Tensorlfow 2.7
2. Freeing the GPU memory with numba+cuda
3. Inference on the GPU using DLC

But we were experiencing crashing on step 3.

I'm not sure if others can confirm this or not. But I felt like it was related to my use of numba + cuda so pulled the change. For inference, one can limit the GPU memory usage (to allow a few applications to run in parallel) but I understand that this may not be possible during training.

alexcwsmith:
@hmaarrfk The way we've implemented this, if someone is going to be staying at the computer and running multiple jobs in serial, we just leave
```
term_gpu=False
```
and there is no issue with the kernel dying. We only set the term_gpu argument to True when running inference on a very large batch of videos... jobs that are going to take 12-24 hours, so the GPU will be shut down as soon as the job is finished regardless of whether someone is paying attention to the computer. I haven't spent much time looking for a workaround that allows TensorFlow to re-start the GPU in the same kernel after using numba.cuda.close(), but that argument defaults to False so it hasn't been an issue for us.

PR #1239, opened 2021-05-13
Improved docs wrt tracklet stitching

### LINKED ISSUE #1236: stitch_tracklets default animal names break create_labeled_video function
Thanks for the amazing work first of all! The new maDLC is quite amazing, more accurate, more network choices. Fantastic work team! I did encounter one challenge when using it, hence this feature request.

In the beta version of maDLC, one could name the individuals in the config file as pleased. This would then be used as the individual names in the h5 file and would lead to no issues when creating a labelled video.

In the new version of maDLC, the stitch_tracklets method by default assigns ids as [ind1,ind2,...]. Then when trying to create a labelled video nothing is displayed as the individuals names in the h5 file and config files do not match and thus the list 'individuals' (which comes from individuals = auxfun_multianimal.IntersectionofIndividualsandOnesGivenbyUser(cfg,displayedindividuals)) becomes empty. Thus when slicing the dataframe from the h5 file, this also becomes empty and nothing is plotted on the frames.

Ideally, by default the stitch_tracklets function should just take the individuals names from the config file. Not sure if this was implemented on purpose this way, but at least for me, it caused some confusion and would be nice if it can be avoided.
If necessary I can make a pull request to ease your workload.

**Describe alternatives you've considered**
Perhaps in the maDLC userguide you could emphasize that if custom names are used (instead of ind_i) then these should also be specified when stitching tracklets.

jeylau:
Fixes #1236

### DIFF
--- a/deeplabcut/gui/refine_tracklets.py
+++ b/deeplabcut/gui/refine_tracklets.py
@@ -313,9 +313,9 @@ class Refine_tracklets(wx.Panel):
 
     def create_tracks(self, event):
         deeplabcut.stitch_tracklets(
+            self.config,
             self.datafile,
             n_tracks=self.ntracks.GetValue(),
-            animal_names=self.cfg["individuals"],
         )
 
     def refine_tracklets(self, event):
--- a/deeplabcut/refine_training_dataset/stitch.py
+++ b/deeplabcut/refine_training_dataset/stitch.py
@@ -7,6 +7,7 @@ import re
 import scipy.linalg.interpolative as sli
 import warnings
 from collections import defaultdict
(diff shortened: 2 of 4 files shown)


AlexEMG (commented):
Don't you think its better to just pass config (as for other functions) and always do this by default?

jeylau:
> Don't you think its better to just pass config (as for other functions) and always do this by default?

Fixed!

AlexEMG (approved):
In general good! I'm wondering what happens if you do inference on a video with more individuals than in the config file? In this case one probably needs the possibility to overwrite the name of the tracks?

PR #1243, opened 2021-05-14
fix two nasty bugs

sin-mike:
1. ffmpeg gets paths as quoted to handle spaces
2. os.path.split instead of .rsplit to handle non-parented files

### DIFF
--- a/deeplabcut/generate_training_dataset/trainingsetmanipulation.py
+++ b/deeplabcut/generate_training_dataset/trainingsetmanipulation.py
@@ -559,7 +559,9 @@ def MakeInference_yaml(itemstochange, saveasconfigfile, defaultconfigfile):
 
 def _robust_path_split(path):
     sep = "\\" if "\\" in path else "/"
-    parent, file = path.rsplit(sep, 1)
+    #import pdb; pdb.set_trace()
+    #parent, file = path.rsplit(sep, 1)
+    parent, file = os.path.split(sep)
     filename, ext = os.path.splitext(file)
     return parent, filename, ext
 
--- a/deeplabcut/generate_training_dataset/trainingsetmanipulation.py
+++ b/deeplabcut/generate_training_dataset/trainingsetmanipulation.py
@@ -559,7 +559,14 @@ def MakeInference_yaml(itemstochange, saveasconfigfile, defaultconfigfile):
 
 def _robust_path_split(path):
     sep = "\\" if "\\" in path else "/"
-    parent, file = path.rsplit(sep, 1)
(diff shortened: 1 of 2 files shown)


AlexEMG (approved):
Thanks!

AlexEMG:
Tests -- fail.

jeylau:
Hey @sin-mike, handling spaces in movie file names is a great fix, but why escaping the quotes?
As for path splitting, what specifically are you trying to fix? Splitting the separator is not doing anything.

sin-mike:
> Hey @sin-mike, handling spaces in movie file names is a great fix, but why escaping the quotes?
> As for path splitting, what specifically are you trying to fix? Splitting the separator is not doing anything.

@jeylau 
1. Escaping quoted makes later changes safer, e.g. if you change the quotation style later automatically. It is just a good practice to explicitly put `\"` instead of just `"`, when you want it to by printed
2. Os-specific filepath parsing is handled by _os.path_ by default. By using os.path.split, instead of .rsplit(os.path.sep) helps to avoid such a problem, as when you put in ipython simply `movie.mp4`, you'll get `['movie.mp4']` with `rsplit`, but `['', 'movie.mp4']` with `os.path.split`

sin-mike:
@AlexEMG 

> Tests -- fail.

I didn't get it, how one half of env passed the test, the other didn't. How to disclose the errors?

jeylau:
os.path.split handles splitting according to the separator of the platform the code is run on, and thus does not work e.g. with data labeled on Windows prior to creating datasets/training on Linux. Note that this function is not part of the public API, and is only called in places where full paths are expected; while I agree it fails when only a file name is given, this is not what's stored in the annotated data files, nor in config['video_sets']. I hope that helps 😊

AlexEMG:
Indeed not a public function. @jeylau  Independent for this conversation, but maybe for future maintainability it would be good to actually put a comment, why we split the paths like this (to all cross-platform data transfer).

sin-mike:
> os.path.split handles splitting according to the separator of the platform the code is run on, and thus does not work e.g. with data labeled on Windows prior to creating datasets/training on Linux. Note that this function is not part of the public API, and is only called in places where full paths are expected; while I agree it fails when only a file name is given, this is not what's stored in the annotated data files, nor in config['video_sets']. I hope that helps 😊

Yep, It is all clear. Then a quick and dirty fix will be checking if len of `rsplit` is more than 2, and giving back an empty parent folder otherwise.

jeylau (approved):

jeylau:
Thanks a lot for the edits @sin-mike, looks greater now! ☺️

PR #1251, opened 2021-05-17
Smarter inclusion of identity information and tests

jeylau:
- [x] Fix missing identity information in tracklets
- [x] Allow animal assembly with identity only
- [x] Automatically stitch with identity when possible
- [x] Allow tracking with identity only (optimal ID assignment based on soft voting of body part identity predictions)


AlexEMG (approved):
Great

MMathisLab (approved):

PR #1257, opened 2021-05-19
(Much) faster multi-animal dataset creation

jeylau:
- Tested against the previous function w/ and w/o a single individual
- Fixed slight mismatch between desired and actual train fraction when cropped images had non-unique filenames
- Runs at ~23k FPS on a huge (300k+ rows) dataset


AlexEMG (dismissed):
Let's put the data here: https://github.com/DeepLabCut/UnitTestData

PR #1261, opened 2021-05-20
Add cuda toolkit 10.0 dependency to GPU yaml file

hmaarrfk:
I believe conda-forge built cudnn=7 with both cuda toolkit 10.0 and cuda toolkit 10.1. Therefore you have to specify to use the appropriate version of cuda toolkit otherwise the conda solver will default to 10.1

I believe this is what you need, but I'm not too sure.

Let me know.

### DIFF
--- a/conda-environments/DLC-GPU.yaml
+++ b/conda-environments/DLC-GPU.yaml
@@ -21,6 +21,7 @@ dependencies:
   - python=3.7
   - pip
   - cudnn=7
+  - cudatoolkit=10.0
   - jupyter
   - nb_conda
   - Shapely


MMathisLab:
Hi @hmaarrfk thanks for the contribution, but for CUDA, we feel it's best to let the user install this first CUDA and the driver; as we note here: https://github.com/DeepLabCut/DeepLabCut/blob/master/docs/installation.md#gpu-support

hmaarrfk:
hmm. ok. I kinda understand but the instruction is as simple as install the Nvidia drivers (460 as is the case for ubuntu 20.04) and use the conda command above. 

the challenge is that the tensor flow package you pin to is quite old (mostly because conda and GPU stuff moves so fast) and new cudatoolkit packages have been made available since.

finally i don't think we package tensor flow GPU for conda forge.

MMathisLab:
agree, and we are moving to TF within the month (see [blog](http://www.mackenziemathislab.org/deeplabcutblog/2020/11/23/rolling-up-to-tensorflow-2) and [note here on roadmap](https://github.com/DeepLabCut/DeepLabCut/blob/master/docs/roadmap.md)). Which means cuda 10 is then too old as well. :)

hmaarrfk:
ok. I'll focus my packaging efforts on TF2 then. once we have that with GPU support on conda forge it should be breezy

PR #1263, opened 2021-05-21
Faster video inference

jeylau:
- Local peak extraction is moved to GPUs
- Highly-vectorized edge affinity computation (+ corrected trapezoidal integration)
- Computation with float32
- Big clean-up
- ~4x inference speed up

Further improvements could also be made with threaded video I/O, but I'll leave this for a future PR.


MMathisLab (approved):

AlexEMG (commented):
in the default:
locref_smooth: false >> isn't that more of a parameter for test/pose_cfg.yaml

AlexEMG (approved):

PR #1318, opened 2021-06-10
Support images that are smaller (both sides or one side) than the desired crop size

yeshaokai:
Handle the case where the image is smaller (both side or one side) than the desired crop size


AlexEMG:
Looks great to me -- @jeylau can you please do some functional test.

jeylau:
Tested and looking great @yeshaokai!

jeylau (approved):

PR #1337, opened 2021-06-18
added opencv based video check_integrity_robust

backyardbiomech:
### Description

Many users experience issues with "corrupted" video files. The current `check_integrity` function uses ffmpeg to check for corruption, but deeplabcut depends on opencv for video loading in some functions.  Opencv is notoriously not robust against certain types of video file corruption, so while `check_integrity` and ffmpeg may not log problems with specific video files, functions depending on opencv may still fail.

This PR adds an opencv-based `check_integrity_robust` to the VideoReader class that uses opencv to load the video frame-by-frame until either the end of the video is reached or opencv fails to read a frame. This will take a long time for long videos, but is the only way I can think (other than simply directing users to re-encode the video as trial-and-error) to demonstrate an opencv-dependent corruption problem.

### Usage

```
>>>from deeplabcut.utils.auxfun_videos import VideoReader
>>>vid = VideoReader('path/to/video')
>>>vid.check_integrity_robust()
```

### Tests

I have tested on macOS 10.15.7, but don't have any video files that I know are corrupted to test against. It logs no error for uncorrupted videos.

### DIFF
--- a/deeplabcut/utils/auxfun_videos.py
+++ b/deeplabcut/utils/auxfun_videos.py
@@ -42,6 +42,15 @@ class VideoReader:
         if os.path.getsize(dest) != 0:
             warnings.warn(f'Video contains errors. See "{dest}" for a detailed report.')
 
+    def check_integrity_robust(self):
+        numframes = self.video.get(cv2.CAP_PROP_FRAME_COUNT)
+        fr = 0
+        while fr < numframes:
+            success, img = self.video.read()
+            if success == False:
+                warnings.warn(f'Opencv failed to load frame {fr}. Use ffmpeg to re-encode video file')
+            fr += 1
+
     @property
     def name(self):
         return os.path.splitext(os.path.split(self.video_path)[1])[0]
--- a/deeplabcut/utils/auxfun_videos.py
+++ b/deeplabcut/utils/auxfun_videos.py
@@ -45,9 +45,9 @@ class VideoReader:
     def check_integrity_robust(self):
         numframes = self.video.get(cv2.CAP_PROP_FRAME_COUNT)
         fr = 0
-        while self.video.isOpened():
-            success, img = self.video.read()
-            if success == False:
+        while fr < numframes:
+            success, frame = self.video.read()
+            if not success or frame is None:
                 warnings.warn(f'Opencv failed to load frame {fr}. Use ffmpeg to re-encode video file')
             fr += 1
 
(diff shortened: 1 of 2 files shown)


deeplabcut/utils/auxfun_videos.py
```diff
@@ -42,6 +42,15 @@ def check_integrity(self):
         if os.path.getsize(dest) != 0:
             warnings.warn(f'Video contains errors. See "{dest}" for a detailed report.')
 
+    def check_integrity_robust(self):
+        numframes = self.video.get(cv2.CAP_PROP_FRAME_COUNT)
```
jeylau, line 46: Since opencv can also return the wrong number of frames with corrupted metadata, what do you think about the following?
```python
while self.video.isOpened():
   ...
```

backyardbiomech:
I like it, Jessy.

jeylau:
Great contribution! Thanks so much, @backyardbiomech 😊

jeylau (approved):

AlexEMG:
This is great guys, maybe we make a DLC function? 

`deeplabcut.utils.check_video_integrity('path/to/video')`

that basically calls: 
```
from deeplabcut.utils.auxfun_videos import VideoReader
vid = VideoReader('path/to/video')
vid.check_integrity_robust()

```
Then we could just ask users to run this function (when it looks like a video corruption error)?

jeylau:
@backyardbiomech, you want to give it a shot? 😊

backyardbiomech (commented):
I just tested and this doesn't work. Opencv doesn't close the file automatically when it reaches the end, so this leads to an endless file read. 
The typical strategy is to look for "success" or an empty frame, which can also be a source of  the corruption. So I think the best thing to do is to call the ffmpeg based test to check metadata and basic formatting, then go back to looping through the numframes to make sure every frame within the numframes is successfully loaded (and has information in it).
I'm putting together a PR now.

backyardbiomech:
@jeylau I think you should be able to see the changes I just made. When calling the "main" function `deeplabcut.utils.check_video_integrity('path/to/video')`, it first calls the ffmpeg-based check, then then opencv-based check, looking for both failed frame loads and/or empty frames. If the first failed frame load (or first empty frame) coincides with the metadata's number of frames, then we can assume the video is not corrupt.

jeylau:
I've just added very minor edits, but looks great to me! Thanks again @backyardbiomech! ☺️

jeylau (approved):

AlexEMG (approved):

PR #1391, opened 2021-07-08
Update requirements.txt

MMathisLab:
unpin TF to 2.5; just greater than 2; also unpin matplotlib; evaluation plotting tested, and label GUI tested, working and plot bars in the correct order.


jeylau:
@MMathisLab, do the images look great now when saved during evaluation?

MMathisLab:
*same as with 3.1.3

Issue #1398, opened 2021-07-09
AttributeError: module 'tables' has no attribute 'exceptions'


CorinneAKelly:
I am on a Windows 10 PC, DLC 2.2rc3 in an anaconda env.

I am trying to use the refine tracklets tab in the GUI. Originally, I was getting the error that "tables" didn't exist, and to pip install it. However, when I tried that it simply said it was already there. Someone on another thread mentioned that they had to uninstall pytables then pip install tables, so I tried this. Now I am getting a new error that I'm not sure how to resolve.

<details><summary>Code output</summary><p>

NoSuchNodeError                           Traceback (most recent call last)
~\Anaconda3\envs\DEEPLABCUT\lib\site-packages\pandas\io\pytables.py in get_node(self, key)
   1502         try:
-> 1503             node = self._handle.get_node(self.root, key)
   1504         except _table_mod.exceptions.NoSuchNodeError:

~\Anaconda3\envs\DEEPLABCUT\lib\site-packages\tables\file.py in get_node(self, where, name, classname)
   1634             nodepath = join_path(basepath, name or '') or '/'
-> 1635             node = where._v_file._get_node(nodepath)
   1636         elif isinstance(where, (str, numpy.str_)):

~\Anaconda3\envs\DEEPLABCUT\lib\site-packages\tables\file.py in _get_node(self, nodepath)
   1589
-> 1590         node = self._node_manager.get_node(nodepath)
   1591         assert node is not None, "unable to instantiate node ``%s``" % nodepath

~\Anaconda3\envs\DEEPLABCUT\lib\site-packages\tables\file.py in get_node(self, key)
    431         if self.node_factory:
--> 432             node = self.node_factory(key)
    433             self.cache_node(node, key)

~\Anaconda3\envs\DEEPLABCUT\lib\site-packages\tables\group.py in _g_load_child(self, childname)
   1177         # Is the node a group or a leaf?
-> 1178         node_type = self._g_check_has_child(childname)
   1179

~\Anaconda3\envs\DEEPLABCUT\lib\site-packages\tables\group.py in _g_check_has_child(self, name)
    392         if node_type == "NoSuchNode":
--> 393             raise NoSuchNodeError(
    394                 "group ``%s`` does not have a child named ``%s``"

NoSuchNodeError: group ``/`` does not have a child named ``/tracks``

During handling of the above exception, another exception occurred:

AttributeError                            Traceback (most recent call last)
[...]

Issue #1368, opened 2021-07-01
pytables error in 2.2rc3


backyardbiomech:
**Describe the bug**

To test out 2.2rc3, I followed the installation instructions to install a new environment on my DLC computer (env file downloaded from the blog), rather than the latest from the repo. I've encountered two errors:

1. the instructions to run `pip install deeplabcut[gui]` from within the env did not grab rc3 (got 2.1.10.4 instead). I had to run `pip install deeplabcut[gui]==2.2rc3` Instructions should be clarified.
2. I was in the middle of a workflow, so I decided to jump right into evaluating a new network, but got a tables error, which I know was a bit of a problem during rc3 testing.

**Desktop (please complete the following information about your system):**
 - OS: Windows10
 - DeepLabCut Version `2.2rc3`
 - conda version 4.10.1 (ran `conda update conda` just before installation of rc3)
 - Single animal, called in jupyter lab nb

**To Reproduce**
Steps to reproduce the behavior, i.e.:
1. create env with `DEEPLABCUT.yaml`
2. activate env, and run `pip install deeplabcut[gui]==2.2rc3`
3. confirm version with `deeplabcut.__version__`
4. ran `dlc.evaluate_network(config, plotting=True, gputouse=0, show_errors=True, Shuffles=[1])`
4. See error:

<details><summary>TRACEBACK</summary><p>

```python

----> 5 dlc.evaluate_network(config, plotting=True, gputouse=0, show_errors=True, Shuffles=[1])

~\Miniconda3\envs\dlc22rc3\lib\site-packages\deeplabcut\pose_estimation_tensorflow\core\evaluate.py in evaluate_network(config, Shuffles, trainingsetindex, plotting, show_errors, comparisonbodyparts, gputouse, rescale, modelprefix, c_engine)
    613         # Loading human annotatated data
    614         trainingsetfolder = auxiliaryfunctions.GetTrainingSetFolder(cfg)
--> 615         Data = pd.read_hdf(
    616             os.path.join(
    617                 cfg["project_path"],

~\Miniconda3\envs\dlc22rc3\lib\site-packages\pandas\io\pytables.py in read_hdf(path_or_buf, key, mode, errors, where, start, stop, columns, iterator, chunksize, **kwargs)
    397             raise FileNotFoundError(f"File {path_or_buf} does not exist")
    398 
--> 399         store = HDFStore(path_or_buf, mode=mode, errors=errors, **kwargs)
[...]

PR #1400, opened 2021-07-10
Update DEEPLABCUT.yaml

backyardbiomech:
Users appear to be grabbing `DEEPLABCUT.yaml` from the repo in order to install `2.2rc3`. Unfortunately, it installs 2.1.10.4, and gets tables from conda rather than pip. Therefore, the instructions elsewhere to install the newest versions of the code with this conda env file are inaccurate. 

To eliminate confusion, the repo version of `DEEPLABCUT.yaml` should be updated to the same as the version on the blog.

On windows 10, I have tried using the existing repo file to install 2.1.10.4, then running `pip install deeplabcut[gui]==2.2rc3`. While it appears to install ok, I get a tables error (seen in #1368 and #1398 ). Uninstalling pytables with conda and reinstalling with pip works sometimes (#1368), but just now I tested it again and created a tensorflow error.

### DIFF
--- a/conda-environments/DEEPLABCUT.yaml
+++ b/conda-environments/DEEPLABCUT.yaml
@@ -22,7 +22,6 @@ dependencies:
   - pip
   - jupyter
   - nb_conda
-  - pytables
   - ffmpeg
   - pip:
-    - deeplabcut[gui]
+    - 'deeplabcut[gui]'==2.2rc3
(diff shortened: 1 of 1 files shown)


AlexEMG (approved):
Good change!

MMathisLab (approved):

jeylau (approved):

PR #1423, opened 2021-07-16
Added posecfg_template option to create_training_dataset

alexcwsmith:
This commit adds an optional parameter 'posecfg_template' to create_training_dataset, which allows the user to pass a path to a previous iteration pose_cfg file, and the function will use that for creation of a new pose_cfg alongside a new training dataset, rather than the default. It is convenient to be able to copy a previous training iteration's pose_cfg file to quickly create a custom pose_cfg without having to modify the default pose_cfg.yaml file (i.e. modifying the default for all users). If None is passed, it uses the default. This also enables very easy sharing of pose_cfg.yaml files between users.

I added some print statements warning of situations where the user has run create_training_dataset with net_type or augmenter_type that is different than what is specified in the posecfg_template, not sure if I am overlooking anything there or if it would be better/safer for that to result in an error than just a printed warning.

The added docstring is:
```
    posecfg_template: string (optional, default=None)
        Path to a pose_cfg.yaml file to use as a template for generating the new one for the current iteration. Useful if you
        would like to start with the same parameters a previous training iteration. None uses the default pose_cfg.yaml.

```


alexcwsmith:
I just realized it's hard to see the changes I made in this commit because I also ran 'black' on this code which made a lot of changes. If you would prefer I can make a new commit that I haven't run black on so it will show only the relevant parts of this PR.

AlexEMG (approved):

MMathisLab:
@jeylau @AlexEMG can be merged?

jeylau:
Merge conflicts are fixed :) I'll ask @intergalactic-mammoth's review and should be ready to merge.

jeylau (approved):

aristotelis-economides (approved):

PR #1464, opened 2021-08-17
Upgrade to Tenserflow 2.0

TheRook:
We are using DeepLabCut with hardware acceleration (Coral TPU and GPUs) which _require_ TF 2.0.  So we upgraded DeepLabCut and it working well!  Yeah, this upgrade is an absolute requirement, the depreciated TF 1.* branch will have more problems than just compatibility. 

We have tested this code on our side, and it appears to work like a charm.  No additional changes needed.


MMathisLab:
Hey @TheRook thanks! - note, TF is still possible with 2.2; as we don't want to drop backward compatibility just yet. Can you given me a better sense of the logic i.e. you want to drop tf_slim with this PR?

jeylau:
Hi @TheRook, dropping tf_slim in some places is always welcome 😊 I don't see how it is limiting in your case though. As Mackenzie wrote above, it is TF2 under the hood since deeplabcut rc2. Also, what's with the files you deleted in `generate_training_dataset`?

Issue #949, opened 2020-10-07
Unable to extract/refine tracklets, output h5 file contains all nan


rly:
OS: Windows 10
DeepLabCut Version: 2.2b8
Anaconda env used: DLC-CPU

**Describe the problem**

I am using the GUI and following the steps in this video: https://www.youtube.com/watch?v=JDsa8R5J0nQ&ab_channel=DeepLabCut

After evaluating and cross-validating the network, analyzing the videos, and converting to tracklets, I went to Refine Tracklets and launched the GUI. On the GUI, there were no markers, and I could not click/select anything. Hitting Save generated an h5 file where the (x, y, likelihood) values were all NaN.

I do not get an error message.

The generated file cat2DLC_resnet50_CatOct5shuffle1_2000_full.mp4 contains markers so it seems like some tracking was successful.

**Screenshots**
Refine Tracklets view. Are my inputs appropriate? 
![image](https://user-images.githubusercontent.com/310197/95378530-e4b43a80-0898-11eb-8865-a2809f7f1dd1.png)

Here is a link to my DLC files on Google Drive:
https://drive.google.com/drive/folders/1ZJQTlaJfro0Lz7kiNsz8hiRME0icn0l7?usp=sharing

Thanks in advance!

MMathisLab:
Hi @rly - there are many parameters you need to set yourself before the refine tracklets GUI step, I suggest looking here, and seeing if some of these parameters change the outcome (I suspect so, if you have a good pose estimation network):

https://github.com/DeepLabCut/DeepLabCut/blob/master/docs/maDLC_AdvUserGuide.md#---------------------decision--break-point--------------------

jeylau:
@rly, if you still haven't solved your issue, you could try `deeplabcut.utils.make_labeled_video.create_video_from_pickled_tracks` to see if tracking looks good before jumping to the tracklets; the mp4 video you produced earlier only tells you that detections look great, but it may be that something is failing when connecting the detections into animals. It could also help to re-run the cross-validation step and `deeplabcut.convert_detections2tracklets`.

bkacsoh:
I am having the same issue, where the test video looks great, but nothing is tracked when the video is analyzed.

bkacsoh:
![Screenshot from 2020-10-20 12-55-49](https://user-images.githubusercontent.com/53061481/96618982-9ad54680-12d3-11eb-9179-f41d3c3f9182.png)
[...]

PR #1466, opened 2021-08-20
shift tracklets['single'] as 'frame based' 

aiporre:
As I was mentioning in my comment [here](https://github.com/DeepLabCut/DeepLabCut/issues/949#issuecomment-902347902) I got a problem when the detection of my unique label was not present since the first frame, the after running `            d.convert_detections2tracklets(config_path, videos=videos)` the individual where shifted as shown in my comment:
image

this is fixed if we use the frame numbers as index, the same way that is done for the other multiparts, then the shift is not there anymore:
image


jeylau:
Great catch @aiporre, thanks a lot!

jeylau (approved):

PR #834, opened 2020-08-03
Epipolar lines and related changes

daltondm:
The purpose of this pull request is to add optional epipolar lines to images in the labeling toolbox. This can improve accuracy of the labels in each of the camera images. I also made changes to related code to make this process easier, as well as UseOverviewGuide.md and the DOCSTRINGS in each relevant file.

The updated code was tested on Ubuntu 18.04.

The multianimal testscript seemed to fail somewhere in the training or evaluating phase, but I don't believe this is due to any of my changes. My changes don't extend past the labeling phase and don't change how the labels are saved, and I have successfully used my modifications to take a project past the training and evaluation steps.

Please let me know if you have any questions or need more information.

[Description_of_changes.txt](https://github.com/DeepLabCut/DeepLabCut/files/5017446/Description_of_changes.txt)
epipolar_lines_example_image
[testscript_3d_output.txt](https://github.com/DeepLabCut/DeepLabCut/files/5017447/testscript_3d_output.txt)
[testscript_multianimal_output.txt](https://github.com/DeepLabCut/DeepLabCut/files/5017448/testscript_multianimal_output.txt)
[testscript_output.txt](https://github.com/DeepLabCut/DeepLabCut/files/5017449/testscript_output.txt)

PR #1479, opened 2021-08-24
Fix image color when labeling with epipolar lines 

daltondm:
After PR #834 was merged and closed, I noticed that some of the lines that dealt with BGR --> RGB conversion had not been accepted, which resulted in images displayed with the reversed colors when labeling with epipolar lines. My previous lines to address this were clunky, so I've removed them. Now I just pass im.copy() into the drawEpLines function rather than im - this is necessary because cv2 can't work with arrays that are views of other arrays with shared memory. Separately, I added back a comment to make the multiple_individuals and basic toolboxes more similar.

Thank you in advance for your time.


MMathisLab (approved):

jeylau:
Looks good to me! Thanks @daltondm 😊

jeylau (approved):

PR #1482, opened 2021-08-25
Copy video codecs when manipulating videos

jeylau:
Fixes the bug introduced in #1473
This copies the video bitstream to the output (i.e., there is no re-encoding), thus ensuring video quality is preserved.

### DIFF
--- a/deeplabcut/utils/auxfun_videos.py
+++ b/deeplabcut/utils/auxfun_videos.py
@@ -248,7 +248,7 @@ class VideoWriter(VideoReader):
         output_path = self.make_output_path(suffix, dest_folder)
         command = (
             f"ffmpeg -n -i {self.video_path} -ss {start} -to {end} "
-            f"-c:a copy {output_path}"
+            f"-c:v copy {output_path}"
         )
         subprocess.call(command, shell=True)
         return output_path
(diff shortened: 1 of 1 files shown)


MMathisLab (approved):
seems to make sense!

AlexEMG:
Tests fail.

AlexEMG:
Related to (which also didn't pass tests): https://github.com/DeepLabCut/DeepLabCut/pull/1473

KonradDanielewski:
Rescale and crop cannot copy the video with `-c:v copy` because it's not possible to apply filter and copy the stream at the same time. 

It won't be easy to make this robust for multiple codecs and containers, I think the commands would grow a lot because both codecs and bitrates would have to be checked with ffprobe and then applied during the command. At least I don't know how and if this can be done differently. Not sure why it would give any errors during copying the stream when shortening (unless example video was corrupted and stream could not be copied). 

I will read some examples about creating the command based on ffprobe readout of codec and bitrate so maybe I can write something that makes sense.

MMathisLab:
Is this PR still active, or can close?

jeylau:
Closing, as Konrad rightly noted that we cannot filter the stream while avoiding re-encoding at the same time.

PR #1505, opened 2021-09-14
Fix: Impossible to go back up after using DOWN arrow

remi-pr:
When using the DOWN arrow to switch body part, and going to the last one it was impossible to go back up again. See: https://github.com/DeepLabCut/DeepLabCut/pull/319#issuecomment-500646205

This is a very small change in the code of labeling_toolbox.py, just changing a condition, no new function or functionality is added.

### DIFF
--- a/deeplabcut/gui/labeling_toolbox.py
+++ b/deeplabcut/gui/labeling_toolbox.py
@@ -552,7 +552,7 @@ class MainFrame(BaseFrame):
         """
         This function is to create a hotkey to skip up on the radio button panel.
         """
-        if self.rdb.GetSelection() < len(self.bodyparts) - 1:
+        if self.rdb.GetSelection() > 0:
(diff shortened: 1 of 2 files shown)


MMathisLab (approved):

MMathisLab:
tested - macOS! thanks @remi-pr I had totally forgotten about that :D / can you confirm it's fine on windows (if you happened to have used windows)

remi-pr:
Did it on Linux but ckecked on a Windows machine just now and it works.
I don't think it works on multiple animals projects though.  I should do the same modification [here ](https://github.com/DeepLabCut/DeepLabCut/blob/bb01775fef4147a752e1d053423ccffb0dfdf23b/deeplabcut/gui/multiple_individuals_labeling_toolbox.py#L707) as well.

AlexEMG (approved):

PR #1517, opened 2021-09-24
Tips on video reencoding and preprocessing

KonradDanielewski:
Hi, 

I finally had some time to write up a short tutorial on reencoding and preprocessing. I am aware that this is a wall of text but I wanted it to be as extensive as possible. I was thinking about adding a part on GPU usage in video encoding but this might be out of scope of DLC use. Let me know if you'd like me to change something or if you have some general pointers on how to improve this.


MMathisLab:
This is really great, thank you @KonradDanielewski !

MMathisLab (approved):

PR #1526, opened 2021-10-05
Fix right-click bug for removing label

m11o:
This Pull Request fix bug labels aren't able to be removed by right-clicking when using 'refine labels'.
Currently, for removing labels, middle-click is removed labels. But, according to the message of DLC's GUI, the correct way is right-click.

ref: [click event type for matplotlib](https://matplotlib.org/stable/api/backend_bases_api.html#matplotlib.backend_bases.MouseButton)


AlexEMG (approved):

AlexEMG (commented):
We should update the docs rather than the code. Middle click is indeed used for deleting.

m11o:
I modified the documents and comments in #1530 
@AlexEMG

PR #1529, opened 2021-10-08
Substitute cv2 imresize with skimage

### LINKED ISSUE #1528: skimage resize for paf
**Is there an existing issue for this?**
No
**Bug description**
When number of channel for paf exceeds 512, opencv resize cannot hanlde it and raise an error
**Operating System**
Linux 18.06
**DeepLabCut version**
2.2.0.2
**DeepLabCut mode**
Multiple animal
**Device type**
GPU
**Steps To Reproduce**
Have number of joints more than 25 and use full graph

jeylau:
Users (https://forum.image.sc/t/dlc-deeplabcut-extract-save-all-maps-function-gives-an-opencv-error/58083/6) and @yeshaokai reported image resizing failure. Switching to skimage, which imposes no limit on the number of channels, fixes the issue.

Fixes #1528

### DIFF
--- a/deeplabcut/pose_estimation_tensorflow/visualizemaps.py
+++ b/deeplabcut/pose_estimation_tensorflow/visualizemaps.py
@@ -9,10 +9,9 @@ Licensed under GNU Lesser General Public License v3.0
 """
 
 import os
-
-import cv2
 import matplotlib.pyplot as plt
 import numpy as np
+from skimage.transform import resize
 
 
 def extract_maps(
@@ -267,9 +266,7 @@ def extract_maps(
 
 def resize_to_same_shape(array, array_dest):
     shape_dest = array_dest.shape
-    return cv2.resize(
-        array, (shape_dest[1], shape_dest[0]), interpolation=cv2.INTER_CUBIC
-    )
+    return resize(array, (shape_dest[1], shape_dest[0]))
 
 
 def resize_all_maps(image, scmap, locref, paf):
--- a/deeplabcut/pose_estimation_tensorflow/visualizemaps.py
+++ b/deeplabcut/pose_estimation_tensorflow/visualizemaps.py
@@ -266,7 +266,7 @@ def extract_maps(
 
 def resize_to_same_shape(array, array_dest):
     shape_dest = array_dest.shape
-    return resize(array, (shape_dest[1], shape_dest[0]))
+    return resize(array, (shape_dest[0], shape_dest[1]))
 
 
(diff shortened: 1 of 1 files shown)


backyardbiomech:
Also posted to forum:
I just pulled that branch and am testing it. It's not for the faint of (computer) heart.

This is with 31 body parts per individual, one individual. My videos are about 1024x1024, with global_scale=0.8. Windows.

I passed two indices to `extract_save_all_maps`

It took about 6 minutes **per image** to produce all three maps, while also using up about 30+GB RAM and all of my 24GB of GPU memory. The RAM seemed to indicate a memory leak or non-release between images (ever increasing).

AlexEMG:
We could also use TensorFlow here: https://www.tensorflow.org/api_docs/python/tf/image/resize

MMathisLab:
Discuss today with @jeylau -- looking at that many pafs is really not informative anyhow; I would suggest we sub-sample, and/or can be user defined (only the skeleton from config.yaml, or edges directly passed into the function).

jeylau:
By the way @backyardbiomech, I've realized that a subset of body parts can be passed to the function, so only the related maps are shown (at a fraction of the cost I suppose).

backyardbiomech:
Thanks, @jeylau. I'm tweaking some (hundreds of) labels in the napari labeler now, then will try some maps after retraining. Those maps I plotted before with all bodyparts were bluntly less than informative.

jeylau:
@MMathisLab, I'd suggest to close this PR, as `comparisonbodyparts` already allows one to show only a subset of all maps.

backyardbiomech:
I just happened to test with 6 (of 30) body parts entered in `comparisonbodyparts`. It worked, but was still slow (3 minutes for a single image). I assume that with so many other body parts, each bp has lots of connections and paf to everywhere else, so even a single bp takes a lot of processing? 

Only vaguely related to this PR, it would be nice to extract locref and scmap for all (or selected) images without waiting for paf processing. Can we add something like `extractpaf= False` to `extract_save_all_maps`?

MMathisLab:
I like the flag optional PAF, good idea

jeylau:
@backyardbiomech's suggestion is now in 😃

backyardbiomech:
@jeylau , YES! Thank you! 
I think, now, between the auto tracklet/stitching, and this, a user (me) can effectively run a "single animal" project while taking advantage of the skeleton and dlcrsnet for pose detection in difficult filming scenarios.

MMathisLab (approved):

PR #1530, opened 2021-10-11
Modification of documents and comments for label removed by middle click

m11o:
Bellow PR modified the clicking function for removing labels. 
#1526
But, middle-click is correct for removing labels.
So, this PR change the documents and comments for removing labels by middle-clicking.


MMathisLab (approved):

AlexEMG (approved):
Thanks!

PR #1534, opened 2021-10-12
Modify fixed fps with keypoint only option

m11o:
When using `keypoints_only` option in `create_labeled_video`, created video is **fixed 25**.
So, This PR modifies the same fps of the selected video when labeled videos with keypoints_only option are created.

##### selected video
Screenshot from 2021-10-12 17-17-16
##### created video
Screenshot from 2021-10-12 17-16-58

##### selected video
Screenshot from 2021-10-12 17-14-25
##### created video
Screenshot from 2021-10-12 17-13-40

### DIFF
--- a/deeplabcut/utils/make_labeled_video.py
+++ b/deeplabcut/utils/make_labeled_video.py
@@ -680,7 +680,7 @@ def proc_video(
             print(e)
 
 
-DEFAULT_FRAME_RATE_WITH_KEYPOINTS_ONLY = 25
+DEFAULT_FRAME_RATE_FOR_LABELLED_VIDEO = 25
 def create_video_with_keypoints_only(
     df,
     output_name,
@@ -703,7 +703,7 @@ def create_video_with_keypoints_only(
     ny = int(np.nanmax(df.xs("y", axis=1, level="coords")))
 
     if fps is None:
-        fps = DEFAULT_FRAME_RATE_WITH_KEYPOINTS_ONLY
+        fps = DEFAULT_FRAME_RATE_FOR_LABELLED_VIDEO
 
     n_frames = df.shape[0]
     xyp = df.values.reshape((n_frames, -1, 3))
--- a/deeplabcut/utils/make_labeled_video.py
+++ b/deeplabcut/utils/make_labeled_video.py
@@ -599,6 +599,7 @@ def proc_video(
                 if bodyparts2connect:
                     all_bpts = df.columns.get_level_values("bodyparts")[::3]
                     inds = get_segment_indices(bodyparts2connect, all_bpts)
+                clip = vp(fname=video, fps=outputframerate)
                 create_video_with_keypoints_only(
                     df,
                     videooutname,
@@ -609,7 +610,9 @@ def proc_video(
                     skeleton_color=skeleton_color,
                     color_by=color_by,
                     colormap=cfg["colormap"],
+                    fps=clip.fps(),
                 )
+                clip.close()
             elif not fastmode:
                 tmpfolder = os.path.join(str(videofolder), "temp-" + vname)
                 if save_frames:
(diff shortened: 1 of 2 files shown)


deeplabcut/utils/make_labeled_video.py
```diff
@@ -640,6 +643,7 @@ def proc_video(
                     displaycropped,
                     color_by,
                 )
+                clip.close()
```
m11o, line 646: Clip variable is an instance of VideoProcessorCV without keypoints_only and fastmode option.
Besides, it is used in 'CreateVideoSlow' function. But, it isn't closed in 'CreateVideoSlow' function.
So, To release tempfile of video, I fixed that clip call close function after CreateVideoSlow.

ref: 
https://github.com/DeepLabCut/DeepLabCut/blob/174b6d159f6985470ebb627b4efdca183aaaf5af/deeplabcut/utils/make_labeled_video.py#L617
https://github.com/DeepLabCut/DeepLabCut/blob/174b6d159f6985470ebb627b4efdca183aaaf5af/deeplabcut/utils/make_labeled_video.py#L181

deeplabcut/utils/video_processor.py
```diff
@@ -148,5 +148,7 @@ def save_frame(self, frame):
         self.svid.write(np.flip(frame, 2))
 
     def close(self):
-        self.svid.release()
-        self.vid.release()
+        if hasattr(self, 'svid') and self.svid is not None:
+            self.svid.release()
+        if hasattr(self, 'vid') and self.vid is not None:
+            self.vid.release()
```
m11o, line 154: There is some possibility that either `svid` or `vid` don't exist in VideoProcessor class.
So, I fixed that these are released only in the case the instance has each attribute.

AlexEMG: Great!

AlexEMG (approved):
Thanks for your contribution! @jeylau please check too.

deeplabcut/utils/make_labeled_video.py
```diff
@@ -697,6 +702,9 @@ def create_video_with_keypoints_only(
     nx = int(np.nanmax(df.xs("x", axis=1, level="coords")))
     ny = int(np.nanmax(df.xs("y", axis=1, level="coords")))
 
+    if fps is None:
```
jeylau: I don't understand that part: how is it different from having fps=25 as the default argument?

jeylau:
Besides reverting the change to `create_video_with_keypoints_only` above, that looks good to me @masahito1997 ☺️

deeplabcut/utils/make_labeled_video.py
m11o: There is a possibility that the fps argument is `None`. Because FPS method may be None in `VideoProcessor` class. Besides, if users sometimes used `create_video_with_keypoints_only` method, they might take None in fps argument. So, I modified the code that fps expressly is inserted default value (25fps) when it is `None`.
@jeylau

jeylau: `VideoProcessor().fps()` cannot be None: it's either the video FPS (by default) or overwritten by whatever value a user passed in for `fps` (or the value of `outputframerate` for that matter)

m11o: Oh, that's great!
I miss the code. Is the code you tell me below?
https://github.com/DeepLabCut/DeepLabCut/blob/e2df9555b419a340ddae45fa779222a5ff717659/deeplabcut/utils/video_processor.py#L134

m11o: I fixed the default value for fps argument.  0a1c5ae
Would you please review again?

@jeylau @AlexEMG

jeylau:
Sweet! Thanks a lot @masahito1997!

jeylau (approved):

PR #1547, opened 2021-10-17
Update dlc entrypoint

hmaarrfk:
I believe the current entry-point no longer exists now that the package name is `deeplabcut` and not `dlc`.

Is this is the intended entry point?


hmaarrfk:
Am i missing something here? I think that the current `dlc` entry point is broken. Shouldn't it be fixed or removed? Is there maybe a larger cleanup going on and you don't want to have merge conflicts?

PR #1556, opened 2021-10-22
Speed up maDLC workflow when there is only one animal

jeylau:
Bypass assembly and tracking if only one individual is defined. Since there will be only one resulting tracklet, stitching is instantaneous so it doesn't need an update.

### DIFF
--- a/deeplabcut/pose_estimation_tensorflow/lib/inferenceutils.py
+++ b/deeplabcut/pose_estimation_tensorflow/lib/inferenceutils.py
@@ -622,6 +622,13 @@ class Assembler:
         if not any(i in bag for i in range(self.n_multibodyparts)):
             return None, unique
 
+        if self.max_n_individuals == 1:
+            get_attr = operator.attrgetter("confidence")
+            ass = Assembly(self.n_multibodyparts)
+            for joints in bag.values():
+                ass.add_joint(max(joints, key=get_attr))
+            return [ass], unique
+
         if self.identity_only:
             assemblies = []
             get_attr = operator.attrgetter("group")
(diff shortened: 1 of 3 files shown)


AlexEMG (approved):

MMathisLab (approved):

MMathisLab:
@jeylau can you add in the docs somewhere so it's clear for single animal in multi-mode users know how this works? then approved fully 👯

jeylau:
@MMathisLab, nothing changes for the users; it's done under the hood (in pretty much the same way we treat uniquebodyparts) ☺️ Or is that precisely what you want me to document? 😄

backyardbiomech:
Since this would apply to my current project of focus...

I would still run `convert_detections2tracklets` and `stitch_tracklets` separately, but it would be faster?

jeylau:
@backyardbiomech, yes! Instead of assembling keypoints, we keep the ones with highest confidence (as in single animal projects), and there is no "tracking": we'll simply form a single track containing all detections.

MMathisLab:
Hey @jeylau actually though, I would be most excited if someone can just run analze_videos and if they pass something like `num_of_animals=1` then it auto runs  `convert_detections2tracklets` and `stitch_tracklets`

jeylau:
@MMathisLab, I'll PR this separately one. We can bundle all the maDLC steps into `analyze_videos` (with the most relevant arguments) to have a behavior close to that of single animal projects. That's the idea, right?

MMathisLab:
yep! But I think the true single step for maDLC is indeed another PR, but for this I think a simple `#animal=True/1` flag would be good; also for real-time DLC things in the future

jeylau:
Got you! This can be edited in the test/inference_cfg.yaml with `topktoretain` (which is otherwise set to the # of animals in the main config.yaml by default)

MMathisLab:
Yep! exactly; I was thinking that if num_animals=1, that would write into the `inference_cfg.yaml`  `topktoretain` =1, so then your code would then it auto runs `convert_detections2tracklets` and `stitch_tracklets` -- no reason to not have this all together; i.e., why should I even need to run them at all.

PR #1560, opened 2021-10-27
Plotting assemblies in evaluation 

AlexEMG:
Use case for multi-animal projects:
```
deeplabcut.evaluate_network('/analysis/project/reaching-task/config.yaml',Shuffles=[1],plotting=True)
```
Will create plots like this:

Test-videocompressed1-img0000

and 
```
deeplabcut.evaluate_network('/analysis/project/reaching-task/config.yaml',Shuffles=[1],plotting="individual")
```
will create plots like this:

Test-videocompressed1-img0000


AlexEMG:
Proposed use case for multi-animal projects:
```
deeplabcut.evaluate_network('/analysis/project/reaching-task/config.yaml',Shuffles=[1],plotting=True)
```
and 
```
deeplabcut.evaluate_network('/analysis/project/reaching-task/config.yaml',Shuffles=[1],plotting=assembly)
```

See: 
https://github.com/DeepLabCut/DeepLabCut/blob/444da3da70b1f3bfaae39b0849e0aafcff1a9b7a/deeplabcut/pose_estimation_tensorflow/core/evaluate.py#L498

MMathisLab:
@AlexEMG could you merge conflicts first? Also, maybe an example of how this is useful for users?

jeylau (approved):

MMathisLab:
I don't understand the docs: "They can alternatively be colored by bodypart type by passing `plotting`=`bodypart`." It was already colored by bodypart, i.e., each bodypart gets a separate color. Also, `plotting=assembly` is not documented; @AlexEMG

jeylau:
But we agreed this morning that coloring by individual (to visualize assembly clearly) was more useful for maDLC.

aristotelis-economides:
@MMathisLab I think the idea is this:

The default in maDLC with the current PR will be visualizing by individuals if turned on. Therefore `plotting=True` defaults to individuals. If one prefers they can set it to `plotting='bodypart'`. So `plotting='assembly'` that Alex initially suggested is not needed the way it is currently implemented.

Personally I think it is a bit confusing to use both `bool` and `str` in the same variable (but hey, it's python) and would do: 
```python
#default 
plotting=None

#possible values
plotting='individuals' 
plotting='bodypart'
```
An argument against this would be that it breaks backwards compatibility, which is the reason I think @jeylau opted for the way he did.

MMathisLab:
Right, understood, but what I would not like is if `plotting=True` changes the behavior **of maDLC**; what we agreed on was adding `plotting=assembly`, which would plot by individual, and `plotting=True` or `plotting=keypoints` keeps it "as is".

note, I don't know why we deviate from how it is done in check_labels:

```
deeplabcut.check_labels(config_path, visualizeindividuals=True/False)
```

namely, why don't we just do that here, so the nomeclature is the same? pass `visualizeindividuals=True` --> get assemblies.

jeylau:
I'll revert the default for maDLC then. The convention of check_labels doesn't work here: True/False for visualizeindividuals isn't enough to not plot anything at all.

MMathisLab:
I mean: `visualizeindividuals` is what could be passed, vs. "assembly" which is much more of a jargon term...

jeylau:
I have changed "assembly", so the convention is that of `create_labeled_video`: "bodypart" or "individual". Clearer and more consistent, perhaps 🤔

jeylau (approved):

deeplabcut/pose_estimation_tensorflow/core/evaluate.py
```diff
@@ -519,8 +519,9 @@ def evaluate_network(
         Integer specifying which TrainingsetFraction to use. By default the first (note that TrainingFraction is a list in config.yaml). This
         variable can also be set to "all".
 
-    plotting: bool, optional
-        Plots the predictions on the train and test images. The default is ``False``; if provided it must be either ``True`` or ``False``
+    plotting: bool or str, optional
+        Plots the predictions on the train and test images.
+        The default is ``False``; if provided it must be either ``True``, ``False``, "bodypart", or "individual".
```
aristotelis-economides, line 524: Maybe add this here? 

`Setting to ``True`` defaults as "bodypart".`

AlexEMG (commented):
Done and tested.

MMathisLab (approved):

PR #1566, opened 2021-11-04
Ignore single-keypoint assemblies during mAP evaluation

jeylau:
Note that optimal assembly matching is a bit slower (~15%), but that's perhaps more readable (and not that significant in practice)


AlexEMG (approved):
Va bene! 👍

PR #1567, opened 2021-11-05
Uniform analyze_videos API

jeylau:
Automatically chain video analysis, tracking & stitching with the new argument `auto_track=True`. This produces the final h5 file in one function call, equivalent to the behavior of `analyze_videos` in single-animal projects. Note that `track_method` has now been moved to `default_track_method` in the config.yaml, simplifying many core functions (and some GUI tabs).


MMathisLab:
@jeylau would this https://github.com/DeepLabCut/DeepLabCut/pull/1556 still be required then, or?

deeplabcut/pose_estimation_tensorflow/predict_videos.py
```diff
@@ -1129,15 +1155,18 @@ def _convert_detections_to_tracklets(
     data,
     metadata,
     output_path,
-    track_method="ellipse",
     greedy=False,
     calibrate=False,
 ):
+    track_method = cfg.get("default_track_method", "ellipse")
```
MMathisLab, line 1169: should we have a catch for older projects, i.e., if cfg.get("default_track_method") = false, then write it into config.yaml and set default as ellipse?

docs/maDLC_UserGuide.md
```diff
@@ -441,13 +441,13 @@ After pose estimation, now you perform assembly and tracking. *NEW* in 2.2 is a
 ```python
 deeplabcut.analyze_videos(config_path, ['/fullpath/project/videos/'], videotype='.mp4')
 ```
-Note: You do **not** get the .h5/csv file you might be used to getting in standard DLC (this comes next!).
+*NEW* in 2.2.1: `deeplabcut.analyze_videos` has a new argument `auto_track=True`, chaining pose estimation, tracking, and stitching in a single function call with defaults we found to work well. Thus, you'll now get the .h5 file you might be used to getting in standard DLC. If `auto_track=False`, one must run `convert_detections2tracklets` and `stitch_tracklets` manually (see below), granting more control over the last steps of the workflow (ideal for advanced users).
```
MMathisLab: I think we might want to bump to 2.2.0.3 this week; keep 2.2.1 for full clean version

jeylau:
@MMathisLab, yes, because whether functions are chained or not it is still desirable to optimize assembly/tracking when there is only an animal ☺️

MMathisLab (changes requested):

deeplabcut/pose_estimation_tensorflow/predict_videos.py
jeylau: Good idea! I've just added it 😊

MMathisLab (approved):

MMathisLab:
okay looks good to me once new merge conflict from other PR merged! thanks @jeylau 💜

jeylau:
@MMathisLab, resolved ☺️

AlexEMG (changes requested):
Great update! 

Esp. also adding the default_track_method to the config.yaml

However, I'm not sure why we remove track_method as a parameter, just because there is a default now in the config? As for other parameters, users can usually overwrite the default by passing it directly to the function?

AlexEMG:
Put some draft here: https://github.com/DeepLabCut/DeepLabCut/commit/9dfb8afbd2f138f1a9ccf3743e4a041feefb4b29

jeylau:
@AlexEMG, I like it when you're in the refactoring mood 😁

jeylau:
Could you run black on your changes and merge them into mine?

MMathisLab:
@jeylau not sure @AlexEMG can merge into your github, since it's not a DLC branch; can you add his edits and merge, so this PR can be merged? thanks! (as I want to work on GUI)

AlexEMG (approved):

PR #1577, opened 2021-11-13
Avoid hack when forking new processes

hmaarrfk:
I was debugging what seems to be something regarding pyzmq forking off new processes and I found this.

I believe this addresses the hack by making the definition of function global and passing self as a parameter.

I hope this helps.

Best,

### DIFF
--- a/deeplabcut/pose_estimation_tensorflow/lib/inferenceutils.py
+++ b/deeplabcut/pose_estimation_tensorflow/lib/inferenceutils.py
@@ -40,6 +40,13 @@ def _conv_square_to_condensed_indices(ind_row, ind_col, n):
 Position = Tuple[float, float]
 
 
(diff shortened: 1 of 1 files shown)


jeylau:
Hi @hmaarrfk, are assemblies still pickable? If they do, the change looks good to me!

hmaarrfk:
I'm like 95% sure I found this because I was stepping through my code. I made the modification and tested my model with this afterward.

So I think the answer is "yes it works".

https://documentation.help/Python-PEP/node66.html

hmaarrfk:
Happy new year! 

I'm wondering if I could have some feedback as to why this was closed.

Was it introducing a bug I am not aware of?

In my opinion (and that of the original person who wrote the code) using the `global` keyword is typically a good way to make mistakes in python.

PR #1580, opened 2021-11-15
performance fix on multiple_individuals_labeling_toolbox

MitchFuchs:
getColorIndices called only once to reduce process time to load next/previous frame in labeling tool.

process time before 5-7 sec
process time after 1,5 sec

(with 5 individuals and 17 bodyparts)

### DIFF
--- a/deeplabcut/gui/multiple_individuals_labeling_toolbox.py
+++ b/deeplabcut/gui/multiple_individuals_labeling_toolbox.py
@@ -1194,6 +1194,10 @@ class MainFrame(BaseFrame):
         """
         self.drs = []
         self.updatedCoords = []
+        if len(self.individual_names)>1:
+            self.norm, self.colorIndex = self.image_panel.getColorIndices(
+                self.img, self.multibodyparts
+            )
         for j, ind in enumerate(self.individual_names):
             idcolor = self.idmap(j)
             if ind not in self.dataFrame.columns.get_level_values(1):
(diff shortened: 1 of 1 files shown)


jeylau:
Hey @MitchFuchs, thanks for your contribution 😃 Now that `self.norm, self.colorIndex` are defined outside the loop for `multianimalbodyparts`, aren't they overwritten (and never set again) as soon as we loop over a "single" individual?

jeylau:
Also note that we have a new (alpha-released) labeling GUI (https://github.com/DeepLabCut/napari-DeepLabCut) that is much smoother 😄

MitchFuchs:
> Hey @MitchFuchs, thanks for your contribution 😃 Now that `self.norm, self.colorIndex` are defined outside the loop for `multianimalbodyparts`, aren't they overwritten (and never set again) as soon as we loop over a "single" individual?

To my current knowledge the following condition ` if ind == "single":` never returns `True ` if the project is multianimal. So we set it once, outside the loops with condition `len(self.individual_names)>1`. But in case, `ind == "single"`, then `self.norm`, `self.colorIndex` will simply be overwritten.

Does it make sense?

jeylau:
It could evaluate to True if some unique body parts are defined. I just checked though that, if present, 'single' is always the last individual name returned (https://github.com/DeepLabCut/DeepLabCut/blob/b9577b35609b89546cf68b9d89b5e8a76eb777c8/deeplabcut/gui/multiple_individuals_labeling_toolbox.py#L743-L746) so we're safe! Thanks!

jeylau (approved):

AlexEMG:
Nice, much faster! Thanks @MitchFuchs

AlexEMG (approved):

PR #1584, opened 2021-11-19
Better format for cross-platform labelling and training (storing triples) 

jeylau:
Image paths were stored with OS-specific path separators, making labeling/training hard across platforms (e.g., forcing users to create the training data with the machine they train their models on, and occasionally to convert their data). Paths are now stored without separators as MultiIndex (labeled-data, subfolder, imagename), making indexing/merging DataFrames easy and robust; only upon reading an image are the paths reconstructed.
I tested the GUIs as well ☺️

### DIFF
--- a/deeplabcut/generate_training_dataset/multiple_individuals_trainingsetmanipulation.py
+++ b/deeplabcut/generate_training_dataset/multiple_individuals_trainingsetmanipulation.py
@@ -62,7 +62,7 @@ def format_multianimal_training_data(
     array = np.round(array, decimals=n_decimals)
     for i in tqdm(train_inds):
         filename = filenames[i]
-        img_shape = read_image_shape_fast(os.path.join(project_path, filename))
+        img_shape = read_image_shape_fast(os.path.join(project_path, *filename))
         joints = dict()
         has_data = False
         for n, xy in enumerate(array[i]):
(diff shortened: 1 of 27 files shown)


deeplabcut/generate_training_dataset/multiple_individuals_trainingsetmanipulation.py
```diff
@@ -97,7 +97,6 @@ def create_multianimaltraining_dataset(
     config,
     num_shuffles=1,
     Shuffles=None,
-    windows2linux=False,
```
MMathisLab: maybe we should leave, but keep false / make a depreciation warning, as I imagine some people pass this in their code quite a bit?

MMathisLab (changes requested):
maybe just leave term in and add deprecation warning, i..e, it becomes a dummy variable, but then people won't get the error invalid input , etc etc (i..e, in our colab notebooks, etc)

deeplabcut/generate_training_dataset/multiple_individuals_trainingsetmanipulation.py
jeylau: Great point, I'll add the warning!

AlexEMG (changes requested):
- let's add some tests 
- update the examples

AlexEMG (approved):

PR #1607, opened 2021-12-09
Randomly prune large graphs

jeylau:
For projects with a high number of annotated body parts, default graphs grow unnecessarily large, incurring expensive computing costs. In these cases, we've found that graphs could be randomly pruned to half their size with little decrease in mAP (~–1 point).


AlexEMG (commented):
- [x] Maybe not just doing half, but some sublinear function? -> see degree 
- [x] And allowing the user to overwrite how many edges there are?  
- [x] Perhaps also put the graph finding part somewhere in utils (and call it)?

jeylau:
@AlexEMG, I agree with your comment above of using something else than simply pruning the graph to 50% of its original size (which for large graph is definitely still prohibitive). With the marmosets, 50% of 105 edges gives on average nodes of degree 3–4 (53 edges / 15 body parts); that value would be a more reasonable target for pruning, what do you think? For a project with 40 keypoints for example, we'd prune the graph from 780 down to ~40*3.5=140 (vs 390).

AlexEMG:
- [x] Let's make the degree the parameter to pass. _And use e.g. 5 by default for projects with more than 16 body parts._

jeylau:
@AlexEMG, I've used degree 6 by default (as then, on average, exactly 3 * n_bodyparts are required for an undirected graph); I have added it to the unit test. The pruning target can now be either a number of edges or the desired degree ☺️

AlexEMG:
- make the 105, degree a variable.

jeylau:
@AlexEMG, let's rather wait to have dataset creation and training+setting parameters decoupled? That way, the first stays simple, and we can pass more args to the latter for custom training?

AlexEMG (approved):

PR #1611, opened 2021-12-13
Backward compatibility of training on data stored the old way

jeylau:
#1584 assumed that training data would be newly formed (i.e., with path separators already stripped) when training a network. This is not always the case though (e.g., when retraining previous networks with deeplabcut-master...). This is now fixed with this PR.


AlexEMG:
Note - this didn't affect any release. Only master.

PR #1613, opened 2021-12-14
Allow windows to resize to be smaller than the default size

hmaarrfk:
With multimonitors on linux, it seems that wx can sometimes detect a single "big" monitor, instead of multiple ones.

the code in `widgets.py`

```
        displays = (
            wx.Display(i) for i in range(wx.Display.GetCount())
        )  # Gets the number of displays
        screenSizes = [
            display.GetGeometry().GetSize() for display in displays
        ]  # Gets the size of each display
```
Returns a single monitor, with a riddiculously sized screen.

When it is multipled by `0.8` and `0.75` it is still often too big.

While I wish I had time to debug why wx was doing this, seeing as DeepLabCut has intentions on moving to qt, this doesn't seem so worth our time.

However, it seems reasonable to allow the users to resize their screens to something smaller than what is proportional to a window.

I suggest 800 x 600 here but I guess the numbers are adjustable.

This makes this appear:
image

Not "ideal" but I've seen some people with really small monitors. 
1400 x 600 would yield
image

I hope this helps.

### DIFF
--- a/deeplabcut/gui/widgets.py
+++ b/deeplabcut/gui/widgets.py
@@ -51,7 +51,7 @@ class BaseFrame(wx.Frame):
     """Contains the main GUI and button boxes"""
 
     def __init__(self, frame_title="", parent=None, imtypes=None):
-        # Settting the GUI size and panels design
+        # Setting the GUI size and panels design
         displays = (
             wx.Display(i) for i in range(wx.Display.GetCount())
         )  # Gets the number of displays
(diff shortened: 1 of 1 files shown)


hmaarrfk:
Just doing a cursory ping on this small modification.

For users with 2 side by side monitors, this was really necessary to get them going with DLC.

jeylau (approved):

jeylau:
Good catch @hmaarrfk! Thank you, it works fine!

hmaarrfk:
Thank you. Looking forward to the next release!

PR #1616, opened 2021-12-16
Update the model zoo Colab notebook

jeylau:
The skeleton definition in the config is no longer overwritten, and I've made paths more robust in the notebook as they were causing me some trouble. I had no problem otherwise running it with the demo video.

### DIFF
--- a/deeplabcut/create_project/modelzoo.py
+++ b/deeplabcut/create_project/modelzoo.py
@@ -267,25 +267,13 @@ def create_pretrained_project(
         pose_cfg = deeplabcut.auxiliaryfunctions.read_plainconfig(path_train_config)
         print(path_train_config)
         # Updating config file:
-        dict = {
+        dict_ = {
             "default_net_type": pose_cfg["net_type"],
             "default_augmenter": pose_cfg["dataset_type"],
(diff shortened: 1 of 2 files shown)


deeplabcut/create_project/modelzoo.py
```diff
@@ -267,25 +267,13 @@ def create_pretrained_project(
         pose_cfg = deeplabcut.auxiliaryfunctions.read_plainconfig(path_train_config)
         print(path_train_config)
         # Updating config file:
-        dict = {
+        dict_ = {
             "default_net_type": pose_cfg["net_type"],
             "default_augmenter": pose_cfg["dataset_type"],
             "bodyparts": pose_cfg["all_joints_names"],
-            "skeleton": [],  # TODO: update with paf_graph
```
MMathisLab, line 274: I don't think this is the proper fix; the issue is why is this here? This seems to load the pose_cfg / path_train_config, but then edit the wrong config file. this indeed should update the pose_cfg.yaml in thrain folder with the correct variables, including the PAF, as model zoo will be multi-animal by default soon.

MMathisLab (changes requested):

MMathisLab (approved):

PR #1621, opened 2021-12-20
simplified 1-click maDLC analyze video, track, create video within the GUI

MMathisLab:
# Goal:

To simplify GUI default maDLC into a single step: more advanced users should use the colab or terminal commands, but this makes it simple and straightforward for beginners. This PR directly addresses the GUI; as the one-step function was updated in #1567 , which now allows for dlc.analyze_videos(... `auto_track=True`) to complete analysis, stitching and tracking.

- Example of the new GUI layout for single-click maDLC: 
Screen Shot 2021-12-20 at 5 37 39 PM

## Done:

- [x] Analyze video tab now includes a "one-click" video analysis, create_full_detections_video, track+stitch (assuming same # of animals used in training).
Some dropped terms for simplification .. should decide if this is OKAY, or should make a depreciation warning as not fully backwards compatible (BC)? 
- [x] dropped: `video_type` as a selection menu (as the functions now work without passing this anyhow, so this is BC)
- [x] dropped: `trainingindex`: I suspect this is never changed from default 0... but, dropping could break someone's BC. 
- [x] dropped trainingindex in other tabs too
- [x] combined tab for extract outliers and label, simplifies workflow. 
- [x] filtering works + get CSV file for user
- [x] add in n_tracks, so people can pass in # of animals
- [x] more than 1 video selected working (thanks @backyardbiomech ! )
- [x] update how analyze_videos h5 to csv handles videos (thanks @AlexEMG)
- [x] plot_traj and plots pop up for saDLC and maDLC
- [x] able to use ID = true
- [x] able to use calibration = true
- [x] tested: all combinations of the tabs in the GUI for maDLC
- [x] enhancement: better print statement for `filenotfound` during video analysis to remind use to check iteration # in config.yaml too 
- [x] better network defaults for saDLC and maDLC; also dropped all the model options in the drop down in order to guide novice users to best options; left them all for model comparison though!

Here is the (minimal) expected output from the current WIP PR: 

```python
pythonw -m deeplabcut
Starting GUI...
DLC network loading and video analysis starting ...
Using snapshot-5 for model /Users/mwmathis/Documents/DeepLabCut/examples/multi_mouse-dlc_team-2021-09-03/dlc-models/iteration-0/multi_mouseSep3-trainset80shuffle1
[...]

PR #1622, opened 2021-12-22
Ignore numerical DataFrame index

jeylau:
Properly catch the error raised when attempting to split paths that are numerical (typically, when a DataFrame index contains only frame indices rather than image paths).


MMathisLab:
can you rather PR into https://github.com/DeepLabCut/DeepLabCut/pull/1621 as this is where the issue is open?

MMathisLab (approved):

PR #1652, opened 2022-01-25
Set confidence to 0 when instantiating Assembly

### LINKED ISSUE #1650: Non-tracked NaN points in maDLC saved with likelihood of 1.0

### Bug description

Single animal tracking in maDLC. 

If (part of) the animal starts offscreen, or ends up offscreen, the model, appropriately, cannot create tracks for invisible bodyparts for those frames. The saved coordinates are `nan, nan`. However, the saved likelihood is `1.00000`. 

Other than just being inaccurate, this creates problems in two realms:
1. Applying splines or other filtering regimes post-DLC where weights are based on likelihood thresholds.
2. Importing data to other workflows, especially 3d reconstruction in Anipose (my case), which uses a simple likelihood threshold for including data in calculations. Anipose will triangulate a point in a frame based on the lowest likelihood among the cameras. One camera with nans and 1.0 and another camera with points and 0.99 likelihood leads to problematic 3d reconstructions.

Simply setting the default likelihood to 0.0, or setting to 0 the likelihood for any rows of nans per bodypart, would fix it.

I have not checked to see if this is true in single animal DLC (it shouldn't be though since the nans will only show, I believe, when tracklets can't be formed or stitched for certain frames), or with more than 1 animal in maDLC.

### Operating System

Windows 10

### DeepLabCut version

2.2.0.3

### DeepLabCut mode

multi animal

### Device type

gpu

### Steps To Reproduce

_No response_

### Relevant log output

_No response_

### Anything else?

_No response_

### Code of Conduct

jeylau:
Default confidence is set to 0 unless overwritten when adding joints/links to the assembly.

Fixes #1650

### DIFF
--- a/deeplabcut/pose_estimation_tensorflow/lib/inferenceutils.py
+++ b/deeplabcut/pose_estimation_tensorflow/lib/inferenceutils.py
@@ -110,8 +110,10 @@ class Assembly:
         n_bpts = array.shape[0]
         ass = cls(size=n_bpts)
         ass.confidence = 1
-        ass.data[:, : array.shape[1]] = array
-        nonempty = np.flatnonzero(~np.isnan(array).any(axis=1))
+        ass.data[:, :array.shape[1]] = array
+        mask = np.isnan(array).any(axis=1)
+        ass.data[mask, 2] = np.nan
+        nonempty = np.flatnonzero(~mask)
         ass._visible.update(nonempty)
         return ass
 
--- a/deeplabcut/pose_estimation_tensorflow/lib/inferenceutils.py
+++ b/deeplabcut/pose_estimation_tensorflow/lib/inferenceutils.py
@@ -108,11 +108,13 @@ class Assembly:
 
     @classmethod
     def from_array(cls, array):
-        n_bpts = array.shape[0]
+        n_bpts, n_cols = array.shape
         ass = cls(size=n_bpts)
-        ass.data[:, :array.shape[1]] = array
-        nonempty = np.flatnonzero(~np.isnan(array).any(axis=1))
-        ass._visible.update(nonempty)
+        ass.data[:, :n_cols] = array
+        visible = np.flatnonzero(~np.isnan(array).any(axis=1))
+        if n_cols < 3:  # Only xy coordinates are being set
+            ass.data[visible, 2] = 1  # Set detection confidence to 1
+        ass._visible.update(visible)
         return ass
 
     @property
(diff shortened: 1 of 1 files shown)


backyardbiomech (approved):
Tested and that fixes it.  Thanks for the quick work. 
I'm wondering, though, what is more appropriate. A likelihood of `nan`, or `0`.

At least in [anipose](https://github.com/lambdaloop/anipose/blob/f4c229b6341f5b3b15e710e18a39f3c8573f9eab/anipose/triangulate.py#L236), `0` might be better because of the way the score for the triangulated point is calculated:

```python
scores_3d = np.min(all_scores, axis=0)
```
Though I think the handling of missing points in anipose needs to be improved beyond this line. 

But I'm also wondering about things like extracting outliers based on "uncertain" when checking for likelihoods **less than** some threshold. Where and how else is likelihood used, and would `0` or `nan` be better in those workflows?

AlexEMG:
Likelihood 0, seems like a reasonable convention that doesn't break any downstream computations.

jeylau:
It mostly depends on whether we want a frame with a missing detection to be caught as outlier with the "uncertain" method.  I'm not certain this is desirable (e.g., when an animal hides), as we'd end up flagging frames for which no user intervention is truly needed. Thoughts?

backyardbiomech:
This isn't an easy problem. I agree that not wanting to flag as uncertain a hiding animal is good, and in some maDLC projects animals may regularly leave the field of view. But at earlier stages of training a model more subtle losses of tracklet building (e.g. animals or limbs too close to each other) can also lead to nans, and those should be flagged. 

What if it's set to 0, but when extracting outliers there's an option to ignore (or set to nan) all zero values (e.g. <0.0001) using the `uncertain` method? That way, with no intervention, threshold-dependent downstream workflows should work, but the user has the option to ignore hiding animals when refining a model.

AlexEMG:
The outlier extraction methods should be updated anyway, to e.g. distinguish an animal disappearing for a while or some jittery predictions (either of which can be likelihood = 0). So I'd say, let's set it to zero and at some point put better extraction methods in?

MMathisLab:
agreel fine to set to 0 for now; requested change needed.

MMathisLab (changes requested):
nan --> 0

AlexEMG (approved):

jeylau:
Fixed (and edited PR name)!

AlexEMG (approved):

MMathisLab (approved):

PR #1654, opened 2022-01-27
Fix learning rate when retraining

### LINKED ISSUE #1646: Restarting a training is not following the lr pattern laid out in multiple-step

### Bug description

When stopping and restarting a training from a snapshot the match-step starts from the first step definition. However because it requires an exact match between the iteration number and the parameter in match-step it can be stuck in the fist definition for a surprising time!
For example:
'multi_step': [[0.005, 10000], [0.02, 430000], [0.002, 730000], [0.001, 1030000]],
if run  continuously, lr would be .05 till 10000, then .02 till 430000, then etc etc.
If however you stop at iteration 50000, and restart from that snapshot:
lr will start at .005, but since iterations will never be at 10000 again, lr will stay at .005 until iteration 430000 when it will jump to .002.
By changing the multi-step between restarts then one can overcome this, bit it took a while before I understood what was happening

I think the "problem" is in the code below which requires an exit match of iteration number to move to the next lr value:

> class LearningRate(object):
>     def __init__(self, cfg):
>         self.steps = cfg["multi_step"]
>         self.current_step = 0
> 
>     def get_lr(self, iteration):
>         lr = self.steps[self.current_step][0]
>         if iteration == self.steps[self.current_step][1]:
>             self.current_step += 1
> 
>         return lr

### Operating System

Mac-m1

### DeepLabCut version

2.2.0.3

### DeepLabCut mode

single animal

### Device type

gpu
Mac M1

### Steps To Reproduce

_No response_

### Relevant log output

_No response_

### Anything else?

_No response_

### Code of Conduct

jeylau:
The iteration # passed to the scheduler is now correct.
Note that max_iter is relative to start_iter; i.e., retraining from 50k iterations with max_iter=50k will train the model up to 100k.

Fixes #1646

### DIFF
--- a/deeplabcut/pose_estimation_tensorflow/core/train.py
+++ b/deeplabcut/pose_estimation_tensorflow/core/train.py
@@ -264,14 +264,14 @@ def train(
     max_iter += start_iter  # max_iter is relative to start_iter
     for it in range(start_iter, max_iter + 1):
         if "efficientnet" in net_type:
-            dict = {tstep: it - start_iter}
-            current_lr = sess.run(learning_rate, feed_dict=dict)
+            dict_ = {tstep: it - start_iter}
+            current_lr = sess.run(learning_rate, feed_dict=dict_)
         else:
             current_lr = lr_gen.get_lr(it - start_iter)
-            dict = {learning_rate: current_lr}
+            dict_ = {learning_rate: current_lr}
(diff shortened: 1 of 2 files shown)


jeylau:
Why did you close it @AlexEMG? 😅

PR #1657, opened 2022-01-28
Introduce DLC <-> NWB data conversion

jeylau:
DeepLabCut's h5 data files can be readily converted to NWB format either via the GUI from the `Analyze Videos` tab or programmatically, as follows:
```python
import deeplabcut

deeplabcut.analyze_videos_converth5_to_nwb(config_path, video_folder)
```
Note that DLC does not depend on dlc2nwb just yet; if attempting to convert to NWB, a user would be asked to run `pip install dlc2nwb`, first!

This code is based on https://github.com/DeepLabCut/DLC2NWB and the NWB data standard for pose data: https://github.com/rly/ndx-pose. We gratefully acknowledge the generous support from the Kavli Foundation via a Kavli Neurodata Without Borders Seed Grants. We gratefully acknowledge the generous support from the [Kavli Foundation](https://kavlifoundation.org/) via a [Kavli Neurodata Without Borders Seed Grants](https://www.nwb.org/nwb-seed-grants/). We also acknowledge feedback, and our collaboration with [Ben Dichter, Ryan Ly and Oliver Ruebel](https://www.nwb.org/team/) to develop the data standard!

You can also export data via the GUI:
export

### DIFF
--- a/deeplabcut/__init__.py
+++ b/deeplabcut/__init__.py
@@ -76,6 +76,7 @@ from deeplabcut.utils import (
     convert2_maDLC,
     convertcsv2h5,
     analyze_videos_converth5_to_csv,
+    analyze_videos_converth5_to_nwb,
     auxfun_videos,
 )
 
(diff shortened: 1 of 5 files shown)


AlexEMG (approved):
Thanks, that's great!

PR #1658, opened 2022-01-28
Use TF2 documented way to allow growth

jeylau:
I realized that allowing memory growth the old way had no effect when running the testscript; it now does.


AlexEMG (approved):

jeylau:
Do not merge yet @AlexEMG, it's not yet fine :)

PR #1659, opened 2022-02-01
Replace underscores for dashes in foldername

Tetra-quark:
The h5 file in the "evaluation-results" folder uses a "-" to append the snapshot information to the name string whereas the folder containing Test and Training Images in the same directory uses an "_".

### DIFF
--- a/deeplabcut/pose_estimation_tensorflow/core/evaluate.py
+++ b/deeplabcut/pose_estimation_tensorflow/core/evaluate.py
@@ -905,7 +905,7 @@ def evaluate_network(
                                 str(evaluationfolder),
                                 "LabeledImages_"
                                 + DLCscorer
-                                + "_"
+                                + "-"
                                 + Snapshots[snapindex],
                             )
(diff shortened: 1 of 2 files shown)


Tetra-quark:
It's my first pull request so any help or guidance appreciated. My aim is to make this naming convention consistent. 
I am trying to track down if this fixed naming convention might cause problems further down the line in the repository.

I noticed this underscore/dash mismatch in my analysis workflow when I received an error and traced it back to the difference between the folder name `LabeledImages_DLC_resnet101_MyProjectFeb1shuffle1_200000_snapshot-200000` and the h5 file from the same directory `DLC_resnet101_MyProjectFeb1shuffle1_200000-snapshot-200000.h5`

MMathisLab:
Hi @Tetra-quark thanks for the PR! If we change though, it won't be backwards compatible with older projects

Tetra-quark:
Fair enough then, shall I close this request?

MMathisLab:
Thank you though, and of course don't let this discourage you for any and all future PRs, we really appreciate the community involvement :D

PR #1660, opened 2022-02-01
adds frame extraction to add_new_videos

louisadamian:
Adds frame extraction to add_new_videos for **only** the newly added videos. Modifies extract_frames to accept an optional list of strings containing paths, when that list is not empty extract_frames only extracts frames for the videos at those paths.

Currently there is no way to extract frames for only videos added after initial model creation with user input in the command line.

### DIFF
--- a/deeplabcut/create_project/add.py
+++ b/deeplabcut/create_project/add.py
@@ -112,5 +112,5 @@ def add_new_videos(config, videos, copy_videos=False, coords=None, extract_frame
             "New videos were added to the project and frames have been extracted for labeling!"
         )
     else:
-        print("New video was added to the project!")
+        print("New videos were added to the project! use Use the function 'extract_frames' to select frames for labeling.")
     auxiliaryfunctions.write_config(config, cfg)
--- a/deeplabcut/create_project/add.py
+++ b/deeplabcut/create_project/add.py
@@ -9,7 +9,7 @@ Licensed under GNU Lesser General Public License v3.0
 """
 
 
-def add_new_videos(config, videos, copy_videos=False, coords=None):
+def add_new_videos(config, videos, copy_videos=False, coords=None, extract_frames=True):
     """
     Add new videos to the config file at any stage of the project.
 
@@ -24,8 +24,13 @@ def add_new_videos(config, videos, copy_videos=False, coords=None):
     copy_videos : bool, optional
         If this is set to True, the symlink of the videos are copied to the project/videos directory. The default is
         ``False``; if provided it must be either ``True`` or ``False``.
+
     coords: list, optional
-      A list containing the list of cropping coordinates of the video. The default is set to None.
+        A list containing the list of cropping coordinates of the video. The default is set to None.
+
+    extract_frames: bool, optional
+        if this is set to True extract_frames will be run on the new videos
+
     Examples
     --------
     Video will be added, with cropping dimenions according to the frame dimensinos of mouse5.avi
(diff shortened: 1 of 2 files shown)


jeylau:
Thanks @louisadamian for the contribution! I like that behavior of auto-extracting frames upon adding new videos. We may still want to be able to turn it off for the people who extract their frames manually via the GUI, what do you think @MMathisLab?

louisadamian:
I can definitely add an `extract_frames` optional boolean argument

deeplabcut/create_project/add.py
```diff
@@ -99,8 +105,12 @@ def add_new_videos(config, videos, copy_videos=False, coords=None):
...
             cfg["video_sets_original"].update(params)
+    videos_str = [str(video) for video in videos]
+    if extract_frames:
+        frame_extraction.extract_frames(config, userfeedback=False, videos=videos_str)
 
     auxiliaryfunctions.write_config(config, cfg)
+
     print(
-        "New video was added to the project! Use the function 'extract_frames' to select frames for labeling."
+        "New videos were added to the project and frames have been extracted for labeling!"
```
aristotelis-economides: You might want to keep the new print statement within the `if` and put the original print statement in an `else` (so the message is correct depending on the bool value)

deeplabcut/generate_training_dataset/frame_extraction.py
```diff
@@ -142,6 +143,9 @@ def extract_frames(
         The index of the camera that already has extracted frames. This will match frame numbers to extract for all other cameras.
         This parameter is necessary if you wish to use epipolar lines in the labeling toolbox. Only use if mode = 'match' and config3d is provided.
 
+    videos: list, default: None
+            A list of the string containing full paths to videos to extract frames for. If this is left as None all videos in the config file will have frames extracted.
```
aristotelis-economides: minor typo here "A list of the string containing..." --> "A list of string**s** containing..."

aristotelis-economides (changes requested):
Nice stuff @louisadamian thanks for the contribution :) I mentioned some minor things

louisadamian:
thanks @intergalactic-mammoth

deeplabcut/create_project/add.py
aristotelis-economides: Is this fixed? If I'm not mistaken now the function will print `New videos were added to the project and frames have been extracted for labeling` even if `extract_frames` is False.

If it is false it should still print `New videos were added to the project! Use the function 'extract_frames' to select frames for labeling` no?

Or am I missing something?

aristotelis-economides (changes requested):

louisadamian:
@intergalactic-mammoth my bad. It's fixed now.

aristotelis-economides:
> @intergalactic-mammoth my bad. It's fixed now.

Nice! 🙌🔥

Maybe keep this part `Use the function 'extract_frames' to select frames for labeling` in the original print statement so that people can still get directions on what to do next ;)

Otherwise looks great to me!

aristotelis-economides (approved):

deeplabcut/create_project/add.py
```diff
@@ -112,5 +112,7 @@ def add_new_videos(config, videos, copy_videos=False, coords=None, extract_frame
             "New videos were added to the project and frames have been extracted for labeling!"
         )
     else:
-        print("New videos were added to the project! use Use the function 'extract_frames' to select frames for labeling.")
+        print(
+            "New videos were added to the project! use Use the function 'extract_frames' to select frames for labeling."
```
aristotelis-economides: typo -> "use Use"

AlexEMG (approved):

jeylau (approved):

PR #1664, opened 2022-02-03
Make the outer loop go over GT assemblies in evaluate_assembly

jeylau:
If no assemblies are predicted for a given frame, an empty list is passed to `match_assemblies` rather than continuing, extending the `unmatched` list. This guarantees that `ntot` is always correct, even in the extreme case where no animals are predicted at all.
https://github.com/DeepLabCut/DeepLabCut/blob/56e4a0662f8c07646f0ac5391333dfce9bc889ce/deeplabcut/pose_estimation_tensorflow/lib/inferenceutils.py#L1016

### DIFF
--- a/deeplabcut/pose_estimation_tensorflow/core/evaluate_multianimal.py
+++ b/deeplabcut/pose_estimation_tensorflow/core/evaluate_multianimal.py
@@ -480,19 +480,21 @@ def evaluate_multianimal_full(
 
                             print("##########################################")
                             print(
-                                "Average Euclidean distance to GT per individual (in pixels)"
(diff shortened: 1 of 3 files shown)


MMathisLab:
@jeylau can you address https://forum.image.sc/t/question-about-multi-animal-deeplabcut-evaluation-results-average-euclidean-distance-to-gt-per-bodypart-in-pixels/63058/3?u=mwmathis perhaps in this PR?

jeylau:
@MMathisLab, fixed now!

MMathisLab (approved):

PR #1675, opened 2022-02-08
Docs: Add recipes for varying number of animals & Using labels from other software

aristotelis-economides:
Created doc recipes for 
1. Using maDLC for a varying number of animals 
2. Using labels from other software


docs/recipes/VaryNumAnimals.md
```diff
@@ -0,0 +1,47 @@
...
+
+When running [`stitch_tracklets`](https://deeplabcut.github.io/DeepLabCut/docs/maDLC_UserGuide.html#optimized-animal-assembly-video-analysis) you will need to pass the correct number of animals in the `n_tracks` argument. 
+
+
+
+In both cases, bear in mind that the `auto_track=True` argument available in `analyze_videos` since the [2.2.0.3 release](https://github.com/DeepLabCut/DeepLabCut/releases/tag/v2.2.0.3) will not work when dealing with a varying number of animals.
+
```
jeylau, line 44: It does if you edit `individuals` in the config but that's a hack 🤫

aristotelis-economides: Not sure if I should write it, it may be confusing? Since @MMathisLab said "Don’t edit config I think is general better advice." haha

jeylau: No no sure, we wanted `auto_track` to work for the most common use case, and is definitely not meant to handle all scenarios.

MMathisLab:
@AlexEMG I did not approve this PR, nor did Jessy.

docs/recipes/UsingDataLabeledElsewhere.md
```diff
@@ -0,0 +1,31 @@
...
+
+You may have annotation data in different formats, yet want to use the DeepLabCut pipeline. In this case, you need to convert the data to our format. Simply, you can format your data in an excel sheet (.csv file) or pandas array (.h5 file).
+
+Here is a guide to do this via the ".csv" route: (the pandas array route is identical, just format the pandas array in the same way).
+
+1. Create a project as described in our [single animal](https://deeplabcut.github.io/DeepLabCut/docs/standardDeepLabCut_UserGuide.html) and [multi-animal](https://deeplabcut.github.io/DeepLabCut/docs/maDLC_UserGuide.html) user guides. <br><br>
+
+2. Edit the `config.yaml` file to include the body part names. Please take care that spelling, spacing, and capitalization are **identical** to the "labeled data body part names". <br><br>
+
+3. Inspect the excel formatted sheet (.csv) from our demo project. [[this file](https://github.com/AlexEMG/DeepLabCut/blob/master/examples/Reaching-Mackenzie-2018-08-30/labeled-data/reachingvideo1/CollectedData_Mackenzie.csv)] <br><br>
```
MMathisLab, line 11: this only works for single animal; you would need to also add a path for them to a maDLC file format, which is different.

docs/recipes/UsingDataLabeledElsewhere.md
```diff
@@ -0,0 +1,31 @@
...
+
+4. Edit the .csv file such that it contains the X, Y pixel coordinates, the body part names, the scorer name as well as the relative path to the image: e.g. `/labeled-data/somefolder/img17.jpg`, and make sure the scorer name, and body parts are the same in the `config.yaml` file.
```
MMathisLab, line 13: what about individual ID? Unique body parts?

docs/recipes/UsingDataLabeledElsewhere.md
```diff
@@ -0,0 +1,31 @@
...
+    #The expected entries in config.yaml are:
+    video-sets:
+        /dummy/path/that/doesnt/necessarily/exist/some-folder.avi:
+            crop: 0, 640, 0, 480
+        /videos/some-other-folder.avi:
+            crop: 0, 640, 0, 480
+    ```
+    <br><br>
+
+5. Run `deeplabcut.convertcsv2h5('path/to/config.yaml', scorer= 'experimenter')`. The **scorer name must be identical** to the input name for experimenter that you used when you created the project. In our [example demo notebook](https://colab.research.google.com/github/DeepLabCut/DeepLabCut/blob/master/examples/COLAB/COLAB_DEMO_mouse_openfield.ipynb), `scorer: Mackenzie` will automatically update to your name.
```
MMathisLab, line 31: does this function work for maDLC projects? @jeylau this is why I did not move over from the wiki; not sure this is tested.

docs/recipes/VaryNumAnimals.md
```diff
@@ -0,0 +1,47 @@
+# Varying the number of animals in multi-animal DeepLabCut
```
MMathisLab, line 1: bad title; it should be user focused: "Analyzing Videos with a different number of individuals than a model was trained with" would be better

docs/recipes/VaryNumAnimals.md
```diff
@@ -0,0 +1,47 @@
...
+
+In case you have videos of the same experimental setting where the number of animals is changing between videos, DeepLabCut can deal with that like a champ.
```
MMathisLab, line 3: champ is not a word ;)

docs/recipes/VaryNumAnimals.md
```diff
@@ -0,0 +1,47 @@
...
+  - miranda
+  - lucius
+  - vera
+  - muriel
+```
+
+If you cannot tell the unicorns apart:
+```yaml
+individuals:
+  - unicorn 1
```
MMathisLab, line 23: you should not have spaces in the names... unicorn1

docs/recipes/VaryNumAnimals.md
```diff
@@ -0,0 +1,47 @@
...
+  - unicorn 2
+  - unicorn 3
+  - unicorn 4
+  - unicorn 5
+  - unicorn 6
+```
+
+When labeling, it is always best to label hard frames, i.e. that have all the 6 animals. Labeling videos with less animals and even 1 works in principle, but practically assembly and tracking performance will be better when training with "hard frames" that have many occlusions and interactions. This is a general recommendation with DeepLabCut, try to label frames that reflect the statistics of luminance, movement & interactions.
```
MMathisLab, line 31: i..e --> i.e., 

what are hard frames? Frames with animals never touching but 6 animals are not "hard," please use concrete language.

docs/recipes/VaryNumAnimals.md
```diff
@@ -0,0 +1,47 @@
...
+  - unicorn 2
+  - unicorn 3
+  - unicorn 4
+  - unicorn 5
+  - unicorn 6
+```
+
+When labeling, it is always best to label hard frames, i.e. that have all the 6 animals. Labeling videos with less animals and even 1 works in principle, but practically assembly and tracking performance will be better when training with "hard frames" that have many occlusions and interactions. This is a general recommendation with DeepLabCut, try to label frames that reflect the statistics of luminance, movement & interactions. 
+
+## In case of an existing project
```
MMathisLab, line 33: not a clear workflow; please make a diagram for this recipe; there should be clear decision-points

docs/recipes/VaryNumAnimals.md
```diff
@@ -0,0 +1,47 @@
...
+
+When running [`stitch_tracklets`](https://deeplabcut.github.io/DeepLabCut/docs/maDLC_UserGuide.html#optimized-animal-assembly-video-analysis) you will need to pass the correct number of animals in the `n_tracks` argument. 
+
+
+
+In both cases, bear in mind that the `auto_track=True` argument available in `analyze_videos` since the [2.2.0.3 release](https://github.com/DeepLabCut/DeepLabCut/releases/tag/v2.2.0.3) will not work when dealing with a varying number of animals.
```
MMathisLab, line 43: it is not meant to deal with it, thus is "will not work" is not true; is is not meant for this use case. Revise.

docs/recipes/VaryNumAnimals.md
aristotelis-economides: The more descriptive title is great. In this case it implies an already trained model though -- this part "....than a model was trained with" -- and then the "In case of a new project" section is confusing? What if we leave "Analyzing videos with a different number of individuals" instead?

PR #1678, opened 2022-02-09
Clean up requirements.txt

jeylau:
- [x] Get rid of moviepy (which prevents users from having numpy >= 1.20, also causing errors with TF at times)
- [x] Remove jupyter-books (it's already being installed automatically when publishing our docs, and this will let users install ipython >= 8, which has a very nice (predictive) autocomplete feature)
- [x] Ensure scikit-learn >= 1.0
- [x] Pin tables to their newest release (notably with support for numpy 1.21 and HDF5 1.12)

And no ugly red crosses anymore!


aristotelis-economides (approved):
i like removing the need to update things in two places :muscle: :rocket:

setup.py
```diff
@@ -10,10 +10,20 @@
 """
 
 import setuptools
+from pathlib import Path
 
 with open("README.md", "r") as fh:
     long_description = fh.read()
 
+requirements = []
```
MMathisLab: is this required, or why the addition? @jeylau

MMathisLab (changes requested):
left one comment, then great!

setup.py
jeylau: My bad, I had undone the commit but this left this part. I'll remove it :)

MMathisLab (approved):

Issue #1696, opened 2022-02-21
UserGuide seems to have conflicting recommendations


poja:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Bug description

The `UseOverviewGuide` seems to have two recommendations regarding: "I have single animal videos, but I want to use" [DLC 2.2 features]. Also one of the recommendation has a typo (`False`) :)
So maybe there should be a more consistent way of deciding if to use `multianimal`?

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/DeepLabCut/DeepLabCut/blob/master/CODE_OF_CONDUCT.md)

MMathisLab:
@poja the typo is wrong, thank you, we will fix it! But, the other aspect is correct: if you want to use mutli-animal network features, such as PAFs, you need to set `multianimal: True`; then just label 1 individual.

PR #1699, opened 2022-02-22
clarification on when to use ma=false/true #1696 

aristotelis-economides:
fixed `multianimal:False` on the [User Overview Guide](https://github.com/DeepLabCut/DeepLabCut/blob/master/docs/UseOverviewGuide.md)

From this issue: https://github.com/DeepLabCut/DeepLabCut/issues/1696


docs/UseOverviewGuide.md
```diff
@@ -80,15 +80,12 @@ We highly recommend using 2.2 first in the Project Manager GUI ([Option 3](docs/
 ##### *What scenario do you have?*
 
 - **I have single animal videos, but I want to use the advanced tracking features & updated network capabilities introduced (for multi-animal projects) in DLC2.2:**
-   - quick start: when you `create_new_project` just set the flag `multianimal=True`.
+   - quick start: when you `create_new_project` just set the flag `multianimal=False`. You still get lots of upgrades!
```
MMathisLab: this is not correct; it is written correctly. You can use PAFs with single animals, but MA must be set to TRUE.

MMathisLab (dismissed):

aristotelis-economides:
@MMathisLab  oh I see, I was confused. I re-typed it so it is a bit more explicit, let me know if that works

MMathisLab:
most users will not know what this means; I prefer to leave it as it was ;)

MMathisLab:
well, you can leave open as to fix the typo ;)

MMathisLab (approved):

PR #1700, opened 2022-02-22
Adding greedy oks matching as option

LucZot:
Instead of using global matching, one can optionally use greedy matching of assemblies (as pycocotools does it); by adding a flag "greedy_oks" into the inference_cfg.yaml file.


jeylau:
Looks great to me @LucZot! One minor thing though: I think `greedy_matching` as an argument name would be more explicit, what do you think?

jeylau (approved):

AlexEMG (approved):

PR #1713, opened 2022-03-04
FIX: Attribute error when using skeleton tracking method

mdsmarte:
The tracking utils were refactored in PR #1576. Part of this refactor aligned the `n_frames` and `frame_count` attribute names, which represent the same concept, across the `SORT*` classes. The `SORTBase` class declares this attribute as `n_frames` but the `SORTSkeleton.track()` method still accesses it as `frame_count`. This causes an attribute error to be raised when using the skeleton tracking method: `AttributeError: 'SORTSkeleton' object has no attribute 'frame_count'`

This PR is a short fix to use the correct attribute name in the `SORTSkeleton.track()` method.


jeylau:
Great catch @mdsmarte, thank you very much!

PR #1719, opened 2022-03-09
edit: Enable to build skeleton from config

aristotelis-economides:
Enable using the `skeleton` defined in the config file as paf_graph like so:

`deeplabcut.create_multianimaltraining_dataset(config, paf_graph="config")`

### DIFF
--- a/deeplabcut/utils/auxiliaryfunctions.py
+++ b/deeplabcut/utils/auxiliaryfunctions.py
@@ -42,11 +42,14 @@ def create_config_template(multianimal=False):
         uniquebodyparts:
         multianimalbodyparts:
         bodyparts:
+        \n
+    # Fraction of video sto start/stop when extracting frames for labeling/refinement
         start:
         stop:
         numframes2pick:
         \n
     # Plotting configuration
+    # note: skeleton is only used for plotting, not for assembly/tracking
         skeleton:
         skeleton_color:
         pcutoff:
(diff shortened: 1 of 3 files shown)


deeplabcut/utils/auxiliaryfunctions.py
```diff
@@ -90,6 +93,8 @@ def create_config_template(multianimal=False):
     # Annotation data set configuration (and individual video cropping parameters)
         video_sets:
         bodyparts:
+        \n
```
jeylau, line 95: video sto?

AlexEMG:
_Here is the skeleton builder:_ https://github.com/DeepLabCut/DeepLabCut/blob/420b3525abba055f3f5f629599e520f6a64a696c/deeplabcut/utils/skeleton.py#L151

_Now create_training_set gets paf_graph only as argument_ 
https://github.com/DeepLabCut/DeepLabCut/blob/420b3525abba055f3f5f629599e520f6a64a696c/deeplabcut/generate_training_dataset/multiple_individuals_trainingsetmanipulation.py#L102

**Modes:**
- None (default) --> PAF graph will be optimized down the line 
- pass skeleton (as you like it)
- pass 'config' and then it pulls it from the config (skeleton)

Make sure the docs make all that clear!

aristotelis-economides:
( some formatting changes in multiple_individuals_trainingsetmanipulation.py are caused by black )

jeylau (approved):

AlexEMG (approved):

PR #1744, opened 2022-03-28
Address an off by one error in the TQDM progress bar in the predict steps

hmaarrfk:
I've been a little obsessed with getting progress bars for our workflows.

I noticed an off by 1 error in the `predict_video` step.

For my 900 frame video, the tqdm progress bar increases incrementally up to 900, then at the "900th" frame, the progress bar shows that it is analyzing the 910th frame (step size is 10 according to `GetPoseF_GTF`.

image

I believe this is because `tqdm` expects you to call update AFTER the step has been completed. The following "1 liner fix" address the progress bar issue.

I hope this helps.


aristotelis-economides:
Thanks for this contribution @hmaarrfk!

MMathisLab (approved):

AlexEMG (approved):

PR #1751, opened 2022-04-04
Add function for handling too many extracted labels

frankiechang123:
Changes:
1.  Bug fix in `dropimagesduetolackofannotation`;
2.  Add function drop_unlabeled_frames; 
3. The related functions now handles `FileNotFound`, when no h5 annotation files exists 

Before the change, `dropimagesduetolackofannotation` will trigger error : `AttributeError: 'tuple' object has no attribute 'split'`. Bug fix is added. The function now deletes all the extracted png files for which the h5 files have no entries. 

`drop_unlabeled_frames` is a new function that drop all the entries in the h5 file and the csv file such that no body part is labeled. 

Sometimes I accidentally extract too many frames to label. These functions helped me get rid of the unwanted frames. It would be great if some of these features could be integrated into the labeling GUI. 

Thank you!

### DIFF
--- a/deeplabcut/__init__.py
+++ b/deeplabcut/__init__.py
@@ -67,7 +67,7 @@ from deeplabcut.generate_training_dataset import (
     dropimagesduetolackofannotation,
     adddatasetstovideolistandviceversa,
     dropduplicatesinannotatinfiles,
-    drop_unlabeled_frames
+    dropunlabeledframes
 )
 from deeplabcut.utils import (
     create_labeled_video,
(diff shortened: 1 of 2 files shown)


AlexEMG:
- Thanks for the fix! 
- Doesn't: `dropimagesduetolackofannotation` do what you want?

frankiechang123:
Hi!
`dropimagesduetolackofannotation` reads in the h5 file and deletes all images with no record in the h5 file. However, when I pressed "save" button in the labeling GUI, records for all images would be saved in the h5 file with the non-annotated ones having NaN values for each body part coordinate. As a result, no images or annotation record will be dropped.

 `dropunlabeledframes` drops rows with all NaN values in the h5file. Then `dropimagesduetolackofannotation` will then drop the unwanted images. I'm not sure if it's better to just integrate these two functions into one. 

Thank you

AlexEMG (approved):

MMathisLab (approved):

PR #1760, opened 2022-04-07
Match order of individuals in DataFrames accross views

aristotelis-economides:
- Makes sure that the order of individuals in the DataFrame of view1 is kept in view2 (applies the matching)
- Minor fixes


jeylau:
Sweet!

PR #1772, opened 2022-04-11
Delay torch import

jeylau:
torch import is now delayed until the moment a module from `pose_tracking_pytorch` is imported, allowing normal use of DeepLabCut without having to install PyTorch.
`transformer_reID` is readily accessible via `from deeplabcut.pose_tracking_pytorch import transformer_reID`.
(NB: I have just edited a cell of the notebook, I do not know why the diff is so big)

### DIFF
--- a/deeplabcut/__init__.py
+++ b/deeplabcut/__init__.py
@@ -96,10 +96,6 @@ from deeplabcut.pose_estimation_tensorflow import (
     export_model,
 )
 
-from deeplabcut.pose_tracking_pytorch import(
-    transformer_reID
-)
-
 
 from deeplabcut.pose_estimation_3d import (
     calibrate_cameras,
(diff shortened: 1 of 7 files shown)


MMathisLab:
in principle I like it, but also I worry this sort of breaks the flow of normal usage to go from:

```python
deeplabcut.transformer_reID(config, videos_to_analyze, n_tracks=None, videotype="mp4")
```

to 

```python
from deeplabcut.pose_tracking_pytorch import transformer_reID
transformer_reID(config, videos_to_analyze, n_tracks=None, videotype="mp4")
```

no way to keep this intact?

jeylau:
Found a neat way @MMathisLab!

AlexEMG (approved):
Looks good!

PR #1704, opened 2022-02-23
Freshening up `Getlistofvideos`

sneakers-the-rat:
re: https://github.com/DeepLabCut/DeepLabCut/issues/1703

* type hints & docstring update
* make `videotype` optional when given a list of files
* coerce strings to list at start
* use shuffle instead of sampling with list of videos
* don't change directory
* unify filtering into one operation at end of function

PR #1773, opened 2022-04-11
Improved behavior of get_list_of_videos

jeylau:
Follow-up of #1704 ([this one comment](https://github.com/DeepLabCut/DeepLabCut/pull/1704#issuecomment-1049300871) specifically)

- [x] videotype now defaults to "", preserving the old behavior when passing a list of videos, and filtering videos with common extensions when passing a directory
- [x] add option to specify multiple extensions
- [x] unit tested

### DIFF
--- a/deeplabcut/create_project/modelzoo.py
+++ b/deeplabcut/create_project/modelzoo.py
@@ -68,7 +68,7 @@ def create_pretrained_human_project(
     videos,
     working_directory=None,
     copy_videos=False,
-    videotype=".mp4",
+    videotype="",
     createlabeledvideo=True,
     analyzevideo=True,
 ):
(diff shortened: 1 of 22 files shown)


AlexEMG (approved):
This is great, and note for future, it's easy to just expand:

`SUPPORTED_VIDEOS = 'avi', 'mp4', 'mov', 'mpeg', 'mkv'`

This list is not comprehensive, but rather everything that works with OpenCV will work! Feel free to PR, if you want to suggest other endings! 

See: https://en.wikipedia.org/wiki/Video_file_format

Maybe worth adding a few of those, I'm about to depart at Geneve train station and would otherwise do it.

AlexEMG (commented):
Great commit name

jeylau (commented):
🦾

PR #1786, opened 2022-04-22
More visible DLC version

jeylau:
I find that the DLC version is often under-reported (especially when raising issues). It will now appear in the terminal upon importing DeepLabCut or on the Welcome page of the GUI, helping users report it more systematically 😃


MMathisLab (approved):

hmaarrfk:
Would it be acceptable to move this to to `launch_dlc` instead of the `__init__`? I find init time print statements to be very jarring when one is past the exploration phase of development.

hmaarrfk:
I really do sympathies with your desire to get the version information. It is really hard to help people when you have to guess at the version they are using based on the feature set or other typos in their screenshots or logs.

PR #1816, opened 2022-05-02
Update publish-book.yml

MMathisLab:
seems the gui install fails; unclear as it passed on PR branch (https://github.com/DeepLabCut/DeepLabCut/commit/4a1d50041dd0cd233d1d96d7c316343500b7c3a8), but seems to fail on main; any ideas @jeylau @rahulporuri ?

### DIFF
--- a/.github/workflows/publish-book.yml
+++ b/.github/workflows/publish-book.yml
@@ -19,7 +19,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        python -m pip install -e .[gui,docs]
+        python -m pip install -e .[docs]
         pip install jupyter-book
 
     - name: Build the book
--- a/docs/standardDeepLabCut_UserGuide.md
+++ b/docs/standardDeepLabCut_UserGuide.md
@@ -159,13 +159,6 @@ Keyboard arrows: advance frames
 delete key: delete label
 ```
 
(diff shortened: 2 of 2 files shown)


rahulporuri:
Sigh. I think the reason is [lack of wheels](https://pypi.org/project/wxPython/4.0.7.post2/#files) on linux. As you can see in the [job log](https://github.com/DeepLabCut/DeepLabCut/runs/6257071258?check_suite_focus=true), during the wxpython installation, the job was trying to build the package from the tar gz source archive instead of using a wheel. See [official installation instructions](https://wxpython.org/pages/downloads/index.html). We're using ubuntu 20.04 for the job and it looks like [wxpython<4.1 doesn't exist for ubuntu 20.04](https://extras.wxpython.org/wxPython4/extras/linux/gtk3/ubuntu-20.04/).

I put off documenting `deeplabcut.label_frames` and `deeplabcut.refine_labels` for this reason - the modules depend on `wx` and the lack of `wx` in the Python environment when building the docs raises errors.

Along with the changes in this PR, I would recommend removing the "API docs" sections for the above functions from the user guide - otherwise they'll just be empty in the published docs.

MMathisLab:
yes, agree - and we are changing the GUI now to drop wxPython, so will be easy to put back soon :)

PR #1817, opened 2022-05-02
Avoid repeated memory allocations in OpenCV when running inference

hmaarrfk:
@jeylau this should preallocate enough memory to read a single frame, and convert it to RGB, before running it through TF. 

Generally speaking, repeatedly calling memory allocation routines is "bad". I haven't had a chance to benchmark this, but in other workflows I have this optimization is quite desirable especially when one gets to long videos.

### DIFF
--- a/deeplabcut/pose_estimation_tensorflow/predict_videos.py
+++ b/deeplabcut/pose_estimation_tensorflow/predict_videos.py
@@ -45,6 +45,19 @@ from deeplabcut.pose_estimation_tensorflow.core.openvino.session import (
 # Loading data, and defining model folder
 ####################################################
 
+def cv_dtype_to_numpy_dtype(cv_dtype):
+    cv_dtype = int(cv_dtype)
+    # https://github.com/opencv/opencv/blob/d861c03366a0883b2a84ff4a2de06e523891373d/modules/videoio/src/cap_v4l.cpp#L1638
+
+    return {
+        cv2.CV_8U: np.uint8,
+        cv2.CV_8S: np.int8,
+        cv2.CV_16U: np.uint16,
+        cv2.CV_16S: np.int16,
+        cv2.CV_32S: np.int32,
+        cv2.CV_32F: np.float32,
+        cv2.CV_64F: np.float64,
+    }[cv_dtype]
 
 def create_tracking_dataset(
     config,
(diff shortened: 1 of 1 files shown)


hmaarrfk:
It is likely that the `view` of the original array is copied as well. I'll try to address this once these tests pass.

jeylau:
Thanks @hmaarrfk, that looks great! Do you want me to help you finalize the implementation in other places or are you running some tests?
What do you think about getting rid of cv2.cvtColor and frame_rgb altogether, and simply use `frame_bgr[..., ::-1]`?

deeplabcut/pose_estimation_tensorflow/predict_videos.py
```diff
@@ -868,15 +885,20 @@ def GetPoseF_GTF(cfg, dlc_cfg, sess, inputs, outputs, cap, nframes, batchsize):
...
-                )
+                # If we need cropping, get a view of the bgr frame
+                frame_cropped_bgr = frame_bgr[cfg["y1"]: cfg["y2"], cfg["x1"]: cfg["x2"], :]
             else:
-                frames[batch_ind] = img_as_ubyte(frame)
+                # Otherwise, just point to the original bgr frame
+                frame_cropped_bgr = frame_bgr
+
+            if frame_cropped_bgr.dtype != np.uint8:
+                cv2.cvtColor(frame_cropped_bgr, cv2.COLOR_BGR2RGB, dst=frame_rgb)
```
hmaarrfk, line 898: >  What do you think about getting rid of cv2.cvtColor and frame_rgb altogether, and simply use frame_bgr[..., ::-1]?

To be honest, many places don't deal well with slices, they well with "strided" data.

Here, I'm willing to bet that if you cropped the frame, that a contiguous copy of `fram_cropped_bgr` will be made, creating a short lived temporary input for `cv2.cvtColor`.

That said, tensorflow is quite fast, so who knows.

Do you have a benchmark we can run? That would be the best way.

MMathisLab: https://deeplabcut.github.io/DLC-inferencespeed-benchmark/

MMathisLab: https://github.com/DeepLabCut/DLC-inferencespeed-benchmark

hmaarrfk: Thanks. I can try to find how to run these and compare before and after.

jeylau:
@hmaarrfk, I'm realizing now that changes were not made to `GetPoseandCostsF`, but only to the single-animal part of the inference code. The multi-animal part is found in a separate file: https://github.com/DeepLabCut/DeepLabCut/blob/cd97518d6db4dfef479f53d4c1d5b35a0e097dd5/deeplabcut/pose_estimation_tensorflow/predict_multianimal.py#L346-L348

hmaarrfk:
ah thanks for that pointer!

hmaarrfk:
Hmm, I guess I didn't notice it becuase it uses a `VideoWriter` wrapper (which seems like it should really be using the `VideoReader` wrapper) of opencv functions. I'll see what i can do in the coming future. I guess the first thing would be to run the benchmarks and see if any of this is worth it. It is often difficult to see benefit with "small" video files, so I'll have to see :/

jeylau:
I have just tested this on a 15k frame video (1920x1080), and the gain is quite notable (reading speed is 420 vs 345 FPS). Perhaps worth merging?

MMathisLab (approved):

MMathisLab:
@n-poulsen can you fix merge conflicts and run a test yourself then we can merge?

deeplabcut/pose_estimation_tensorflow/predict_videos.py
jeylau, line 898: I hadn't tried it with cropped frames, but looking back at it, it seems like this line would cause some shape mismatch issue, as`frame_cropped_bgr` and `dst` are of different dimensions.

MMathisLab:
We are working on the roadmap item to move to PyTorch, and we will be sure to consider this in that release, thanks @hmaarrfk !

hmaarrfk:
Thanks for the update!

PR #1823, opened 2022-05-07
Ignore failing "wx" imports when building api docs

rahulporuri:
Previously, we were having difficulty building api documentation for functions that are defined in modules which rely on the `wx` package.

This PR configures `sphinx` to ignore the failing `wx` imports when building the api documentation.

This PR reverts changes in https://github.com/DeepLabCut/DeepLabCut/pull/1816.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -4,6 +4,8 @@ logo: docs/images/logo.png
 only_build_toc_files: true
 
 sphinx:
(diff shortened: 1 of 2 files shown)


rahulporuri:
@AlexEMG and @MMathisLab can you take a look at this when you have the chance?

PR #1824, opened 2022-05-07
Cleanup internal links in documentation

rahulporuri:
There were a number of places where the jupyter book documentation linked to external files on github instead of linking to the same files on jupyter book. This PR fixes that issue. Additionally, in some cases internal links were defined using a path to the file - these too have been updated to use the new references.

Two broken links were also fixed in this commit and a broken link was removed as there is no correct equivalent.

Ref https://jupyterbook.org/en/stable/tutorials/references.html for information about references in jupyter book.

With the changes in this PR, there are no links to documentation on GitHub from the Jupyter Book documentation.

Note that all changed links were manually checked but it would definitely be good to have another pair of eyes.

### DIFF
--- a/docs/Governance.md
+++ b/docs/Governance.md
@@ -1,3 +1,4 @@
+(governance-model)=
 # Governance Model of DeepLabCut
 (adapted from https://napari.org/docs/_sources/developers/GOVERNANCE.md.txt)
 
@@ -84,7 +85,7 @@ The current steering council of DeepLabCut consists of the original developers:
 The SC membership is revisited every January. SC members who do
 not actively engage with the SC duties are expected to resign. New members are
 added by nomination by a core developer. Nominees should have demonstrated
-long-term, continued commitment to the project and its [mission and values](https://github.com/DeepLabCut/DeepLabCut/docs/MISSION_AND_VALUES.md). A
+long-term, continued commitment to the project and its [mission and values](mission-and-values). A
 nomination will result in discussion that cannot take more than a month and
 then admission to the SC by consensus. During that time deadlocked votes of the SC will
 be postponed until the new member has joined and another vote can be held.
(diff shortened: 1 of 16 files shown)


rahulporuri:
@AlexEMG and @MMathisLab can you take a look at this when you have the chance?

MMathisLab:
hi @rahulporuri I agree it would be good to clean up. I didn't do it because then the docs on github don't look that nice, and I believe many people still use github repo docs vs. the prefered: https://deeplabcut.github.io/DeepLabCut/docs/intro.html:

i.e., the tags show on the markdown; maybe it's not terrible, but I would love a clean way to do it :) 

Screen Shot 2022-05-12 at 6 00 16 PM

MMathisLab (approved):

rahulporuri:
> I didn't do it because then the docs on github don't look that nice, and I believe many people still use github repo docs vs. the prefered: https://deeplabcut.github.io/DeepLabCut/docs/intro.html

Damn. I didn't realize this when I proposed these changes. I worked on this as I found moving between the preferred docs and the markdown files on GitHub a bit jarring and because it made navigating the documentation more difficult than it needed to be.

Please feel free to ignore and close this PR if this isn't the best for your project.

> i.e., the tags show on the markdown; maybe it's not terrible, but I would love a clean way to do it :)

I don't know of a clean way to do it at the moment. Markdown doesn't recognize these tags as they're specific to Jupyter Book/MyST which is why I think they get rendered in the markdown documents directly.

MMathisLab:
Hey @rahulporuri I think it's still a great idea though; if anything we should suggest people to use the jupyter book docs vs. github versions; @jeylau what are your thoughts?

jeylau:
I agree @MMathisLab, the jupyter book should be the reference doc; even more so now that they also include API documentation by @rahulporuri ☺️

PR #1830, opened 2022-05-12
Avoid KeyError during feature loading

yeshaokai:
#1821  Though this is rare, but it does happen that the miner tries to access frames that do not have any animal in it (thus not saved into the dictionary in the first place).


jeylau (approved):

jeylau:
Looks good @yeshaokai, thanks!

PR #1873, opened 2022-06-10
Bug fix for adding videos on windows

tomsains:
Adding a new movie to be labelled on windows throws an operating system error. 
This was because in add.py the symbolic link creation function is different from that in new.py
I have updated it to catch the operating system error and use a different method


MMathisLab (approved):

MMathisLab:
Thanks for catching this!

Issue #1843, opened 2022-05-20
Controlling save location of pretrained network


nmtimme:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Bug description

Following up on issue #1828 , is there a way to control where the pretrained network is saved during training data set creation? I have been unable to resolve that issue (#1828) using the solution suggested. My main writable directory on the cluster is bound and the container version of the software can created directories in analysis steps before creation of the training data set. Perhaps the directory structures on a cluster were (understandably) not anticipated when designing the container? Thanks!

~Nick

### Operating System

Red Hat Enterprise Linux 7

### DeepLabCut version

2.2.0.2

### DeepLabCut mode

single animal

### Device type

gpu

### Steps To Reproduce

Using singularity, in the latest-gui docker, run...
ipython
import deeplabcut
deeplabcut.create_training_dataset(ConfigPath, augmenter_type='imgaug')

### Relevant log output

```shell
Downloading a ImageNet-pretrained model from http://download.tensorflow.org/models/resnet_v1_50_2016_08_28.tar.gz....
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
<ipython-input-4-128412801f55> in <module>
----> 1 deeplabcut.create_training_dataset(ConfigPath, augmenter_type='imgaug')

/usr/local/lib/python3.8/dist-packages/deeplabcut/generate_training_dataset/trainingsetmanipulation.py in create_training_dataset(config, num_shuffles, Shuffles, windows2linux, userfeedback, trainIndices, testIndices, net_type, augmenter_type)
    798         dlcparent_path = auxiliaryfunctions.get_deeplabcut_path()
    799         defaultconfigfile = os.path.join(dlcparent_path, "pose_cfg.yaml")
--> 800         model_path, num_shuffles = auxfun_models.Check4weights(
    801             net_type, Path(dlcparent_path), num_shuffles
    802         )

/usr/local/lib/python3.8/dist-packages/deeplabcut/utils/auxfun_models.py in Check4weights(modeltype, parent_path, num_shuffles)
     63         else:
     64             if not model_path.is_file():
---> 65                 Downloadweights(modeltype, model_path)
     66 
     67     return str(model_path), num_shuffles

/usr/local/lib/python3.8/dist-packages/deeplabcut/utils/auxfun_models.py in Downloadweights(modeltype, model_path)
     89         response = urllib.request.urlopen(url)
[...]

PR #1874, opened 2022-06-10
Simplify checking and downloading weights

rahulporuri:
Contributes towards #1843 

This PR makes two big changes:

- Simplifies the logic in `Check4weights` to use a dictionary constant from the module to figure out the path where the weight files should be for a given model type
- Simplifies the logic in `Downloadweights`. This simplification became possible with the changes introduced to `Check4weights`

The changes made are backed up by tests, which are also added to the codebase in this PR.

Finally, here is a simple example script that a user can run to download model weights for a given model type ahead of time.

```
import sys

from deeplabcut.utils.auxfun_models import MODELTYPE_FILEPATH_MAP, Downloadweights


def main():
    # Multiple model types can be provided separated by spaces.
    modeltypes = sys.argv[1:]
    valid_modeltypes = list(MODELTYPE_FILEPATH_MAP.keys())
    for modeltype in modeltypes:
        if modeltype in valid_modeltypes:
            Downloadweights(modeltype)
        else:
            print(
                f"{modeltype} is not valid. Valid model types are {valid_modeltypes}"
            )


if __name__ == "__main__":
    main()
```

As mentioned earlier, this doesn't fully fix the issue described in #1843 but it simplifies the code, enabling a simple/elegant solution.

Note to reviewer : It might be easy to review this PR one commit at a time.

### DIFF
--- a/deeplabcut/utils/auxfun_models.py
+++ b/deeplabcut/utils/auxfun_models.py
@@ -47,12 +47,15 @@ def Check4weights(modeltype, parent_path, num_shuffles):
     model_path = parent_path / MODELTYPE_FILEPATH_MAP[modeltype]
 
     if not model_path.exists():
-        Downloadweights(modeltype)
+        if "efficientnet" in modeltype:
+            Downloadweights(modeltype, model_path.parent)
+        else:
+            Downloadweights(modeltype, model_path)
 
     return str(model_path), num_shuffles
 
 
-def Downloadweights(modeltype, model_path=None):
+def Downloadweights(modeltype, model_path):
     """
     Downloads the ImageNet pretrained weights for ResNets, MobileNets et al. from TensorFlow...
     """
@@ -60,8 +63,9 @@ def Downloadweights(modeltype, model_path=None):
     import tarfile
     from io import BytesIO
 
+    target_dir = model_path.parents[0]
     neturls = auxiliaryfunctions.read_plainconfig(
-        MODEL_BASE_PATH / "pretrained_model_urls.yaml"
+        target_dir / "pretrained_model_urls.yaml"
     )
     try:
         if "efficientnet" in modeltype:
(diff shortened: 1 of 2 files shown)


rahulporuri:
@jeylau here are some changes I had in mind when I originally started exploring the codebase. In my opinion, the changes also help towards reaching a solution for #1843 .

MMathisLab (approved):

deeplabcut/utils/auxfun_models.py
```diff
@@ -14,55 +14,43 @@
...
-                + str(modeltype)
-                + "_224.ckpt",
-            )
-        )
-    elif "efficientnet" in modeltype:
-        model_path = Path(
-            os.path.join(
-                parent_path,
-                "pose_estimation_tensorflow/models/pretrained/"
-                + modeltype.replace("_", "-"),
```
rahulporuri, line 48: Note to reviewers : This PR effectively drops support for modeltype values `efficientnet_bX` where X is [0, 6] and only supports providing modeltype as `efficientnet-bX`.

I should have pointed this out earlier @MMathisLab .

To maintain backwards compatibility, should we add the relevant `efficientnet_bX` keys to the dictionary in this PR? I didn't include them because I couldn't find any documented uses of `efficientnet_bX` which is why I ended up dropping them from the dictionary.

MMathisLab: I think actually wasn't used publicly (maybe for a dev repo of the effnets, thanks for the catch!)

AlexEMG (changes requested):
looks great, I just recommend to move the unit-tests to /tests.

AlexEMG:
I moved the unit tests, in this branch: https://github.com/DeepLabCut/DeepLabCut/pull/1886 
(I couldn't contribute to your local branch). I'll merge it there. Thanks so much!!

PR #1880, opened 2022-06-16
Fix TypeError for create_training_dataset when training videos are named by number

hongh-zhang:
Hi DeepLabCut team,

My training videos are named '1.mp4', '2.mp4' etc., which caused their filename/dirname to be interpreted as integers in the create_training_dataset function. (multi-animal version 2.2.1)
TypeError

So I added a simple type conversion to ensure they are read as strings and fixed it in my attempt.

### DIFF
--- a/deeplabcut/utils/conversioncode.py
+++ b/deeplabcut/utils/conversioncode.py
@@ -74,7 +74,6 @@ def convertcsv2h5(config, userfeedback=True, scorer=None):
                 else:
                     index_col = 0
                 data = pd.read_csv(fn, index_col=index_col, header=header)
-                data.index = data.index.set_levels(data.index.levels[1].astype(str), level=1)
                 data.columns = data.columns.set_levels([scorer], level="scorer")
                 guarantee_multiindex_rows(data)
                 data.to_hdf(fn.replace(".csv", ".h5"), key="df_with_missing", mode="w")
@@ -241,6 +240,8 @@ def guarantee_multiindex_rows(df):
             df.index = pd.MultiIndex.from_tuples(splits)
         except TypeError:  #  Ignore numerical index of frame indices
             pass
+    
+    df.index = df.index.set_levels(df.index.levels[1].astype(str), level=1)  # set folder name to str
 
 
 def robust_split_path(s):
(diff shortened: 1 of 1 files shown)


hongh-zhang:
Hi,

I just found the problem is actually caused by the **convertcsv2h5** function. Pandas interpret the folder name as an integer automatically when loading the CSV. So I've added a line there for setting folder name to string and deleted previous commits.

It seems labeling by GUI does not cause the same problem.

jeylau:
Hi @hongh-zhang, I can still reproduce the error after your fix; this is because it happens early when pandas reads the h5 file rather than during conversion. I feel that it'd be better to edit `guarantee_multiindex_rows`, so we'd address your issue in all places in the code at once. What do you think?

jeylau:
For example, something like this should work:
```python
def guarantee_multiindex_rows(df):
    # Make paths platform-agnostic if they are not already
    if not isinstance(df.index, pd.MultiIndex):  # Backwards compatibility
        path = df.index[0]
        try:
            sep = "/" if "/" in path else "\\"
            splits = tuple([list(map(str, row)) for row in df.index.str.split(sep)])
            df.index = pd.MultiIndex.from_tuples(splits)
        except TypeError:  #  Ignore numerical index of frame indices
            pass
```

hongh-zhang:
> For example, something like this should work:
> 
> ```python
> def guarantee_multiindex_rows(df):
>     # Make paths platform-agnostic if they are not already
>     if not isinstance(df.index, pd.MultiIndex):  # Backwards compatibility
>         path = df.index[0]
>         try:
>             sep = "/" if "/" in path else "\\"
>             splits = tuple([list(map(str, row)) for row in df.index.str.split(sep)])
>             df.index = pd.MultiIndex.from_tuples(splits)
>         except TypeError:  #  Ignore numerical index of frame indices
>             pass
> ```

Thanks so much @jeylau! I agree it's much better to fix it there and it works!

But when `convertcsv2h5` is called the second time (well, for some dumb reasons) the TypeError can still occur because `guarantee_multiindex_rows` is now skipped since the csv is already in multiindex format.

Would it be better if I add a type conversion at the end of `gurantee_multiindex_rows`?

jeylau:
Hello @hongh-zhang, I liked your change very much, but I realize that it fails e.g. if the index contains frame indices (as is the case in the final h5 output file) because the index then only contains one level. I'll fix your code and merge.

PR #1905, opened 2022-07-14
use filter instead of for loop

maldil:
Using Python `filter` operator rather than Python for loops is more pythonic and succinct. What do you think about the change?


jeylau:
Thanks @maldil for your suggested change. What about the following list comprehension:
`cpbpts = [bp for bp in allbpts if bp in comparisonbodyparts]`
It is pythonic, succinct, and more readable?

maldil:
For me, using `filter` with a `lambda` function is more readable than `list comp`. However, if you like `list` comp, I am happy to do the change.

PR #1907, opened 2022-07-15
Update pose_cfg.yaml template

sfmig:
We (@sabrinabenas and I) think there might be a bug in how the pose_config template is read for the contrast and convolution parameters for data augmentation. With the template as default, the contrast and convolution methods are always evaluated as False and not added to the pipeline (see pose_imgaug.py)


jeylau:
Looks good to me!

Issue #1895, opened 2022-07-02
Design considerations - 'deeplabcut' Python package and DeepLabCut Application


rahulporuri:
Note that this is a meta-issue. I'm not discussing one-specific issue here. Rather, i'm trying to document big picture/design decisions in the current package/application, decisions which add friction when integrating `deeplabcut` into external applications. I'd like comments/feedback from the authors/maintainers regarding these points.

In this issue, I will be using "DeepLabCut" to refer to the end-user Application and `deeplabcut` to refer to Python package.

Based on my understanding, DeepLabCut was built as an end user application. There are a number of design considerations in the `deeplabcut` package that make sense from the point of view of DeepLabCut. But, when an external application tries to integrate `deeplabcut`, there are a number of awkward workarounds/misfits.

For example, DeepLabCut controls the configuration file, the directory containing the results and the filenames of the result files. The functions in `deeplabcut` used by DeepLabCut create pandas DataFrames, which are what get saved to disk, but they are not returned by the functions.

Another example is the type hints in the package. Widespread adoption of type hints in the package might not be necessary but adding type hints for all top-level functions should improve usability of `deeplabcut` in external applications (e.g. documented in https://deeplabcut.github.io/DeepLabCut/docs/standardDeepLabCut_UserGuide.html).

Another example is `print` calls to `sys.stdout` in `deeplabcut`. DeepLabCut needs to convey information to the user but the use of `print` calls in `deeplabcut` mean that applications that don't  want to print these messages to `stdout` will have to resort to monkeypatching `sys.stdout`/`sys.stderr` to capture the `print` messages and redirect them as `logging` messages. This can be resolved by adding a logger to the DeepLabCut application and replacing the `print` calls with `logger.*` calls e.g. `logger.info`/`logger.error`/`logger.warning`.

Another example is uniform usage of `str` or `pathlib.Path` objects as filepaths to configuration file, video files, destination folders etc. At the moment, it's not clear from the docstrings in `deeplabcut` whether the functions accept `pathlib.Path` objects, resulting in trial-and-error to figure this out (or digging into the code) or resulting in the external application being conservative by only passing string paths.

Another example is the separation of library dependencies vs application dependencies, specifically `matplotlib`. Users of the `deeplabcut` package might not be interested in functionality deeplabcut provides using `matplotlib` so preventing `matplotlib` imports from modules that are "core" to deeplabcut would be useful.

PR #1908, opened 2022-07-16
"deeplabcut.filterpredictions" and "deeplabcut.analyze_skeleton" return mappings

rahulporuri:
`deeplabcut.filterpredictions` at the moment returns `None`. Internally, there are two `try/except` blocks. The first try/except block looks for existing filtered data. The second try/except block looks for analyzed data to filter.

This PR does the following - 

* In the first commit, the second/inner try/except block is pulled out of the first try/except block. The first try/except block continues to the next video in the for loop if filtered data is found.
* In the second commit, the scope of the second try/except block is narrowed. Instead of finding and filtering the analyzed data in the try block, the try block only looks for the analyzed data now and the filtering happens outside.
* In the third and final commit, the PR stores the mapping between video filename and the corresponding filtered data. If no videos are found, an empty dictionary is returned and if analyzed data isnt found, `None` is returned for the corresponding video filename.

Similarly, `deeplabcut.analyze_skeleton` at the moment returns `None`.

Similar to change in `deeplabcut.filterpredictions` earlier, the scope of the `try/except` is also reduced - now the skeleton calculations are performed outside the try/except block. And `deeplabcut.analyze_skeleton` returns a mapping from the video filename to the corresponding skeleton data. If analyzed data isn't found for a given video filename, `None` will be used and if a pre-existing skeleton data file is found, a dataframe will be loaded from the file.

This addresses one of the points mentioned in #1895 - `deeplabcut` functions returning information instead of returning `None`.

Note to reviewer : It might be easier to review this PR one-commit-at-a-time.

### DIFF
--- a/deeplabcut/post_processing/analyze_skeleton.py
+++ b/deeplabcut/post_processing/analyze_skeleton.py
@@ -177,6 +177,7 @@ def analyzeskeleton(
     destfolder=None,
     modelprefix="",
     track_method="",
+    return_data=False,
 ):
     """Extracts length and orientation of each "bone" of the skeleton.
 
@@ -227,6 +228,9 @@ def analyzeskeleton(
         Empty by default (corresponding to a single animal project).
         For multiple animals, must be either 'box', 'skeleton', or 'ellipse' and will
         be taken from the config.yaml file if none is given.
+
+    return_data: bool, optional, default=False
+        If True, returns a dictionary of the filtered data keyed by video names.
 
     Returns
     -------
@@ -292,7 +296,9 @@ def analyzeskeleton(
         skeleton.to_hdf(output_name, "df_with_missing", format="table", mode="w")
         if save_as_csv:
             skeleton.to_csv(output_name.replace(".h5", ".csv"))
-    return video_to_skeleton_df
+
+    if return_data:
+        return video_to_skeleton_df
 
 
 if __name__ == "__main__":
--- a/deeplabcut/post_processing/analyze_skeleton.py
+++ b/deeplabcut/post_processing/analyze_skeleton.py
@@ -230,12 +230,19 @@ def analyzeskeleton(
 
     Returns
     -------
-    None
+    video_to_skeleton_df
+        Dictionary mapping video filepaths to skeleton dataframes.
+
+        * If no videos exist, the dictionary will be empty.
+        * If a video is not analyzed, the corresponding value in the dictionary will be
+          None.
     """
     # Load config file, scorer and videos
     cfg = auxiliaryfunctions.read_config(config)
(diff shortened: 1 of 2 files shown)


jeylau (approved):

jeylau:
Thanks @rahulporuri! Looks great to me!
Just curious, how is that dict output being used on your side afterwards? ☺️

rahulporuri:
Thanks for the quick review @jeylau .

> Just curious, how is that dict output being used on your side afterwards?

I can't go into the full details but like I mentioned in #1895 , we're using DeepLabCut within a larger application. This specific PR allows us to directly get data back from the API instead of reading it back from files.

I wanted to go further and add additional API kwargs to prevent the files from being saved but I didn't want to propose more changes than were necessary.

rahulporuri:
@jeylau do you know if there's any plan to normalize naming in the codebase e.g. renaming functions to use underscores instead of CamelCase naming convention in `auxfun_models` and `auxiliaryfunctions`. If yall are open to the changes, I can open PRs. The differences in convention are slowly getting on my nerves now 😅

jeylau:
Renaming functions to snake case is something we gradually do when refactoring/cleaning code. We're definitely more than happy to get a bit of help here too! 😄

AlexEMG:
Good addition, but I think we should make flag: `returndata=True` or so, so that one has control.

rahulporuri:
> Good addition, but I think we should make flag: `returndata=True` or so, so that one has control.

I'd actually flip it the other way around - add a `save=True` flag which the caller can set to `False` if necessary. But like I mentioned earlier, the APIs already has a lot of arguments so I thought i'll leave this change for a later PR.

AlexEMG:
I think it's best to set the flags so that the functionality is stable with respect to prior versions.

MMathisLab:
hi all, what's the decision on this PR?

MMathisLab:
shall we add the flags cc @AlexEMG @rahulporuri @n-poulsen ? or close?

MMathisLab:
Bump to dev team! ^

n-poulsen:
I think adding the flags is a great idea

MMathisLab:
@n-poulsen can you add the flag so we can merge/close this PR?

jeylau:
@AlexEMG @MMathisLab flag added, good to merge now! Thanks @rahulporuri, sorry for getting back to your PR only now!

rahulporuri:
> @AlexEMG @MMathisLab flag added, good to merge now! Thanks @rahulporuri, sorry for getting back to your PR only now!

I apologize for not making the time to come back to my PR. Thank you very much for continuing to work on it!

Issue #1679, opened 2022-02-09
Colab Freezes at Collecting statsmodels>=0.11   Using cached statsmodels-0.13.2.tar.gz (17.9 MB)


tomwilson41986:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Bug description

Myself and other users are experiencing an issue with Colab this morning.

When installing dependencies, Colab is crashing on the step

Collecting statsmodels>=0.11
Using cached statsmodels-0.13.2.tar.gz (17.9 MB)

We are trying to install:
!pip install deeplabcut
!pip install opencv-python-headless==4.5.1.48
!pip install --upgrade tables
!pip install tensorflow==2.7

Describe the expected behavior
Dependencies usually install within 2-3 mins. Now it is freezing for over 1 hour.

What web browser you are using
Chrome

### Operating System

operating system

### DeepLabCut version

dlc version

### DeepLabCut mode

single animal

### Device type

gpu

### Steps To Reproduce

https://colab.research.google.com/drive/1YbTE3snFNIjQ3l2J1NQ2C4O8ER7tWxf-?usp=sharing

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/DeepLabCut/DeepLabCut/blob/master/CODE_OF_CONDUCT.md)

jeylau:
There is indeed something wrong with their latest release specifically. I got it to work simply by adding `!pip install statsmodels==0.13.1` before the line `!pip install deeplabcut`. Don't hesitate to report the issue on their repo though. ☺️

PR #1910, opened 2022-07-19
Unpin statsmodels

### LINKED ISSUE #1870: Statsmodels 0.13.2?
I noticed that statsmodels 0.13.2 is disallowed during the installation of deeplabcut.

Was this due to installation woes reported in 
- https://github.com/statsmodels/statsmodels/issues/8279
- https://github.com/DeepLabCut/DeepLabCut/issues/1679

or was there a runtime requirement change?

Thank you for your help.

jeylau:
We had pinned statsmodels a while ago, as v0.13.2 caused Colab to freeze when installing DLC (#1679). I found that this is no longer the case, so I'd suggest removing the pin.

Fixes #1870

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,7 @@ pyyaml
 scikit-image>=0.17,<1.0.0
 scikit-learn>=1.0
 scipy>=1.4
-statsmodels>=0.11,<=0.13.1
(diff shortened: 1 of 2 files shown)


hmaarrfk:
Thanks

PR #1913, opened 2022-07-21
Rename functions in `auxiliaryfunctions.py` from CamelCase to underscores

rahulporuri:
In this PR, we went through every function in the `deeplabcut.utils.auxiliaryfunctions` and renamed functions from CamelCase to underscores.

Each function was renamed and the function usage updated in a single commit so this PR can easily be reviewed one commit at a time.

Note that CamelCase aliases are added to the end of the module for backwards compatibility. I'm not sure how many people are using `deeplabcut` as a package and importing functions from inside the package. If backwards compatibility isn't a concern and the maintainers are sure that there is no need to maintain the aliases, I can remove them. If necessary, I can add `DeprecationWarning`s when the aliases are being used, if necessary.

### DIFF
--- a/deeplabcut/generate_training_dataset/multiple_individuals_trainingsetmanipulation.py
+++ b/deeplabcut/generate_training_dataset/multiple_individuals_trainingsetmanipulation.py
@@ -196,7 +196,7 @@ def create_multianimaltraining_dataset(
     scorer = cfg["scorer"]
     project_path = cfg["project_path"]
     # Create path for training sets & store data there
-    trainingsetfolder = auxiliaryfunctions.GetTrainingSetFolder(cfg)
+    trainingsetfolder = auxiliaryfunctions.get_training_set_folder(cfg)
     full_training_path = Path(project_path, trainingsetfolder)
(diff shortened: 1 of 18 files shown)


rahulporuri:
@jeylau @AlexEMG ^ one of the renaming PRs I had in mind

jeylau (approved):

jeylau:
Looks good to me @rahulporuri, thanks a lot!

AlexEMG (approved):

AlexEMG:
Thanks, maybe good to add `DeprecationWarning`s for the CamelCase functions.

rahulporuri:
> Thanks, maybe good to add `DeprecationWarnings` for the CamelCase functions.

Sure. I'll try to add those soon.

MMathisLab:
Is this backwards compatible though? It has to be backwards comp. to the 2.0 code base cc @jeylau

jeylau:
@MMathisLab, yes, @rahulporuri handled it with the aliases added here https://github.com/DeepLabCut/DeepLabCut/blob/86d741ec713801b2436e393e3b336474797e9c23/deeplabcut/utils/auxiliaryfunctions.py#L817-L828

PR #1945, opened 2022-08-12
fliplr augmentation for single animal

biol-jsh:
Added the fliplr augmentation for the single animal augmentation.

### DIFF
--- a/deeplabcut/pose_estimation_tensorflow/datasets/pose_imgaug.py
+++ b/deeplabcut/pose_estimation_tensorflow/datasets/pose_imgaug.py
@@ -22,7 +22,7 @@ import pickle
 import imgaug.augmenters as iaa
 import numpy as np
 import scipy.io as sio
-
+from deeplabcut.pose_estimation_tensorflow.datasets import augmentation
 from deeplabcut.utils.auxfun_videos import imread
 from deeplabcut.utils.conversioncode import robust_split_path
 from .factory import PoseDatasetFactory
@@ -35,6 +35,7 @@ from .utils import DataItem, Batch
 class ImgaugPoseDataset(BasePoseDataset):
     def __init__(self, cfg):
         super(ImgaugPoseDataset, self).__init__(cfg)
+        self._n_kpts = len(cfg["all_joints_names"])
         self.data = self.load_dataset()
         self.batch_size = cfg.get("batch_size", 1)
         self.num_images = len(self.data)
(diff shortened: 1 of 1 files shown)


MMathisLab:
great, could you post some metrics on performance with this aug type in the main PR comment? Looks exciting! #AIResident :)

biol-jsh:
> great, could you post some metrics on performance with this aug type in the main PR comment? Looks exciting! #AIResident :)

Yes, but those metrics won't be available until Monday.

biol-jsh:
Training was interrupted in the weekend but for the first case I've tested (half of the training data, OOD i.e. no refining), the pixel error went down with around 2.7 px (average for 50k and 100k iterations) without pcut and around 1.5 px with pcut 0.6. So it seems to be working. Haven't done any statistics yet but will tomorrow.

jeylau (approved):

jeylau:
Looks good to me, thanks @biol-jsh!

MMathisLab:
@jeylau default will be false, right? but let's benchmark on classic DLC openField dataset, just for ease of mind

jeylau:
@MMathisLab, that's right! It will only be used if a user specifies fliplr and symmetric_pairs in the pose_cfg.

MMathisLab:
we want to merge, right? not closed on purpose @biol-jsh ?

Tetra-quark:
@biol-jsh I just need a clarification: Does this augmentation both flip the image and swap the symmetric label pairs in training? Effectively doubling the training samples and resulting in the left and right being learnt as distinct from each other? (Or conversely does it make left and right indistinguishable?)

biol-jsh:
@Tetra-quark Hi! It is to improve the network's ability to tell left from right. I'm not sure if it doubles the amount of training data, but in essence, yes, it flips both symmetrical pairs and the image. Imagine that you train on several videos of a dog walking from the left to the right. Then you want to analyze a video of a dog going from the right to the left. The network would typically struggle with that and confuse left and right. But by augmenting with fliplr, your network will train on videos (and labels) that are flipped so that the dogs walk from the right to the left as well, and therefore do better at telling left from right. See ["Balancing bats", DeepLabCut Blog, Aug 31, 2022](https://deeplabcut.medium.com/balancing-bats-d3c5bcfb71e1) for an example of how it can help.

Tetra-quark:
@biol-jsh Thanks so much! Really happy to see this feature.

PR #1946, opened 2022-08-12
Fix fliplr augmentation for multi animal

biol-jsh:
The current implementation of fliplr breaks when keypoints are hidden since only labels for visible points are forwarded to the pipeline.

### DIFF
--- a/deeplabcut/pose_estimation_tensorflow/datasets/pose_multianimal_imgaug.py
+++ b/deeplabcut/pose_estimation_tensorflow/datasets/pose_multianimal_imgaug.py
@@ -251,7 +251,6 @@ class MAImgaugPoseDataset(BasePoseDataset):
         batch_images = []
         batch_joints = []
         joint_ids = []
-        inds_visible = []
         data_items = []
         for i in range(self.batch_size):
             data_item = self.data[img_idx[i]]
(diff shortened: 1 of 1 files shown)


deeplabcut/pose_estimation_tensorflow/datasets/pose_multianimal_imgaug.py
```diff
@@ -266,18 +265,17 @@ def get_batch(self):
             if self.has_gt:
                 Joints = data_item.joints
                 kpts = np.zeros((self._n_kpts * self._n_animals, 2))
```
jeylau:
```suggestion
                kpts = np.full((self._n_kpts * self._n_animals, 2), np.nan)
```

MMathisLab:
cc @AlexEMG

jeylau:
Jonas and I experienced a bug that is now fixed with #2037. This PR will be fine to merge afterwards.

MMathisLab:
@n-poulsen actually I can't re-run tests as first it needs conflicts merged; could you take a stab at it? :)

n-poulsen:
As the branch from which the PR was made (patch-2) was deleted from the forked repository, no changes could be pushed to this PR. Fixes to the code were made to the duplicate issue (PR #2043).

This PR should be closed, as merging #2043 will resolve this issue.

PR #1952, opened 2022-08-17
add validation for fps when fps*1000 overflow 65535(max of 16bit)

yueoka:
Thank you for the great dlc programs.

I applied
> deeplabcut.create_labeled_video(config_path, [video_path])

After dlc added points to all frames, dlc cannot save the data as mp4 with following message.
> [mpeg4 @ 000001eb3d342700] timebase 1000/66667 not supported by MPEG 4 standard, the maximum admitted value for the timebase denominator is 65535

To solve the problem, I added 2 rows.
Please confirm it works well.

Best regards,

### DIFF
--- a/deeplabcut/utils/make_labeled_video.py
+++ b/deeplabcut/utils/make_labeled_video.py
@@ -88,8 +88,9 @@ def CreateVideo(
         ny, nx = clip.height(), clip.width()
 
     fps = clip.fps()
-    if fps*1000 > 65535:
-        fps = round(fps)
+    if isinstance(fps, float):
+        if fps*1000 > 65535:
+            fps = round(fps)
     nframes = clip.nframes
     duration = nframes / fps
 
--- a/deeplabcut/utils/make_labeled_video.py
+++ b/deeplabcut/utils/make_labeled_video.py
@@ -88,6 +88,8 @@ def CreateVideo(
         ny, nx = clip.height(), clip.width()
 
     fps = clip.fps()
+    if fps*1000 > 65535:
+        fps = round(fps)
     nframes = clip.nframes
     duration = nframes / fps
 


jeylau:
Hi @yueoka, good catch! I would prefer though if you could add the `round()` there, directly into the `VideoProcessor`:
https://github.com/DeepLabCut/DeepLabCut/blob/22258aa411fa92ab7d470ad28ad11e8a33326526/deeplabcut/utils/video_processor.py#L71

Moreover, reading more [here](https://blender.stackexchange.com/questions/56893/timebase-error-when-encoding), I'm not sure you need to hardcode 1000: rounding fps to the nearest integer seems to be enough.

jeylau:
Why was this merged?

yueoka:
We may have to discuss more before merge.

1. Adding `round()` to the `VideoProcessor` looks good.
2. If simply rounding fps, fps which is float and under 65.535 will also be rounded, though the fps is not needed to be rounded, I thought. 

BTW, adding codec to the `deeplabcut.create_labeled_video` like
```
deeplabcut.create_labeled_video(config_path, [video_path], codec='avc1')
```
can save the mp4 data without rounding fps in my local Macbook environment, though still cannot save in google colab environment.
This might suggest that rounding fps is not needed to solve the mp4 saving problem.

I want you staffs to tell me what to do.
For example, 
* keep merge, and end this PR
* revert and change rounding process to `VideoProcessor` with if sentences
* revert and change rounding process to `VideoProcessor` without if sentences (just apply rounding to fps)
* revert and not merge any commit and end this PR
* etc...

jeylau:
Thanks for reporting back to us! Since users do use Colab though, I think it's better to keep the rounding ☺️

PR #1954, opened 2022-08-18
Fix scorer_3d in triangulate()

### LINKED ISSUE #1941: v2.2.1.1: 3D scorer name is written incorrectly into the resulting 3D.h5 file

### Bug description

Historically, the H5 file that results after running `deeplabcut.triangulation` would write the 3D scorer name in the file. This is the correct behavior. 

Currently, it puts a 2D scorer name into the H5, which is suboptimal -- this does not tell us about the 3d project info, and how would one decide which scorer to pull anyhow? One can have different DLC models for each camera, or use the same, etc. 

Here is an example: top is correct (dlc VERSION 2.0.7) and bottom (version 2.2.1.1):
image (4)

### Operating System

OS indepent (but ubuntu)

### DeepLabCut version

dlc version 2.0.7 vs 2.2.1.1

### DeepLabCut mode

3d

### Device type

N/A

### Steps To Reproduce

Run `deeplabcut.triangulation` in version 2.0.7 and version 2.2.1.1

### Relevant log output

_No response_

### Anything else?

issues around here likely...

https://github.com/DeepLabCut/DeepLabCut/blob/8ddb7abde7ee065289149cd24b6cdd33b3605f07/deeplabcut/pose_estimation_3d/triangulation.py#L233

### Code of Conduct

jeylau:
Fixes #1941

### DIFF
--- a/deeplabcut/pose_estimation_3d/triangulation.py
+++ b/deeplabcut/pose_estimation_3d/triangulation.py
@@ -351,7 +351,6 @@ def triangulate(
             scorer_cam1 = dataFrame_camera1_undistort.columns.get_level_values(0)[0]
             scorer_cam2 = dataFrame_camera2_undistort.columns.get_level_values(0)[0]
 
-            scorer_3d = scorer_cam1
             bodyparts = dataFrame_camera1_undistort.columns.get_level_values(
                 "bodyparts"
             ).unique()


MMathisLab (approved):
nice - so this this keeps the 3d scorer in header

PR #1960, opened 2022-08-24
Update the Default frame extraction algorithm.

JacobBumgarner:
This is an extremely minor fix. The documented default frame extraction method is the ``"uniform"`` distribution algorithm, but the argument default is `"kmeans"`. This minor patch fixes that issue.

### DIFF
--- a/deeplabcut/generate_training_dataset/frame_extraction.py
+++ b/deeplabcut/generate_training_dataset/frame_extraction.py
@@ -64,7 +64,7 @@ def select_cropping_area(config, videos=None):
 def extract_frames(
     config,
     mode="automatic",
-    algo="uniform",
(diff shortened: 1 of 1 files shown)


MMathisLab:
ah, thanks! actually we should update the docs - kmeans is generally much better default to find maximally different frames, which is ideal -- if you want to modify (roll back), and PR doc change, that would be awesome, thanks @JacobBumgarner

AlexEMG:
Agree, the docs not the code should be updated.

JacobBumgarner:
Sounds good. I'll update this PR tonight!

MMathisLab (approved):

AlexEMG (approved):

PR #1972, opened 2022-08-31
Active learning recipe

sfmig:
maybe this could be a blog post instead, let me know :)


MMathisLab:
Hey @sfmig totally agree it is a fantastic blog post: how about editing this page to add a note about "see this blog post about how to test augmentations methods, etc" ? 
https://github.com/DeepLabCut/DeepLabCut/blob/master/docs/recipes/nn.md#using-custom-image-augmentation

and I will port now to Medium 🔥

AlexEMG:
I agree better as a blog post for now, and once relevant parts are in DLC, we can put it as a recipe :)

sfmig:
cool, thanks for the feedback! I added the data augmentation comment :)

I also removed some data augm scripts I had pushed by mistake

MMathisLab:
okay moved to BLOG and linked! https://deeplabcut.medium.com/exploring-active-learning-with-deeplabcut-an-ai-residents-journey-e441bbd5a71c

PR #1991, opened 2022-09-20
Megadetector & DLC batch processing

vchiang001:
Megadetector & DLC batch processing in http://deeplabcut/deeplabcut/utils/


AlexEMG (approved):
Looks great!

PR #2014, opened 2022-10-14
Added inference code for incoming SuperAnimal models

yeshaokai:
Added a new API for inferencing the incoming super model.

Supports arbitrary path loading for the super model snapshot

Supports test time spatial pyramid


MMathisLab:
@yeshaokai  please add:
- [ ] unit test
- [ ] functional test
- [x] docs, which should be nested here: 
Screen Shot 2022-10-19 at 10 35 29 AM

deeplabcut/pose_estimation_tensorflow/predict_supermodel.py
```diff
@@ -0,0 +1,673 @@
...
+    ret["inputs"] = inputs
+    ret["outputs"] = outputs
+    ret["destfolder"] = destfolder
+    ret["save_frames"] = save_frames
+    ret["init_weights"] = init_weights
+
+    return ret
+
+
+def get_multi_scale_frames(frame, scale_list=[]):
```
jeylau: Can `scale_list` really be empty?

deeplabcut/pose_estimation_tensorflow/predict_supermodel.py
```diff
@@ -0,0 +1,673 @@
...
+    cfg,
+    test_cfg,
+    sess,
+    inputs,
+    outputs,
+    cap,
+    nframes,
+    batchsize,
+    invert_color=False,
+    scale_list=[],
```
jeylau, line 189: Same question here: if `scale_list` is empty, `get_multi_scale_frames` returns empty lists of frames and shapes, causing errors further down.

deeplabcut/pose_estimation_tensorflow/predict_supermodel.py
```diff
@@ -0,0 +1,673 @@
...
+    predict_multianimal as predict
+from deeplabcut.pose_estimation_tensorflow.lib import inferenceutils
+from deeplabcut.pose_estimation_tensorflow.lib import trackingutils
+from deeplabcut.utils import auxfun_multianimal
+from deeplabcut.utils import auxiliaryfunctions
+from deeplabcut.utils.auxfun_videos import VideoWriter
+
+
+# instead of having these in a lengthy function, I made this a separate function
+def get_nuances(
```
jeylau, line 24: Can we also reuse this function (which should probably be renamed, it's not very descriptive) elsewhere?

deeplabcut/pose_estimation_tensorflow/predict_supermodel.py
```diff
@@ -0,0 +1,673 @@
...
+
+    ret_pred = {}
+    num_kpts = len(preds[0]["coordinates"][0])
+    # PrediteData is a complicate multi dimensional python list
+    ret_pred["coordinates"] = [[[]] * num_kpts]
+    ret_pred["confidence"] = [[]] * num_kpts
+
+    for scale_id, pred in enumerate(preds):
+        # better handle the case where the pred is empty
+        if pred == []:
```
jeylau:
```suggestion
        if not len(pred):
```

deeplabcut/pose_estimation_tensorflow/predict_supermodel.py
```diff
@@ -0,0 +1,673 @@
...
+        candidates = np.array(ret_pred["coordinates"][0][kpt_id])
+        dist = []
+
+        for i in range(len(candidates)):
+            dist.append(cosine_similarity(candidates[i], mean_vec))
+
+        filter_indices = []
+
+        for idx, ele in enumerate(ret_pred["coordinates"][0][kpt_id]):
+            if dist[idx] < 0.997 or ret_pred["confidence"][kpt_id][idx] < 0.1:
```
jeylau: what are those hardcoded values (dist in particular)?

deeplabcut/pose_estimation_tensorflow/predict_supermodel.py
```diff
@@ -0,0 +1,673 @@
...
+        if _frame is not None:
+            frame = img_as_ubyte(_frame)
+
+            if invert_color:
+                frame = 255 - frame
+
+            old_shape = frame.shape
+            frames, frame_shapes = get_multi_scale_frames(frame, scale_list)
+
+            if multi_scale_batched_frames == None:
```
jeylau:
```suggestion
            if multi_scale_batched_frames is None:
```

deeplabcut/pose_estimation_tensorflow/predict_supermodel.py
```diff
@@ -0,0 +1,673 @@
...
+    Parameters
+    ----------
+    config: string
+        Full path of the config.yaml file as a string.
+
+    videos: list
+        A list of strings containing the full paths to videos for analysis or a path to the directory, where all the videos with same extension are stored.
+
+    scale_list: list
+        A list of int containing the target height of the multi scale test time augmentation. By default it uses the original size. Users are adviced to try a wide range of scale list when the super model does not give reasonable results
```
jeylau:
```suggestion
        A list of int containing the target height of the multi scale test time augmentation. By default it uses the original size. Users are advised to try a wide range of scale list when the super model does not give reasonable results
```

deeplabcut/pose_estimation_tensorflow/predict_supermodel.py
```diff
@@ -0,0 +1,673 @@
...
+        confidence[kpt_id] = confidence_list[max_idx]
+        coordinate[kpt_id] = max_pred
+    return pred
+
+
+def _average_multiple_scale_preds(preds, scale_list):
+
+    ret_pred = {}
+    num_kpts = len(preds[0]["coordinates"][0])
+    # PrediteData is a complicate multi dimensional python list
```
jeylau:
```suggestion
```

deeplabcut/pose_estimation_tensorflow/predict_supermodel.py
jeylau: Besides scale_list, is that function different from AnalyzeVideo? Wouldn't it be preferable to extend the latter function, so we keep dynamic cropping for example?

yeshaokai: I purposely removed dynamic cropping as I wanted to control the resolution of the image the model sees.

deeplabcut/pose_estimation_tensorflow/predict_supermodel.py
yeshaokai: You mean in other part of DLC code? Yes we should.

deeplabcut/pose_estimation_tensorflow/predict_supermodel.py
yeshaokai: dist is cosine similarity between the prediction and the median of the predictions. I will make it a parameter to pass

deeplabcut/pose_estimation_tensorflow/predict_supermodel.py
yeshaokai: Will never be empty given how other part of the code is written

deeplabcut/pose_estimation_tensorflow/predict_supermodel.py
yeshaokai: If scale_list is given empty, we read the resolution of the video and take the shape of the video.

deeplabcut/pose_estimation_tensorflow/predict_supermodel.py
jeylau: then let's remove the default empty list, just to make sure the argument is mandatory

MMathisLab:
I think the name should be "SuperAnimal" not SuperModel fyi: cc @yeshaokai @jeylau @AlexEMG - thoughts?

yeshaokai:
> I think the name should be "SuperAnimal" not SuperModel fyi: cc @yeshaokai @jeylau @AlexEMG - thoughts?

You mean changing the API's name from video_inference_supermodel to video_inference_superanimal? Sure we can do that.

docs/ModelZoo.md
```diff
@@ -0,0 +1,67 @@
...
+```{admonition} Click the button to see API Docs
+:class: dropdown
+```{eval-rst}
+.. include:: ./api/deeplabcut.video_inference_superanimal.rst
+```
+
+### To see the list of available models, check out the [Home page](http://modelzoo.deeplabcut.org/). 
+
+**Coming soon:** The DeepLabCut Project Manager GUI will allow you to use the SuperAnimal Models. You can run the model and do ``active learning" to improve performance on your data. 
+Spefically, we have *new* video adaptation methods to make your tracking extra smooth and robust!
```
jeylau:
```suggestion
Specifically, we have *new* video adaptation methods to make your tracking extra smooth and robust!
```

deeplabcut/pose_estimation_tensorflow/superanimal_configs/superquadruped.yaml
```diff
@@ -0,0 +1,1632 @@
+all_joints:
```
jeylau, line 1: I think we should remove the PAF graphs and set partaffinityfield_predict: false in both configs.

yeshaokai: Yes I did that in the next PR

deeplabcut/pose_estimation_tensorflow/predict_supermodel.py
```diff
@@ -0,0 +1,638 @@
...
+        Evaluate a video's number of frames in a robust manner.
+        This option is slower (as the whole video is read frame-by-frame),
+        but does not rely on metadata, hence its robustness against file corruption.
+
+    allow_growth: bool, default false.
+        For some smaller GPUs the memory issues happen. If true, the memory allocator does not pre-allocate the entire specified
+        GPU memory region, instead starting small and growing as needed. See issue: https://forum.image.sc/t/how-to-stop-running-out-of-vram/30551/2
+
+    save_frames: bool, default false.
+        If video adaptation is needed later, this flag should be set as True
```
jeylau: That's not intuitive to me; instead of writing frames to disk, cannot we use the VideoReader to get video frames during video adaptation?

yeshaokai: Can we do random access with VideoReader for SGD?

jeylau: yes, we could generate random indices and pick frames with reader.set_to_frame()

jeylau: I'll add this later to the video adaptation PR, as just discussed.

deeplabcut/pose_estimation_tensorflow/predict_supermodel.py
```diff
@@ -0,0 +1,602 @@
...
+            ret_pred["confidence"][kpt_id] = np.array(
+                [np.nanmedian(np.array(ret_pred["confidence"][kpt_id]), axis=0)]
+            )
+        else:
+            ret_pred["coordinates"][0][kpt_id] = np.array([[np.nan, np.nan]])
+            ret_pred["confidence"][kpt_id] = np.array([[np.nan]])
+    return ret_pred
+
+
+def video_inference(
```
jeylau: Should we make this one "private"?

yeshaokai: Yes we should. I don't know if we want to make it private by putting _ as prefix or simply not import it under deeplabcut module

jeylau: Both I suppose.

yeshaokai: Usually I don't import those private functions in the __init__.py. We can of course add a _ in the function name

PR #2043, opened 2022-11-10
Fix fliplr augmentation for multi animal 2

biol-jsh:
The current implementation of fliplr breaks when keypoints are hidden since only labels for visible points are forwarded to the pipeline. This is the second attempt at fixing this as an earlier attempt failed due to the error fixed in PR #2037.


MMathisLab:
Hi @n-poulsen the tests are failing, can you look into it? `FAILED tests/test_pose_multianimal_imgaug.py::test_get_targetmaps[3] - TypeError: get_targetmaps_update() missing 1 required positional argument: 'scale'`

jeylau:
I'm still having some tests failing when training with identity; I'll push a fix as soon as I find why.

jeylau:
Ready to merge @MMathisLab @n-poulsen

MMathisLab:
Thanks @jeylau - can we get some # on performance? It won't be a default setting, right? But there are docs for it somewhere? :)

n-poulsen (commented):
Looks good to me!

MMathisLab (approved):

PR #2064, opened 2022-11-24
Adding dogs and cats back to model zoo (from Hugging Face). 

AlexEMG:
Also updated notebook to install from main.


ibenhami:
Hello, 
I started discovering deeplabcut since two months. 
I have tried the zoo model with a cheetah video and it worked well.  
However, today I couldn't find cheetah Zoo model on Google Collab
Could you help me with this ?
Thanks in advance

PR #2065, opened 2022-11-26
do not write when frame is None

jh85:
When creating a labeled video with a bad frame, save_frame() fails. This is just to add a sanity check for the argument frame.


jeylau (approved):

jeylau:
Good catch @jh85, thanks!

PR #2078, opened 2022-12-07
Video adaptation api

yeshaokai:
Added video reader based video adaptation training. Now we don't need images on disk for adaptation training.

Added a new modelzoo folder to host all related apis. 

Note that we need train configs (instead of merely inference config) under superanimal_configs to support video adaptation training.

Occlusion in topview might require gradient masking to prevent undesired learning for occlusion. But empirically setting adaptation iteration as 1000 avoids that.


deeplabcut/modelzoo/apis/spatiotemporal_adapt.py
```diff
@@ -0,0 +1,194 @@
...
+                                       modelfolder = "temp_topview",
+                                       videotype = videotype)
+
+        adapter.before_adapt_inference()
+        adapter.adaptation_training()
+        adapter.after_adapt_inference()
+
+
+        """
+        assert supermodel_name in ["superquadruped", "supertopview"]
```
jeylau:
```suggestion
        if supermodel_name not in ["superquadruped", "supertopview"]:
            raise ValueError("`supermodel_name` should be either 'superquadruped' or 'supertopview'")
```

deeplabcut/modelzoo/apis/spatiotemporal_adapt.py
```diff
@@ -0,0 +1,194 @@
...
+        self.adapt_modelprefix = vname + "_video_adaptation"
+        self.adapt_iterations = adapt_iterations
+        self.modelfolder = modelfolder
+        self.pcutoff = pcutoff
+        self.pseudo_threshold = pseudo_threshold
+
+        if modelfolder != "":
+            os.makedirs(modelfolder, exist_ok=True)
+
+            dlc_root_path = os.sep.join(deeplabcut.__file__.split(os.sep)[:-1])
```
jeylau, line 83: Let's use the dedicated auxiliary function `get_deeplabcut_path` to get the DLC path.

deeplabcut/modelzoo/apis/spatiotemporal_adapt.py
```diff
@@ -0,0 +1,194 @@
...
+
+            name_dict = {
+                "supertopview": "supertopview.yaml",
+                "superquadruped": "superquadruped.yaml",
+            }
```
jeylau: I feel that those should be defined outside, maybe in a modelzoo/config.py or similar, where we'd define all constants relative to modelzoo.

deeplabcut/pose_estimation_tensorflow/core/train_multianimal.py
```diff
@@ -39,15 +39,39 @@ def train(
...
+    # in case there was already a graph
+    tf.compat.v1.reset_default_graph()
+
     start_path = os.getcwd()
-    os.chdir(
-        str(Path(config_yaml).parents[0])
-    )  # switch to folder of config_yaml (for logging)
+    if modelfolder == "":
+        os.chdir(
+            str(Path(train_config_path).parents[0])
```
jeylau, line 53: train_config_path is undefined

deeplabcut/modelzoo/apis/spatiotemporal_adapt.py
```diff
@@ -0,0 +1,194 @@
...
+        _, pseudo_label_path, _, _ = deeplabcut.auxiliaryfunctions.load_analyzed_data(
+            video_root, vname, DLCscorer, False, ""
+        )
+
+        self.train_without_project(pseudo_label_path)
+
+    def after_adapt_inference(self):
+
+        pattern = os.path.join(
+            self.modelfolder, f"*/snapshot-{self.adapt_iterations}.index"
```
jeylau: I think this would fail on Windows; we probably need:
```suggestion
            self.modelfolder, "*", f"snapshot-{self.adapt_iterations}.index"
```

deeplabcut/modelzoo/apis/spatiotemporal_adapt.py
```diff
@@ -0,0 +1,194 @@
...
+        )
+        ref_proj_config_path = ""
+
+        files = glob.glob(pattern)
+
+        assert len(files) > 0
```
jeylau:
```suggestion
        if not len(files):
            raise ValueError("Weights were not found.")
```

jeylau (changes requested):

deeplabcut/modelzoo/apis/spatiotemporal_adapt.py
```diff
@@ -0,0 +1,196 @@
...
+        deeplabcut.create_labeled_video(
+            "",
+            [self.video_path],
+            videotype=self.videotype,
+            filtered=False,
+            init_weights=self.init_weights,
+            draw_skeleton=True,
+            superanimal_name=self.supermodel_name,
+            pcutoff=self.pcutoff,
+        )
```
jeylau: Is that still something we need since we no longer write images to disk?

yeshaokai: The create_labeled_video call under before_adapt is just to showcase what's the difference between using or without using video adaptation. It's not necessary but might be informative.

jeylau: I don't think that should be the default behavior: that'd be very costly on a long video, and I do not believe users would run the function blindly. You'd need to inspect the video first to determine whether adaptation is needed, don't you think?

yeshaokai: I agree.. But "inspecting video" is exactly what create_labeled_video under before_adapt is doing.

yeshaokai: So by default maybe users do not want to call adaptation training and after_adapt_inference. But before_adapt_inference is always the first step.

deeplabcut/pose_estimation_tensorflow/datasets/pose_multianimal_imgaug.py
```diff
@@ -97,6 +121,53 @@ def load_dataset(self):
...
         return data
 
+    @lru_cache(maxsize=None)
+    def read_image_shape_fast(self, path):
+        # Blazing fast and does not load the image into memory
+        with Image.open(path) as img:
+            width, height = img.size
+            return len(img.getbands()), height, width
+
+    def _load_pseudo_data_from_h5(self, cfg, threshold=0.5):
```
jeylau, line 126: `threshold` is not used; where should it go?

yeshaokai: Right let me fix it

yeshaokai: Thanks! Fixed

PR #2091, opened 2022-12-16
prevents replotting of existing plots during model evaluation

ludovicbellier:
Hi team DLC!

Here's my use case and the optimization I propose in this PR: I trained a DLC model on 200k iterations, saving snapshots every 10k iterations; I evaluated all iterations with plotting=True; when I looked at the train/test error curves, I realize my model needed more iterations (best snapshot was at 190k iterations, so it didn't plateaued yet); I restarted training from init_weight snapshot #200k for 500k more iterations; once training finished, I relaunched evaluate_network (still with plotting=True), and even though the 20 first snapshots had already been evaluated and were correctly detected as such ("This net has already been evaluated!" output), the plotting step re-occurred, adding about 1min per snapshot to unnecessarily recreate existing plots.

In this PR, I simply updated evaluate.py around l955 (case when `notanalyzed` is False) to skip the Plotting call if folder already exists (it could be cleaner to check if all frames within folder exist, but to me this is good enough). That way, I just saved myself 20 min. Could also have specified snapshotindex in config.yaml, but it wasn't immediately clear how to set this parameter (besides -1, all or an int).

As it's my first PR in this repo, I would like to take a second and thank you for providing to the community such an amazing toolbox!

Best,
Ludovic

### DIFF
--- a/deeplabcut/pose_estimation_tensorflow/core/evaluate.py
+++ b/deeplabcut/pose_estimation_tensorflow/core/evaluate.py
@@ -275,8 +275,10 @@ def return_evaluate_network_data(
     # Data=pd.read_hdf(os.path.join(cfg["project_path"],str(trainingsetfolder),'CollectedData_' + cfg["scorer"] + '.h5'),'df_with_missing')
 
     # Get list of body parts to evaluate network for
-    comparisonbodyparts = auxiliaryfunctions.intersection_of_body_parts_and_ones_given_by_user(
-        cfg, comparisonbodyparts
+    comparisonbodyparts = (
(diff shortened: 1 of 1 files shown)


ludovicbellier:
Also changed the order of pandas dataframes concatenated when evaluating a model after further training. It was putting the latest snapshots first, which isn't a terrible thing as we could still use the snapshot column as x-axis, but this commit makes the csv more readable.

MMathisLab (approved):

jeylau:
Awesome, thanks @ludovicbellier! 😊

jeylau (approved):

ludovicbellier:
Wonderful, thanks @jeylau and @MMathisLab!

Issue #2094, opened 2022-12-21
efficiently masking the videos


ludovicbellier:
Hi team DLC,

I have a particular use case for which I'm currently developing some code, and would like to get your input about 1) whether you have insights on how to best get this done, and 2) whether you would be interested in getting this as a feature (in such case I would create a PR).

Here's the use case: I have a set of videos in which I want to track one animal, but there is a second animal in part of the FOV, contained within a specific area, who can't move much and for which I'm not interested in getting pose estimates. I've tried to analyze one of these videos with a model trained on single-animal videos (hoping it would generalize like in Mathis et al., 2018, Fig. 4b), but it seemed like there could only be one instance of each body part label per frame (so sometimes the left back paw label was placed onto the animal I want to ignore, thus "stealing" the label from the animal of interest). I bypassed this issue by applying a mask on the video before frame extraction, effectively hiding the FOV area containing the second animal (and in which the animal of interest can't physically go). Hope this makes sense so far!

Now, here's the feature idea: instead of rewriting all videos to get the corresponding masked version, I was thinking we could do the same as cropping, i.e., in the config.yaml file, below each video we could have the option to put "mask: <some params>" to apply a mask on the video in an online way. The mask could either be a path to an image file of the exact same dimensions than the cropped frames, or even (messier but more "online") parameters for drawing shapes in the FOV (e.g., "circle", center_x, center_y, radius). And there would be relevant flags for all functions (extract_frames obviously, but also analyze_videos and create_labeled_videos).

And that's it, basically it's about adding a built-in masking feature alongside the existing cropping feature, so that we don't have to do this step beforehand loosing time and disk space doing so. I don't know if such use cases where masking is needed occur often though, hence this issue to get your insights on it first. Also, if you have suggestions on better ways to handle this use case than using masks, I'm looking forward to reading them!

Thank you,
Ludovic

MMathisLab:
> I've tried to analyze one of these videos with a model trained on single-animal videos (hoping it would generalize like in Mathis et al., 2018, Fig. 4b), but it seemed like there could only be one instance of each body part label per frame (so sometimes the left back paw label was placed onto the animal I want to ignore, thus "stealing" the label from the animal of interest). 

You need to set `num_outputs` > 1 if you just want to get the points across animals, otherwise indeed it only takes the top-one hit.


> instead of rewriting all videos to get the corresponding masked version, I was thinking we could do the same as cropping, i.e., in the config.yaml file

[...]

PR #2096, opened 2022-12-22
Label video when num_outputs > 1 (and overwriting flag)

ludovicbellier:
Hi DLC team!

Based on a reply of @MMathisLab on an [unrelated PR](https://github.com/DeepLabCut/DeepLabCut/issues/2094#issuecomment-1361688793), I went ahead and added `num_outputs: 2` in the `config.yaml` file of a model trained from frames/videos of a single animal, but which I wanted to use to analyze videos containing two animals. Calling `analyze_video` worked fine, and the resulting HDF file contained for each body part two sets of x,y,likelihood (namely, 'x', 'y', 'likelihood', 'x2', 'y2', 'likelihood2'). However, calling `create_labeled_video` resulted in an indexing error ("IndexError: index N is out of bounds for axis 0 with size N", N being the number of body parts), and it came from the fact that while I had 2N sets of estimates (two per body part), there were only N colors (one per body part; further details in the code).

This PR fixes the bug described above by taking into account situations where num_outputs>1 in single animal models, in a parsimonious approach (i.e., by replacing existing lines of code so that output using num_outputs=1 doesn't change at all, and output using num_outputs>1 is now valid, as opposed to adding a new situation).

Also, while doing this debugging I constantly had to manually delete the created movie, so I implemented an overwriting flag in `create_labeled_video` that defaults at False. Turning it to True forces overwriting.

(besides, here's the bottom line of my attempt at estimating poses for the two animals: while the video did show two labels per body part after debugging, on many frames my animal of interest had two labels on it for the same body part, so it looks like a dead end for this strategy in my use case, and I'll have to rely on masking!)

Please don't hesitate if you need further information!
Best,
Ludovic

### DIFF
--- a/deeplabcut/pose_estimation_tensorflow/core/evaluate.py
+++ b/deeplabcut/pose_estimation_tensorflow/core/evaluate.py
@@ -275,8 +275,10 @@ def return_evaluate_network_data(
     # Data=pd.read_hdf(os.path.join(cfg["project_path"],str(trainingsetfolder),'CollectedData_' + cfg["scorer"] + '.h5'),'df_with_missing')
 
     # Get list of body parts to evaluate network for
-    comparisonbodyparts = auxiliaryfunctions.intersection_of_body_parts_and_ones_given_by_user(
-        cfg, comparisonbodyparts
+    comparisonbodyparts = (
+        auxiliaryfunctions.intersection_of_body_parts_and_ones_given_by_user(
+            cfg, comparisonbodyparts
+        )
     )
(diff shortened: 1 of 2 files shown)


MMathisLab:
this looks great @ludovicbellier - we will review and merge after the break :D

MMathisLab (approved):

jeylau:
That's great @ludovicbellier, thank you very much!

jeylau (approved):

PR #2105, opened 2022-12-29
improve model zoo install and docs

MMathisLab:


AlexEMG (approved):
Thanks, lovely additions.

PR #2126, opened 2023-01-20
GUI fixes

KonradDanielewski:
I'm currently testing the GUI and trying to fix reported issues. One is more design improvement, i.e. being able to change the working dir also via textbox, not only by clicking the `Location` text (I think maybe it should stand out somehow - it doesn't indicate in any way that it's a button unless you hover over it)

Another thing I noticed was that label removal wasn't working in the tracklet toolbox, so also fixed that. Seems to work now.

Some other reported things that I didn't know if should be changed: 

- `Labelling GUI` button disabled by defualt in Extract Outlier Frames tab - what if you extracted before and are going back to refining?
- Manual extraction of frames still points to old GUI when called via `deeplabcut.extract_frames()`

### DIFF
--- a/deeplabcut/generate_training_dataset/frame_extraction.py
+++ b/deeplabcut/generate_training_dataset/frame_extraction.py
@@ -263,11 +263,9 @@ def extract_frames(
     from deeplabcut.utils import auxiliaryfunctions
 
     if mode == "manual":
-        wd = Path(config).resolve().parents[0]
-        os.chdir(str(wd))
-        from deeplabcut.gui import frame_extraction_toolbox
-
-        frame_extraction_toolbox.show(config, slider_width)
+        from deeplabcut.gui.widgets import launch_napari
+        _ = launch_napari()
+        return
 
     elif mode == "automatic":
         config_file = Path(config).resolve()
--- a/deeplabcut/gui/tabs/analyze_videos.py
+++ b/deeplabcut/gui/tabs/analyze_videos.py
@@ -258,8 +258,6 @@ class AnalyzeVideos(DefaultTab):
 
         videos = list(self.files)
         save_as_csv = self.save_as_csv.checkState() == Qt.Checked
-        save_as_nwb = self.save_as_nwb.checkState() == Qt.Checked
-        filter_data = self.filter_predictions.checkState() == Qt.Checked
         videotype = self.video_selection_widget.videotype_widget.currentText()
 
         if self.root.is_multianimal:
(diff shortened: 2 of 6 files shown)


jeylau:
Thanks @KonradDanielewski! I agree that the Labeling GUI button should be enabled.
Do you want to address the `deeplabcut.extract_trames()` in this PR?

KonradDanielewski:
Sure, I'll enable the button and pointing to napari in frame extraction today

KonradDanielewski:
@jeylau So another reported issue is `QObject::startTimer: Timers can only be used with threads started with QThread` when using the player feature of the tracklet toolbox. If I understand correctly, the BackgroundPlayer class has to be made into a QObject and TrackletVisualizer class converted to work with a QThread that will run that QObject - I can also be entirely incorrect since it's my first time working with Qt

jeylau:
Awesome, thanks @KonradDanielewski! Does the Timer issue prevent the tracklet GUI from running though?

KonradDanielewski:
The tracklet GUI is running and the player even starts and stops albeit with some delay, but when you try to skip forward or backward with `alt+->`/`alt+<-` it does nothing and if tried a couple of times throws `Assertion fctx->async_lock failed at libavcodec/pthread_frame.c:167` which ultimately crashes the whole GUI (tracklet and general one).

KonradDanielewski:
I just noticed that the way I implemented the Location directory change is easily broken when the path isn't pasted first (every input into the Location textbox updates from all the textboxes). I'll see if I can fix this but if not it should be discarded from the pull request.

KonradDanielewski:
Partially adresses #2161 - still haven't found out why the data from refinement isn't extracted and saved

KonradDanielewski:
About the `Assertion fctx->async_lock failed at libavcodec/pthread_frame.c:167` it's also happening in the older GUI when you try to move forward during the playback. I read a little more to try and understand this and I might have an idea. 

The `alt + left/right` works in an intended way, just producing a lot of warnings. If I understand correctly, all the things happening in the tracklet_toolbox are run multithreaded and somehow if this skip by one frame by just pressing an arrow key without `alt` happens during the video playback, the video read goes out of sync. So I guess it's more of an unexpected user input not being handled rather than anything else.

jeylau (approved):

KonradDanielewski:
I removed the textbox based location change since it wasn't working properly

jeylau:
@KonradDanielewski, I'm looking at https://github.com/DeepLabCut/DeepLabCut/issues/2152 separately. Is your PR ready otherwise?

KonradDanielewski:
The rest works as intended - though one issue is not resolved completely - the refinement not extracting data and creating `machinelabels` file - but I'm not sure I can fix this one - to my eyes it all looks correct but for some reason doesn't work.

KonradDanielewski:
OK, honest mistake. It doesn't export data cause there is no export in the `refine_tracklets.py`, only merge which doesn't do anything without the export.

KonradDanielewski:
Fixed the refined tracklets data export and mutliindex creation. Also added handling for two pandas deprecation warnings

KonradDanielewski:
Also fixed the `analyze_videos` tab since it wasn't working properly. I guess now the most obvious GUI bugs are fixed

jeylau:
Sweet! I fixed the tracklet refinement GUI in #2191, so should be good to merge!

PR #2144, opened 2023-02-03
Fix creating video after using cropping during analysis

lilfer:
Analyzing a video with some `cropping` coordinates, and then creating the labeled video with `fastmode=True` and `displaycropped=False` (default values) will result in the the labels being misplaced in the video.
Should be fixed now.


lilfer:
fixed by #2147

PR #2153, opened 2023-02-10
added argument 'test_only' to 'evaluate_network' and 'evaluate_multianimal_full'

MitchFuchs:
By specifying `test_only=True`, model is evaluated only on its test indexes. In case of a training fraction of 0.95, evaluation time is decreased by a factor of 20.

Tested `evaluate_network` (incl. `evaluate_multianimal_full`) with multi-animal project and following default arguments: `plotting=False, show_errors=True, comparisonbodyparts="all", rescale=False`. 

Evaluation results on test (e. g. 'Test error (px)' and 'Test error with p-cutoff') with `test_only=True` and `test_only=False` are identical in the _CombinedEvaluation-result.csv_.

### DIFF
--- a/deeplabcut/pose_estimation_tensorflow/core/evaluate.py
+++ b/deeplabcut/pose_estimation_tensorflow/core/evaluate.py
@@ -501,6 +501,7 @@ def evaluate_network(
     gputouse=None,
     rescale=False,
     modelprefix="",
+    test_only=False
 ):
     """Evaluates the network.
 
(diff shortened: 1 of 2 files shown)


MMathisLab:
bump for @yeshaokai @jeylau review :)

yeshaokai:
@MitchFuchs 
Thanks for the PR. One concern I have for this is that, since now you change the shape of the Data by doing Data.iloc[testIndices], this results in the saved Data to be test only. Then the train test indices that are passed to the assembler will be invalid.

For example, if you look at here https://github.com/DeepLabCut/DeepLabCut/blob/main/deeplabcut/pose_estimation_tensorflow/lib/crossvalutils.py#L429

For the assembler to work correctly, the indices and corresponding Data need to match.

PR #2158, opened 2023-02-18
add new way to visualize low-likelihood pose estimates

ludovicbellier:
Hi team DLC!

During a recent project (I work at Inscopix, by the way =)) a colleague was wondering where low-likelihood labels would be placed if we could see them, which made me realize I couldn't find any existing way to visualize that: by default labels with likelihood <= pcutoff are not plotted when creating a labelled video; we can sure drop pcutoff to 0 but then all frames are labeled and there is no way to differentiate between above and below pcutoff labels.

Here is a simple PR keeping the default mode ("hide"), so it won't go in the way of existing scripts, and adding a "darken" mode when label colors are darkened when body part likelihood gets at or below pcutoff.

Also added a way to override pcutoff and dotsize values from config.yaml directly in the function call, so that these parameters could be changed more conveniently and, more importantly, by users without writing access to config.yaml but still wanting to use the related trained model.

Looking forward to discussing these features with you, and again, thank you for putting out there such a great package!
Have a beautiful weekend,
Ludovic

### DIFF
--- a/deeplabcut/utils/make_labeled_video.py
+++ b/deeplabcut/utils/make_labeled_video.py
@@ -68,6 +68,7 @@ def get_segment_indices(bodyparts2connect, all_bpts):
 def CreateVideo(
     clip,
     Dataframe,
+    pcutoff_mode,
     pcutoff,
     dotsize,
     colormap,
@@ -171,11 +172,13 @@ def CreateVideo(
                         image[rr, cc] = color_for_skeleton
 
             for ind, num_bp, num_ind in bpts2color:
-                if df_likelihood[ind, index] > pcutoff:
+                if df_likelihood[ind, index] > pcutoff or pcutoff_mode == "darken":
                     if color_by == "bodypart":
                         color = colors[num_bp]
                     else:
                         color = colors[num_ind]
+                    if df_likelihood[ind, index] <= pcutoff:
+                        color = np.ceil(color * .2).astype(int)
                     if trailpoints > 0:
                         for k in range(1, min(trailpoints, index + 1)):
                             rr, cc = disk(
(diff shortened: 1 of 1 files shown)


MMathisLab:
HI @ludovicbellier super idea! thanks very much -- I can see where this becomes handy in several ways; @jeylau want to check for maDLC compatibility? otherwise, lgtm also cc @AlexEMG - fine for you? We also need to consider backwards compatibility with config file

jeylau:
This is cool! An alternative would have been to display low-confidence keypoints as unfilled circles (as is done with when viewing outlier keypoints during refinement I think), but I do not see an advantage for doing so.
It is backward compatible too as no new variable is read from the config file; only the mode is passed in in memory in the dictionary 😊

jeylau (approved):

jeylau (commented):
Remove duplicate argument

deeplabcut/utils/make_labeled_video.py
```diff
@@ -377,8 +383,10 @@ def create_labeled_video(
...
     init_weights="",
     track_method="",
+    pcutoff_mode="hide",
+    pcutoff=None,
+    dotsize=None,
     superanimal_name="",
-    pcutoff=0.6,
     skeleton=[],
     skeleton_color="white",
     dotsize=8,
```
ludovicbellier: This one should be removed too (causes the test failures)

ludovicbellier:
Thanks @MMathisLab and @jeylau for your replies, and glad you like it! Agreed on alternative ways to plot this: I just went for color change as this was the simplest to implement (as labels are plotted using sklearn.draw.disk), but using other markers or mapping dotsize to likelihood (e.g., half the size when below likelihood threshold) would also be efficient at visually conveying this information!

AlexEMG (approved):

jeylau (approved):

n-poulsen:
An alternative implementation of this same feature (alpha represents likelihood of keypoints) was added with pull request #2319.

PR #2179, opened 2023-03-20
pip error after updating anaconda env

SaeedF36:
I try to install DLC on M1. I get a pip error after deactivating and activating.
**export:12: not valid in this context: XML_CATALOG_FILES+**

I use miniconda3.


jeylau:
@SaeedF36, I guess you intended to open an issue rather than a pull request? Can I close this one?

PR #2198, opened 2023-04-06
Additions to modelzoo APIs

yeshaokai:
Spatial pyramid is applied by default in video adaptation now, to act as filtering and to reduce quantization error

If it's topview, gradient masking is also applied in video adaptation

Some other stuff for developers' use

### DIFF
--- a/deeplabcut/pose_estimation_tensorflow/core/train_multianimal.py
+++ b/deeplabcut/pose_estimation_tensorflow/core/train_multianimal.py
@@ -44,8 +44,8 @@ def train(
     modelfolder="",
     traintime_resize=False,
     video_path="",
-    topview = False,
-    trim_ends = None #trim the both ends of the video for video adaptation
+    superanimal=None,
+    trim_ends = None  # trim the both ends of the video for video adaptation
 ):
     # in case there was already a graph
     tf.compat.v1.reset_default_graph()
@@ -65,8 +65,10 @@ def train(
     cfg["pseudo_threshold"] = pseudo_threshold
     cfg["video_path"] = video_path
     cfg["traintime_resize"] = traintime_resize
-    cfg["topview"] = topview
-    
+
+    if superanimal is not None:
+        cfg["superanimal"] = superanimal
+
     if pseudo_labels != "":
         cfg["pseudo_label"] = pseudo_labels
 
--- a/deeplabcut/pose_estimation_tensorflow/core/train_multianimal.py
+++ b/deeplabcut/pose_estimation_tensorflow/core/train_multianimal.py
@@ -40,10 +40,12 @@ def train(
     allow_growth=True,
     pseudo_labels="",
     init_weights="",
-    pseudo_threshold=0,
+    pseudo_threshold=0.1,
     modelfolder="",
     traintime_resize=False,
     video_path="",
+    topview = False,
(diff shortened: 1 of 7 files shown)


MMathisLab (approved):

MMathisLab:
@yeshaokai please fix tests errors

jeylau:
@yeshaokai, looks good to me. Could we fix #2212 in this one though before merging?

jeylau (approved):

yeshaokai:
I fixed the empty prediction following the user's advice. After the tests pass we should be good.

deeplabcut/pose_estimation_tensorflow/core/train_multianimal.py
```diff
@@ -63,7 +65,8 @@ def train(
     cfg["pseudo_threshold"] = pseudo_threshold
     cfg["video_path"] = video_path
     cfg["traintime_resize"] = traintime_resize
-
+    cfg["topview"] = topview
```
MMathisLab: can you explain why this is needed?

MMathisLab (changes requested):

deeplabcut/pose_estimation_tensorflow/core/train_multianimal.py
yeshaokai: To add gradient masking in video adaptation. If the user choose high threshold for video adaptation, then many keypoints are dropped thus masking is needed. But "masking for video adaptation" is only applicable for topview. During inference time,  masking for  side-view quadruped models will make the model confuse occluded keypoints and visible but filtered keypoints .

MMathisLab: lets minimally change to ["superanimal"] then topview is okay; then we are forward thinking for new models. Agree @yeshaokai then we can merge actually

deeplabcut/pose_estimation_tensorflow/predict_supermodel.py
```diff
@@ -20,6 +20,10 @@ def video_inference_superanimal(
     video_adapt=False,
     plot_trajectories=True,
     pcutoff=0.1,
+    init_weights = '',
```
n-poulsen: This argument isn't used in the method, it should be removed

examples/testscript_superanimal_adaptation.py
```diff
@@ -41,15 +41,3 @@
...
-
-    superanimal_name = "superanimal_quadruped"
-    deeplabcut.video_inference_superanimal(
-        [video],
-        superanimal_name,
-        videotype=".mp4",
-        video_adapt=True,
-        scale_list=scale_list,
-        pcutoff=0.3,
-    )
```
n-poulsen: As we have `video_adapt=True`, adaptation training will be run for the (default) number of iterations, which is 1000. I think this is a bit slow for a test script,  and we should only run the adaptation for ~50 iterations.

```python
        pcutoff=0.3,
        adapt_iterations=50,
    )
```

n-poulsen (dismissed):

MMathisLab (approved):

yeshaokai:
@n-poulsen Thanks for the edits! They look fine to me

MMathisLab (approved):

AlexEMG (approved):

PR #2217, opened 2023-04-27
codespell: config, action and a few typos fixed

yarikoptic:
To help keeping DeepLabCut typos free. Individual commits might have more reasoning etc.

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,3 +1,5 @@
 [codespell]
 skip = .git,*.pdf,*.svg
-ignore-words-list = sie
+# MOT,SIE - legit acronyms
+# tThe - for \tThe. codespell is not good detecting those yet
+ignore-words-list = mot,sie,tthe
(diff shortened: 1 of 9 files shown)


jeylau:
Hey @yarikoptic, this is cool, thanks a lot! Just one question though: if a typo is detected when the GitHub action runs, does the workflow entirely stop or does codespell auto-fix it? If it doesn't, I'm thinking that maybe we should have a deeplabcut[dev] that installs e.g. codespell, black, pytest.

yarikoptic:
Only fails/stops so author could either fix or ignore it. Many projects use pre-commit to automate use of those tools on developer side. Black etc could be added like we have eg in https://github.com/dandi/dandi-cli/blob/master/.pre-commit-config.yaml . But that should be separate/complimentary to this PR IMHO.

AlexEMG (approved):

MMathisLab:
can merge after checks pass, thanks for the contribution @yarikoptic

PR #2220, opened 2023-04-28
Auto publish to pypi with tagged commits

jeylau:
Commits tagged with "v*" trigger a release, and [setuptools_scm](https://github.com/pypa/setuptools_scm) automatically writes the version of DLC to _version.py (this is also the workflow of https://github.com/DeepLabCut/napari-deeplabcut), CEBRA, and CellSeg3D :D!

Example code to trigger a release:

 create a release, tagged with a v* (i.e. v3.0.0rc6!)


MMathisLab (approved):
- @AlexEMG you will need to set up your secret key in org (I can show you), otherwise looks good to me!

.github/workflows/python-package.yml
```diff
@@ -64,3 +66,29 @@ jobs:
...
+          python -m pip install --upgrade pip
+          pip install -U setuptools setuptools_scm wheel twine build
+      - name: Build and publish
+        env:
+          TWINE_USERNAME: __token__
+          TWINE_PASSWORD: ${{ secrets.TWINE_API_KEY }}
+        run: |
+          git tag
+          python -m build .
+          twine upload dist/*
```
stes: you could consider pushing to test pypi either on PR merge or when the release PR is tagged --- that way the full workflow ran once before you finalize? but optional ofc, heres an example: https://github.com/AdaptiveMotorControlLab/CEBRA/blob/0378db0b2431d0c50a1e9b80aa1b865869586851/.github/workflows/release-pypi.yml#L34-L43

.github/workflows/python-package.yml
```diff
@@ -64,3 +66,29 @@ jobs:
...
+
+  deploy:
+    # this will run when you have tagged a commit, starting with "v*"
+    # and requires that you have put your twine API key in your
+    # github secrets (see readme for details)
+    needs: [build]
+    runs-on: ubuntu-latest
+    if: contains(github.ref, 'tags')
+    steps:
+      - uses: actions/checkout@v3
```
stes: would use latest version, which is v4 https://github.com/actions/checkout

.github/workflows/python-package.yml
```diff
@@ -64,3 +66,29 @@ jobs:
...
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -U setuptools setuptools_scm wheel twine build
+      - name: Build and publish
+        env:
+          TWINE_USERNAME: __token__
+          TWINE_PASSWORD: ${{ secrets.TWINE_API_KEY }}
+        run: |
+          git tag
```
stes, line 99: maybe optional, but u could check that the tag actually matches the output of `deeplabcut.__version__` to enforce consistency

pyproject.toml
```diff
@@ -0,0 +1,7 @@
+[build-system]
+requires = ["setuptools", "wheel", "setuptools_scm"]
+build-backend = "setuptools.build_meta"
+
+
+[tool.setuptools_scm]
+write_to = "deeplabcut/_version.py"
```
stes, line 7: what is this for, version written during `python -m build`?

.github/workflows/python-package.yml
```diff
@@ -64,3 +66,29 @@ jobs:
...
+    # github secrets (see readme for details)
+    needs: [build]
+    runs-on: ubuntu-latest
+    if: contains(github.ref, 'tags')
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.x"
```
stes: maybe better to fix a python version for the build?

stes (commented):
some small comments/qs

jeylau:
Thanks @stes for the comments, I'll answer/fix asap.

.github/workflows/python-package.yml
stes: I think not needed given how you set the version!

PR #2221, opened 2023-04-30
Fix typo in refine_tracklets.py

xun468:
Noticed self.root.viz.export_to_training_data() on line 255 was throwing an attribute error when refining tracklets. Merging refined tracklets into training data now seems to work.


jeylau:
Good catch @xun468, thank you very much!

jeylau (approved):

Issue #2205, opened 2023-04-14
deeplabcut.create_labeled_video_3d()  ValueError: cannot reshape array of size 2826 into shape (161,newaxis,3)


wangziying1004:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Bug description

Hello, I'm wangziying.
I have a problem with  deeplabcut.create_labeled_video_3d() function.
I tried:
```
deeplabcut.create_labeled_video_3d(config_path, ['/content/drive/MyDrive/testing3d-DLC-2023-04-14-3d/videofoldertest'], start=0, end=157,videotype = '.mp4')
but I got this:
Analyzing all the videos in the directory
[['/content/drive/MyDrive/testing3d-DLC-2023-04-14-3d/videofoldertest/handtest1_DLC_3D.h5', '/content/drive/MyDrive/testing3d-DLC-2023-04-14-3d/videofoldertest/handtest1-camera-1.mp4', '/content/drive/MyDrive/testing3d-DLC-2023-04-14-3d/videofoldertest/handtest1-camera-2.mp4']]
Creating 3D video from handtest1-camera-1.mp4 and handtest1-camera-2.mp4 using handtest1_DLC_3D.h5
Looking for filtered predictions...
Found the following filtered data:  /content/drive/MyDrive/testing3d-DLC-2023-04-14-3d/videofoldertest/*handtest1-camera-1DLC_resnet50_handtest1Apr14shuffle1_8000*filtered.h5 /content/drive/MyDrive/testing3d-DLC-2023-04-14-3d/videofoldertest/*handtest1-camera-2DLC_resnet50_handtest1.1Apr14shuffle1_10500*filtered.h5
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
[<ipython-input-8-9a76379fb8de>](https://localhost:8080/#) in <cell line: 1>()
----> 1 deeplabcut.create_labeled_video_3d(config_path, ['/content/drive/MyDrive/testing3d-DLC-2023-04-14-3d/videofoldertest'], start=0, end=157,videotype = '.mp4')

[/usr/local/lib/python3.9/dist-packages/deeplabcut/pose_estimation_3d/plotting3D.py](https://localhost:8080/#) in create_labeled_video_3d(config, path, videofolder, start, end, trailpoints, videotype, view, xlim, ylim, zlim, draw_skeleton, color_by, figsize, fps, dpi)
[...]

PR #2226, opened 2023-05-03
Changing except Error type to match no files found

nattse:
If there are no filtered files, `pd.read_hdf(glob.glob('/some/path/*filtered.h5')[0])` will raise an IndexError as a result of indexing the empty list returned by glob, which will precede/prevent read_hdf from raising the expected FileNotFoundError. I cannot think of any way that glob will return a list of length > 0 but the contents lead to a nonexistent .h5 file

### DIFF
--- a/deeplabcut/pose_estimation_3d/plotting3D.py
+++ b/deeplabcut/pose_estimation_3d/plotting3D.py
@@ -269,10 +269,10 @@ def create_labeled_video_3d(
 
             # Format data
             mask2d = df_cam1.columns.get_level_values("bodyparts").isin(bodyparts2plot)
-            xy1 = df_cam1.loc[:, mask2d].to_numpy().reshape((len(df_cam1), -1, 3))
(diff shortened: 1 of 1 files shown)


nattse:
Second commit addresses #2205. 

If videos from camera-1 and camera-2 differ in length, the resulting .h5 tracking files will also differ in length. Original code reads both camera .h5 files and attempts to reshape them according to camera-1 array size, resulting in a ValueError when camera-2 data is not of the same size. However, any difference in video length is already addressed in triangulation.py, which creates a 3d.h5 file that contains as many frames as the shortest video. Therefore, we can slice both 2D tracking arrays according to the 3d.h5 data size and then reshape, knowing that their sizes will match.

Example of my own run-in with this issue below; camera-1 video had 3575 frames and camera-2 video had 3576 frames.
valueerror

jeylau:
Really great catch @nattse, thank you so much for troubleshooting this one!

jeylau (approved):

PR #2243, opened 2023-05-13
Use QFrame.Shape.StyledPanel instead of QFrame.StyledPanel

hmaarrfk:
I think this helps with qtpy which doesn't provide a few of the "conveniences" of PySide6.

I understand if you might not care, but I do think this helps provide a bit of compatibility with a few other environments.

Thank you for considering this

### DIFF
--- a/deeplabcut/gui/tabs/create_project.py
+++ b/deeplabcut/gui/tabs/create_project.py
@@ -43,7 +43,7 @@ class ProjectCreator(QtWidgets.QDialog):
 
     def lay_out_user_frame(self):
         user_frame = QtWidgets.QFrame(self)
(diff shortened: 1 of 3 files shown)


hmaarrfk:
I'm not sure if this might be the cause but on conda-forge we have newer versons of pyside6 beyond 6.3

hmaarrfk:
in an other change, i'm suggesting to ensure that setLineWidth corresponds to integers:

https://doc.qt.io/qt-6/qframe.html#lineWidth-prop

i would have made a distinct PR but it would have created too many merge conflicts.

MMathisLab (approved):

PR #2244, opened 2023-05-13
Explicitely use QRegularExpression instead of implicit string conversion

hmaarrfk:
Two things here:

- The use of the r-string, or raw string helps ensures that the `\d` is preserved.
- I think that qtpy or conda-forge's PySide6 will throw an error if the regular expression isn't a Qt Regular Expression

Thank you for considering small improvements to compatibility.

Best,

Mark

xref: https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals

### DIFF
--- a/deeplabcut/gui/tabs/modelzoo.py
+++ b/deeplabcut/gui/tabs/modelzoo.py
@@ -10,7 +10,7 @@
 #
 import deeplabcut
 from PySide6 import QtWidgets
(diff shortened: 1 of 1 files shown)


hmaarrfk:
I'm not sure if this might be the cause but on conda-forge we have newer versons of pyside6 beyond 6.3

n-poulsen (approved):

PR #2246, opened 2023-05-13
add outputs for multi-animal  model export

jefromson:
Hi!

I found that these outputs were required to successfully export a multi-animal tracking model. Please let me know if this works and/or if I can help with anything else here.

Thanks!


MMathisLab:
thanks very much! Let's add a unit test for this @n-poulsen to check for backwards/forwards compat, otherwise looks good

n-poulsen:
I need more information before I can approve this PR. As reflected by the test I added, the current version of the code calls `setup_GPUpose_prediction` if `TFGPUinference ==  True`, and otherwise calls `setup_pose_prediction`.

For multi-animal projects, this PR calls `setup_pose_prediction` no matter the value of `TFGPUinference`. More information is needed about the keys required for multi-animal projects for GPU and CPU pose prediction.

hmaarrfk:
Interesting findings. I was always puzzled by the dedicated code paths for the GPU. I suspected they were relics from old tensorflow code that had significant behavior changes between CPU/GPU.

In my experience, i load the same "model" on the CPU and GPU, but it is really for a limited subset of the functionality you have all developed.

PR #2273, opened 2023-06-13
stitch_tracklets crashes when not enough tracks are found

n-poulsen:
When there aren't enough tracks found by the stitcher, the current implementation fails. By running

```python
deeplabcut.analyze_videos(
    "project/config.yaml", ["/videos/"], videotype='.avi', auto_track=True, n_tracks=10,
)
```

The following error occurs as `remaining_nodes` is an empty list
```python
deeplabcut/refine_training_dataset/stitch.py:690: UserWarning: No optimal solution found. Employing black magic...
  File "deeplabcut/refine_training_dataset/stitch.py", line 687, in stitch
    _, self.flow = nx.capacity_scaling(self.G)
  File "/lib/python3.9/site-packages/networkx/algorithms/flow/capacityscaling.py", line 394, in capacity_scaling
    raise nx.NetworkXUnfeasible("No flow satisfying all demands.")
networkx.exception.NetworkXUnfeasible: No flow satisfying all demands.

During handling of the above exception, another exception occurred:

  File "deeplabcut/refine_training_dataset/stitch.py", line 755, in stitch
    self.build_graph(list(remaining_nodes), max_gap=np.inf)
  File "deeplabcut/refine_training_dataset/stitch.py", line 644, in build_graph
    nodes_in, nodes_out = zip(
ValueError: not enough values to unpack (expected 2, got 0)

During handling of the above exception, another exception occurred:

  File "<stdin>", line 1, in <module>
  File "deeplabcut/pose_estimation_tensorflow/predict_videos.py", line 651, in analyze_videos
    stitch_tracklets(
  File "deeplabcut/refine_training_dataset/stitch.py", line 1201, in stitch_tracklets
    stitcher.stitch()
  File "deeplabcut/refine_training_dataset/stitch.py", line 766, in stitch
    raise ValueError(
ValueError: Could not reconstruct 10 tracks from the tracklets given.
```

This PR simply returns once there are no remaining nodes, even if the number of tracks found is smaller than the desired number of tracks (as if there are no remaining nodes, no new tracks can be found). To keep the shape of the output dataframe identical to when tracks are found for all individuals, NaN-filled columns are added for the missing tracks. A user warning is printed so the user is informed that not all tracks were found:

```python
deeplabcut/refine_training_dataset/stitch.py:763: UserWarning: Only 4 tracks could be reconstructed.
```

### DIFF
--- a/deeplabcut/refine_training_dataset/stitch.py
+++ b/deeplabcut/refine_training_dataset/stitch.py
@@ -723,40 +723,41 @@ class TrackletStitcher:
                 for node in self.G
                 if node not in ("source", "sink")
             )
-            if (
-                incomplete_tracks == 1
-            ):  # All remaining nodes must belong to the same track
-                # Verify whether there are overlapping tracklets
-                for t1, t2 in combinations(remaining_nodes, 2):
-                    if t1 in t2:
-                        # Pick the segment that minimizes "smoothness", computed here
-                        # with the coefficient of variation of the differences.
-                        if t1 in remaining_nodes:
-                            remaining_nodes.remove(t1)
-                        if t2 in remaining_nodes:
-                            remaining_nodes.remove(t2)
-                        track = sum(remaining_nodes)
-                        hyp1 = track + t1
-                        hyp2 = track + t2
-                        dx1 = np.diff(hyp1.centroid, axis=0)
-                        cv1 = dx1.std() / np.abs(dx1).mean()
-                        dx2 = np.diff(hyp2.centroid, axis=0)
-                        cv2 = dx2.std() / np.abs(dx2).mean()
-                        if cv1 < cv2:
-                            remaining_nodes.add(t1)
-                            self.residuals.append(t2)
-                        else:
-                            remaining_nodes.add(t2)
-                            self.residuals.append(t1)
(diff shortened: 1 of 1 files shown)


deeplabcut/refine_training_dataset/stitch.py
```diff
@@ -723,40 +723,41 @@ def stitch(self, add_back_residuals=True):
...
-                paths.append(list(remaining_nodes))
-            elif incomplete_tracks > 1:
-                # Rebuild a full graph from the remaining nodes without
-                # temporal constraint on what tracklets can be stitched together.
-                self.build_graph(list(remaining_nodes), max_gap=np.inf)
-                self.G.nodes["source"]["demand"] = -incomplete_tracks
-                self.G.nodes["sink"]["demand"] = incomplete_tracks
-                _, self.flow = nx.capacity_scaling(self.G)
-                paths += self.reconstruct_paths()
+            if len(remaining_nodes) > 0:
```
jeylau, line 726: Great! By the way, I checked, and we could also have:
```
if not remaining_nodes:
    return
```
as the `finally` clause is executed before the function exits, no matter what.
But that's a detail 😃

jeylau (approved):
Awesome, thanks!

Issue #2282, opened 2023-06-24
SpatiotemporalAdaptation kwargs


rahim-hashim:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Bug description

When running SpatiotemporalAdaptation `adapter.adaptation_training` as defined in the documentation, there is an error looking for the adapt_iterations kwargs, even though it is assigned to the adapter object in `adapter.before_adapt_inference()` by default as 1000. 

```
video_path = video_path_list[0]
superanimal_name = 'superanimal_quadruped'

from deeplabcut.modelzoo.api import SpatiotemporalAdaptation
adapter = SpatiotemporalAdaptation(
    video_path,
    superanimal_name,
    modelfolder = 'temp',
    videotype = videotype,
)

adapter.before_adapt_inference()
adapter.adaptation_training()        # <–––––- error
adapter.after_adapt_inference()
```

You can confirm by running `adapter.adapt_iterations` after `adapter.before_adapt_inference()` and seeing the value. It is fixed easily by assigning a new kwarg `adapt_iterations = 1000` and feeding it into the adapter_training:

```
video_path = video_path_list[0]
superanimal_name = 'superanimal_quadruped'
adapt_iterations = 1000

from deeplabcut.modelzoo.api import SpatiotemporalAdaptation
adapter = SpatiotemporalAdaptation(
    video_path,
    superanimal_name,
    modelfolder = 'temp',
    videotype = videotype,
)

adapter.before_adapt_inference()
adapter.adaptation_training(adapt_iterations=adapt_iterations)  # <–––––- error looking for kwargs
adapter.after_adapt_inference()
```
even though this step should not be needed, or else included in the specifications. Small issue, and simple solution, but worth fixing for future use.

### Operating System

MacOS Ventura

### DeepLabCut version

DLC 2.3.5

### DeepLabCut mode

single animal

### Device type

M1

### Relevant log output

```shell
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[24], line 14
      6 adapter = SpatiotemporalAdaptation(
      7     video_path,
      8     superanimal_name,
      9     modelfolder = 'temp',
     10     videotype = videotype,
     11 )
     13 adapter.before_adapt_inference()
---> 14 adapter.adaptation_training()
     15 adapter.after_adapt_inference()

[...]

PR #2291, opened 2023-06-27
Bug fix spatiotemporal adaptation kwargs

n-poulsen:
This PR fixes the bug documented in issue #2282. In `adaptation_training`, an `adapt_iterations` key was read from the `kwargs` without providing a default value. This behavior changes to use the default `adapt_iterations` attribute for the class.

Additionally, `train_without_project` now uses the `adapt_iterations ` of the class as a default, instead of overwriting it with 1000.

Black was also run on the file, leading to style changes.


MMathisLab (approved):

yeshaokai (approved):
Looks good to me

PR #2317, opened 2023-07-19
Avoid assembling if already done and overwrite == False

KonradDanielewski:
This QoL significantly reduces time needed for testing different inference parameters by reading assemblies from an existing file (if already run once). Works for both identity True and False.

### DIFF
--- a/deeplabcut/pose_estimation_tensorflow/lib/inferenceutils.py
+++ b/deeplabcut/pose_estimation_tensorflow/lib/inferenceutils.py
@@ -813,9 +813,8 @@ class Assembler:
     def from_pickle(self, pickle_path):
         with open(pickle_path, "rb") as file:
             data = pickle.load(file)
-        
-        self.assemblies = {key: value for key, value in data.items() if key != "single"}
-        self.unique = data["single"] if "single" in data.keys() else dict()
+        self.unique = data.pop('single', {})
+        self.assemblies = data
         
     @staticmethod
     def parse_metadata(data):
--- a/deeplabcut/pose_estimation_tensorflow/lib/inferenceutils.py
+++ b/deeplabcut/pose_estimation_tensorflow/lib/inferenceutils.py
@@ -813,7 +813,7 @@ class Assembler:
     def from_pickle(self, pickle_path):
         with open(pickle_path, "rb") as file:
             data = pickle.load(file)
-        self.unique = data.pop('single', {})
+        self.unique = data.pop("single", {})
         self.assemblies = data
         
     @staticmethod
--- a/deeplabcut/pose_estimation_tensorflow/lib/inferenceutils.py
+++ b/deeplabcut/pose_estimation_tensorflow/lib/inferenceutils.py
@@ -809,7 +809,14 @@ class Assembler:
                         if unique is not None:
                             self.unique[i] = unique
                         pbar.update()
-
+    
+    def from_pickle(self, pickle_path):
+        with open(pickle_path, "rb") as file:
+            data = pickle.load(file)
+        
+        self.assemblies = {key: value for key, value in data.items() if key != "single"}
+        self.unique = data["single"] if "single" in data.keys() else dict()
+        
     @staticmethod
     def parse_metadata(data):
         params = dict()
--- a/deeplabcut/pose_estimation_tensorflow/predict_videos.py
+++ b/deeplabcut/pose_estimation_tensorflow/predict_videos.py
@@ -1821,7 +1821,8 @@ def convert_detections2tracklets(
                     ass.assemble()
                     ass.to_pickle(dataname.split(".h5")[0] + "_assemblies.pickle")
                 else:
-                    print(f"Loading assemblies from {dataname.split(".h5")[0] + "_assemblies.pickle"}")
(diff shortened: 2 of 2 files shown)


deeplabcut/pose_estimation_tensorflow/predict_videos.py
```diff
@@ -1798,26 +1799,31 @@ def convert_detections2tracklets(
...
+                            str(trainingsetfolder),
+                            "CollectedData_" + cfg["scorer"] + ".h5",
+                        )
+                        ass.calibrate(train_data_file)
+                    ass.assemble()
+                    ass.to_pickle(dataname.split(".h5")[0] + "_assemblies.pickle")
+                else:
+                    ass_filename = dataname.split(".h5")[0] + "_assemblies.pickle"
+                    print(f"Loading assemblies from {ass_filename}")
+                    ass = pd.read_pickle(dataname.split(".h5")[0] + "_assemblies.pickle")
```
jeylau: @KonradDanielewski, I'd be in favor of adding a `classmethod` called `Assembler.from_pickle`, as another constructor: what would you think about that? We'd encapsulate the code you wrote into that function, so we manipulate early on an `Assembler` object only, simplifying the code further below. I also imagine that creating an Assembler quickly may be helpful for benchmarking in the future.

jeylau (changes requested):

deeplabcut/pose_estimation_tensorflow/predict_videos.py
KonradDanielewski: Sure, that makes more sense. I'll try to get it done today.

KonradDanielewski: Like we discussed, because the `full.pickle` is loaded anyway, we just initialize the Assembler and then fill the assemblies and unique from the file instead of creating a `classsmethod`. This avoids the necessity of creating fake metadata etc.

deeplabcut/pose_estimation_tensorflow/lib/inferenceutils.py
```diff
@@ -809,7 +809,14 @@ def wrapped(i):
                         if unique is not None:
                             self.unique[i] = unique
                         pbar.update()
-
+    
+    def from_pickle(self, pickle_path):
+        with open(pickle_path, "rb") as file:
+            data = pickle.load(file)
+        
+        self.assemblies = {key: value for key, value in data.items() if key != "single"}
```
jeylau:
```python
self.unique = data.pop('single', {})
self.assemblies = data
```

KonradDanielewski: Done. A small annoyance is that it hangs a little when loading the `assemblies.pickle`, but I don't think this can be avoided

jeylau: No worries! That's nothing in comparison to the time that would otherwise be spent forming the assemblies.

deeplabcut/pose_estimation_tensorflow/lib/inferenceutils.py
```diff
@@ -809,7 +809,13 @@ def wrapped(i):
                         if unique is not None:
                             self.unique[i] = unique
                         pbar.update()
-
+    
+    def from_pickle(self, pickle_path):
+        with open(pickle_path, "rb") as file:
+            data = pickle.load(file)
+        self.unique = data.pop('single', {})
```
jeylau:
```suggestion
        self.unique = data.pop("single", {})
```

deeplabcut/pose_estimation_tensorflow/predict_videos.py
```diff
@@ -1808,16 +1808,21 @@ def convert_detections2tracklets(
...
+                    if calibrate:
+                        trainingsetfolder = auxiliaryfunctions.get_training_set_folder(cfg)
+                        train_data_file = os.path.join(
+                            cfg["project_path"],
+                            str(trainingsetfolder),
+                            "CollectedData_" + cfg["scorer"] + ".h5",
+                        )
+                        ass.calibrate(train_data_file)
+                    ass.assemble()
+                    ass.to_pickle(dataname.split(".h5")[0] + "_assemblies.pickle")
```
jeylau: Since the file path is repeated three times, I'd define above before the `if`, and use it where needed.

jeylau: You could move `ass_filename` up actually, and reuse this variable.

jeylau (dismissed):
Just left a minor comment about `ass_filename`; all good otherwise!

deeplabcut/pose_estimation_tensorflow/predict_videos.py
MMathisLab: `ass_` is not a great variable name, please change throughout ;)

MMathisLab (changes requested):

KonradDanielewski:
Renamed to `assembly_builder`, seems appropriate and avoids confusion with the class name. Also added `assemblies_filename` as a variable before the if statement

MMathisLab (approved):

jeylau:
Looking good, thanks @KonradDanielewski!

MMathisLab (approved):

PR #2319, opened 2023-07-19
Plot keypoint confidence as alpha value

n-poulsen:
Adds the ability to use the confidence scores for keypoints for the alpha values for disks when creating labeled videos.

### DIFF
--- a/deeplabcut/utils/make_labeled_video.py
+++ b/deeplabcut/utils/make_labeled_video.py
@@ -1047,7 +1047,7 @@ def create_video_with_all_detections(
         If False, all keypoints will be plot with alpha=1. Otherwise, this can be
         defined as a function f: [0, 1] -> [0, 1] such that the alpha value for a
         keypoint will be set as a function of its score: alpha = f(score). The default
-        function used when True is f(x) = max(0, (x - pcutoff)/(1 - pcutoff)).
+        function used when True is f(x) = x.
     """
     from deeplabcut.pose_estimation_tensorflow.lib.inferenceutils import Assembler
     import re
--- a/deeplabcut/utils/make_labeled_video.py
+++ b/deeplabcut/utils/make_labeled_video.py
@@ -618,8 +618,12 @@ def create_labeled_video(
     )
     if draw_skeleton:
         bodyparts2connect = cfg["skeleton"]
-        if displayedbodyparts!="all":
-            bodyparts2connect = [pair for pair in bodyparts2connect if all(element in displayedbodyparts for element in pair)]
+        if displayedbodyparts != "all":
+            bodyparts2connect = [
+                pair
+                for pair in bodyparts2connect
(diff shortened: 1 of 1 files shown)


jeylau (approved):

n-poulsen:
The code was updated so that `confidence_to_alpha` can now be a function. As the docstring describes:
> confidence_to_alpha: Union[bool, Callable[[float], float], default=False
> If False, all keypoints will be plot with alpha=1. Otherwise, this can be defined as a function f: [0, 1] -> [0, 1] such that the alpha value for a keypoint will be set as a function of its score: alpha = f(score). The default function used when True is f(x) = max(0, (x - pcutoff)/(1 - pcutoff)).

As an example of the resulting video obtained with the plotting function, here are the results on a toy project of running `create_video_with_all_detections`  with two different values:

```python
deeplabcut.create_video_with_all_detections(..., confidence_to_alpha=False)
```
https://github.com/DeepLabCut/DeepLabCut/assets/45132115/092e74f2-7f23-4fb6-b6b7-191722848508




```python
pcutoff = 0.1
def confidence_to_alpha(c):
    return np.clip((c - pcutoff) / (1 - pcutoff), 0, 1)

deeplabcut.create_video_with_all_detections(..., confidence_to_alpha=confidence_to_alpha)
```
https://github.com/DeepLabCut/DeepLabCut/assets/45132115/2cd75576-9b5d-4548-a9f6-13dc695e5137

In the second video, the low scores for the red keypoint in the top right corner is very clear, while it cannot be inferred from the first video with `confidence_to_alpha=False`.

jeylau:
Looks good @n-poulsen! Low confidence points though will barely be visible: should the edge color remain, and only the face color be made more/less transparent?

MMathisLab:
I think it's good if low conf. are barely/not visible; pcutoff of 0 then wouldn't be shown, and if we can pass a range that is even better

n-poulsen:
Just two things to note:
- keypoints with `score < cfg["pcutoff"]` will not be displayed even if their alpha value is > 0 in the `confidence_to_alpha` function (so if you define your own function, you should adapt your pcutoff to see all keypoints you want to see)
- this plotting method was only implemented for the "fastmode" (both for `create_labeled_videos` and `create_video_with_all_detections`)

MMathisLab (approved):

MMathisLab:
I think this needs revisited. It is still clumsy. 

Right now a user must:

```
pcutoff = 0.1
import numpy as np
def confidence_to_alpha(c):
    return np.clip((c - pcutoff) / (1 - pcutoff), 0, 1)
```

then run: 
```
deeplabcut.create_labeled_video(config_path,demovideo , shuffle=1,videotype="mp4", fastmode=True,pcutoff=pcutoff,confidence_to_alpha=confidence_to_alpha, overwrite=True)
```

Why can we not make this so people can simply pass `confidence_to_alpha=[0.1,1]` ?

PR #2339, opened 2023-08-01
Rename test_modelzoo.py

jeylau:
Note that the test of the download of all individual model zoo models is skipped by default (but can be run locally occasionally).

- [ ] mark download test with attribute


MMathisLab:
why don't we make it run tag-based? ie. like this: https://github.com/AdaptiveMotorControlLab/CEBRA/blob/a21ba0ec17a8c7cc67fc24187f119a52c4132cd0/tests/test_datasets.py#L142 in https://github.com/AdaptiveMotorControlLab/CEBRA/pull/52

jeylau:
Yes, good idea! I'd go for `@parametrize.mark.slow` then, which can easily be run from the CLI. Let me move this to WIP.

MMathisLab:
maybe a small bump here so we can have this in before SA PRs are merged

jeylau:
I'd close the PR and delete the test_modelzoo file, since all model downloads are now systematically tested at `dlclibrary` https://github.com/DeepLabCut/DLClibrary/blob/main/tests/test_modeldownload.py.

PR #2346, opened 2023-08-07
bug fix: assemblies filename print

n-poulsen:
Fixes the name of a variable printed to the console when the assemblies filename exists. Broken in #2317.


jeylau (approved):
Good catch!

PR #2360, opened 2023-08-24
Remove unused parameter `num_shuffles` from `check_for_weights` function. 

Tetra-quark:
I noticed that the `check_for_weights` function takes and return the `num_shuffles` parameter without using it. I searched the repository, verified that it would be safe to remove and made all the alterations necessary.

(I would like to contribute to the DLC repository but up till now was not super comfortable with the workflow. I thought this could be a good test for a first contribution, although small. I understand fully if the pull request is denied for reasons I have not considered.)

### DIFF
--- a/deeplabcut/create_project/modelzoo.py
+++ b/deeplabcut/create_project/modelzoo.py
@@ -273,7 +273,7 @@ def create_pretrained_project(
         auxiliaryfunctions.edit_config(cfg, dict_)
 
         # downloading base encoder / not required unless on re-trains (but when a training set is created this happens anyway)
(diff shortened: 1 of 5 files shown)


jeylau:
Hi @Tetra-quark, good catch! Looks all good to me, thank you very much!

jeylau:
Just made a minor fix directly in the PR to get the tests to pass 😊

Tetra-quark:
Thanks for the fix! 
I imagine the remaining test failures are not related to the changes in this pull request right?

jeylau (approved):

jeylau:
> Thanks for the fix!
> I imagine the remaining test failures are not related to the changes in this pull request right?

You're correct, these can be ignored.

PR #2374, opened 2023-09-11
Fix `interp` kwarg in `imresize` function.

Tetra-quark:
I noticed the `imresize` function had usages that referred to the kwarg `interp` but the actual definition of the function had a different name `interpolationmethod`. 

I then noticed that the args were provided as strings but the underlying [OpenCV method actually takes an enum](https://docs.opencv.org/3.4/da/d54/group__imgproc__transform.html#ga5bb5a1fea74ea38e1a5445ca803ff121).

I applied the necessary corrections although I have not tested the scripts myself. I was interested in visualizing the augmentations without actively training but I wrote my own script for that.


MMathisLab:
@Tetra-quark thanks! also if of use, check out the notebooks that go with our Primer, https://github.com/DeepLabCut/Primer-MotionCapture, as they plot the augmentations

n-poulsen (approved):
Looks good to me! Thanks for the help!

PR #2396, opened 2023-09-25
Address underlying compatibility issue with scipy 1.11.0

hmaarrfk:
See warning in scipy 1.9's documentation.
See warning in https://docs.scipy.org/doc/scipy-1.9.0/reference/generated/scipy.stats.mode.html

The SPEC0 suggests that scipy 1.8 be dropped in 2024, but I think it is OK to drop it a little early.

If you don't want to drop it, we could add a version check for the keyword argument, but that seems a little much.

The keyword argument `keepdims` was added in scipy 1.9

xref: https://github.com/DeepLabCut/DeepLabCut/pull/2290

### DIFF
--- a/deeplabcut/refine_training_dataset/stitch.py
+++ b/deeplabcut/refine_training_dataset/stitch.py
@@ -135,7 +135,12 @@ class Tracklet:
     def identity(self):
         """Return the average predicted identity of all Tracklet detections."""
         try:
-            return mode(self.data[..., 3], axis=None, nan_policy="omit")[0][0]
+            return mode(
+                self.data[..., 3],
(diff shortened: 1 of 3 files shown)


MMathisLab:
NOTE: this would drop 3.7 support, but I think it's fine?

Screen Shot 2023-10-02 at 9 17 05 PM

hmaarrfk:
SPEC0 (not that you are forced to support it) would even have you drop Python 3.8 and 3.9 today ;)

https://scientific-python.org/specs/spec-0000/

jeylau (approved):
Great, thanks @hmaarrfk!

PR #2401, opened 2023-10-03
Update python-package.yml

MMathisLab:
Keeping up with the times. Here, I propose to drop 3.7 and 3.8 testing, see also PR #2290 vs. #2396 where this is needed to keep up with our core dependency, scipy!

- minor spelling errors fixed.


jeylau (approved):
Looking good!

jeylau:
In addition, we can update the conda environment files to use 3.9.

PR #2405, opened 2023-10-05
Fix bug dataset and metadataset fixes DeepLabCut/DeepLabCut#2390

### LINKED ISSUE #2390: Hidden Bug: Training uses `dataset` path defined in pose_cfg but Evaluation ignores it and uses path returned by auxiliary function.

### Bug description

There are two ways to get the filenames for a model's dataset (.mat) and metadataset (.pickle) and mixing them can lead to incorrect evaluation results which can easily go undetected:
- The first way is to access the `dataset` and `metadataset` filepaths through the `pose_cfg.yaml` loaded as a dict eg. `cfg["dataset"]`. The only usages of this in the repository are to access the `dataset` filepath in the augmentation modules (`/datasets` directory). The metadataset (.pickle) file path is in fact never accessed through the `pose_cfg`.
- The second way is to use the auxiliary function `get_data_and_metadata_filenames()` which combines various variables to create the path from the provided args: `trainingsetfolder`, `trainFraction`, `shuffle`, `cfg` (`config.yaml`). This method is the one used the most often, mainly during the dataset creation but also notably during evaluation. 

This means that if one wants to edit the `pose_cfg.yaml` of a model to use a specific training dataset other than the default one for that shuffle, it won't behave as intended. The augmentation modules (`/datasets` folder) use `cfg["dataset"]` to obtain the path to the .mat file and thus the training is done using the dataset defined in the configuration. However, the evaluation (`evaluate.py`) accesses this file path using `get_data_and_metadata_filenames()` and therefore uses the original, default dataset regardless of what `dataset` is specified in either the `train/pose_cfg.yaml` or `test/pose_cfg.yaml`. 

For context, I am trying to use the `modelprefix` parameter to give a clear naming scheme to the groups of models I would like to train. In this workflow the training datasets I want to use don't need to correspond to any specific 'shuffle'. 

### Operating System

MacOS 12.3.1
(& Colab)

### DeepLabCut version

dlc version 2.3.5

### DeepLabCut mode

single animal

### Device type

CPU

### Steps To Reproduce

I can provide a script to demonstrate that that evaluation ignores the pose_cfg if required but it doesn't explicitly show that the training _does_ use the parameter from the pose_cfg. 

### Relevant log output

_No response_

### Anything else?

_No response_

### Code of Conduct

Tetra-quark:
Pull request to fix issue #2390 

TODOs:
- [x]  Fix getting pose_cfg paths in nested loop over shuffle and trainingfractions. Needs to be able to handle `trainingsetindex='all'`. Fix needs to be applied to `evaluate.py` and `multianimal_evaluate.py`. 
- [x] Look for other usages of `get_data_and_metadata_filenames` in other areas of the project and evaluate whether to replace with these with simple dict access from pose_cfg. (For file creation the method should be used but I think afterwards ideally it should be accessed from pose_cfg). 
- [x] Rename variable `dlc_cfg` to `test_pose_cfg` to be more specific and mirror the naming of `train_pose_cfg`
- [x] Handle exception if `train/pose_cfg.yaml` is not found similarly to how load test config is in a try/catch block.
- [ ] Update visualizemaps.py


AlexEMG:
Thanks!

Tetra-quark:
I have looked for other areas of the code that use `get_data_and_metadata_filenames` they are either related to dataset/metadataset file creation (no need to change these) or are test scripts where it matters less how these filenames are accessed.

However, `visualizemaps.py` should be updated given it's structural similiarity to `evaluate.py` and `evaluate_multianimal.py`. But before doing that I was hoping to get some feedback on the code review I opened above as it is relevant to all three files. 

I also wanted to note explicitly that my changes now mean that the `metadataset` parameter in the `train/pose_cfg.yaml` is being accessed by modules that previously only accessed the `test/pose_cfg.yaml`. This is because while there is a `dataset` parameter in the test/pose_cfg.yaml as far as I can see from the testing scripts there is no `metadataset` parameter.

deeplabcut/pose_estimation_tensorflow/core/evaluate.py
```diff
@@ -139,13 +135,17 @@ def calculatepafdistancebounds(
...
                             )
                         else:
                             distances = None
 
                         if distances is not None:
                             if onlytrain:
+                                train_pose_cfg = load_config(str(path_train_config))
+                                _, trainIndices, _, _ = auxiliaryfunctions.load_metadata(
+                                    Path(cfg["project_path"]) / train_pose_cfg['metadataset']
+                                )
```
jeylau: I'd take this out of the loop to avoid loading the metadata over and over again.

deeplabcut/pose_estimation_tensorflow/core/evaluate.py
```diff
@@ -297,23 +294,30 @@ def return_evaluate_network_data(
...
     except FileNotFoundError:
         raise FileNotFoundError(
             "It seems the model for shuffle %s and trainFraction %s does not exist."
             % (shuffle, trainFraction)
         )
 
+    train_pose_cfg = load_config(str(path_train_config))
+    # Load meta data
+    data, trainIndices, testIndices, _ = auxiliaryfunctions.load_metadata(
+        Path(cfg["project_path"]) / train_pose_cfg["metadataset"],
```
jeylau, line 317: I haven't tested it but is passing a `Path` instead of a `str` to `load_metadata()` fine?

jeylau (changes requested):
Thanks @Tetra-quark! I'd only ask you to load the train image indices outside the loop, and that looks great!

jeylau (approved):

jeylau:
Great, thanks! I think this is good to merge, and if some code refactoring is needed in evaluate/visualizemaps, that can be done in a separate PR @MMathisLab @Tetra-quark.

deeplabcut/pose_estimation_tensorflow/core/evaluate.py
```diff
@@ -18,6 +18,8 @@
 import pandas as pd
 from tqdm import tqdm
 
+import deeplabcut
```
n-poulsen: I'm not a fan of importing all of `deeplabcut` in the `core/evaluate` code... would it perhaps be better to just import the module/method that we need?

I believe that here it's just for the `return_train_network_path` method, which could be imported as 
```python
from deeplabcut.pose_estimation_tensorflow.training import return_train_network_path
```

deeplabcut/pose_estimation_tensorflow/core/evaluate_multianimal.py
```diff
@@ -176,50 +177,47 @@ def evaluate_multianimal_full(
...
-                trainingsetfolder, trainFraction, shuffle, cfg
-            )
-            modelfolder = os.path.join(
-                cfg["project_path"],
-                str(
-                    auxiliaryfunctions.get_model_folder(
-                        trainFraction, shuffle, cfg, modelprefix=modelprefix
-                    )
-                ),
+            modelfolder_rel_path = auxiliaryfunctions.get_model_folder(trainFraction, shuffle, cfg, modelprefix=modelprefix)
```
n-poulsen: `black` should be run on the code, as this line (and a few others) is too long

n-poulsen (commented):
Some minor stye comments, otherwise looks good to me

PR #2406, opened 2023-10-05
Napari Flowchart

Timokleia:
I changed the _config file and the flowchart syntax and I expect that now the chart looks well in the book.
Jessy would be great if you try it out.


jeylau (approved):
Looks good, thanks Tim!

PR #2409, opened 2023-10-09
Fix link typos and conversion

AlexEMG:


jeylau (approved):

deeplabcut/create_project/demo_data.py
```diff
@@ -60,7 +60,7 @@ def transform_data(config):
     else:
         print("This is not an official demo dataset.")
 
-    if "WILL BE AUTOMATICALLY UPDATED BY DEMO CODE" in cfg["video_sets"].keys():
+    if "WILL BE AUTOMATICALLY UPDATED BY DEMO CODE" in dict(cfg["video_sets"]).keys():
         cfg["video_sets"][str(video_file)] = cfg["video_sets"].pop(
```
n-poulsen: Does this line need to be cast to a `dict` as well, or can we still `pop` a key from the object that it was?

Not sure this is needed if the above code is tested and works:
```suggestion
        cfg["video_sets"] = dict(cfg["video_sets"])
        cfg["video_sets"][str(video_file)] = cfg["video_sets"].pop(
```

PR #2431, opened 2023-10-31
Improvements to the model zoo's GUI tab

jeylau:
- Move the execution of the video adaptation code to a separate thread
- Add infinite progress bar in the GUI status bar
- Add extra widgets to grant users more control

Screenshot 2023-10-31 at 6 44 44 PM

### DIFF
--- a/deeplabcut/gui/tabs/modelzoo.py
+++ b/deeplabcut/gui/tabs/modelzoo.py
@@ -106,6 +106,18 @@ class ModelZoo(DefaultTab):
         self.run_button.clicked.connect(self.run_video_adaptation)
         self.main_layout.addWidget(self.run_button, alignment=Qt.AlignRight)
 
+        self.help_button = QtWidgets.QPushButton("Help")
+        self.help_button.clicked.connect(self.show_help_dialog)
+        self.main_layout.addWidget(self.help_button, alignment=Qt.AlignLeft)
+
+    def show_help_dialog(self):
+        dialog = QtWidgets.QDialog(self)
+        layout = QtWidgets.QVBoxLayout()
+        label = QtWidgets.QLabel(deeplabcut.video_inference_superanimal.__doc__, self)
+        layout.addWidget(label)
+        dialog.setLayout(layout)
+        dialog.exec_()
+
     def _handle_validation_change(self, state):
         if state == RegExpValidator.Invalid:
             color = "red"
--- a/deeplabcut/gui/tabs/modelzoo.py
+++ b/deeplabcut/gui/tabs/modelzoo.py
@@ -65,11 +65,38 @@ class ModelZoo(DefaultTab):
         validator.validationChanged.connect(self._handle_validation_change)
         self.scales_line.setValidator(validator)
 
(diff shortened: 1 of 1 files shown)


AlexEMG:
Looks great, @yeshaokai what do you think?

yeshaokai:
Discussed with Jessy and it looks good to me

MMathisLab:
Is there a help button? Any guidance to help the user to know what "threshold" is? Why they should or should not use adaptation? 

What form should scale be? This should be smartly set and not exposed - it's also irrelevant for quadruped model with top down...

MMathisLab (changes requested):
The users need more guidance. We really need to bring back the help button!

Scale list should be hidden or a long paragraph on what size our model is trained on, and dynamically updated based on the selection of a model.

jeylau:
I don't think we can hide scale list, as it requires user's input (as it's dependent on video/animal size, but not model type); if left unspecified, it internally defaults to video height +/- 50). If specified by the user though, note that we explicitly indicate that these should be a comma-separated list of integers, and we have an additional input validator that colors the input field frame (red, yellow, green) whether the input format is wrong, incomplete, or valid.

MMathisLab:
Sure, but without docs it's not intuitive what it even is - so sure integers is clear, but we can do this smarter. Measure the video pixel size, scale it to our mean superanimal training dataset size... drop the box 💪

jeylau:
- tooltip display on hovering over the question mark icon
Screenshot 2023-11-03 at 4 42 30 PM

- scrollable help dialog
Screenshot 2023-11-03 at 4 42 46 PM

AlexEMG (approved):

n-poulsen:
The code looks good to me, but I get an issue when running with video adaptation (200 iterations):

```bash
/miniconda/envs/upa/lib/python3.9/site-packages/tensorflow/python/keras/engine/base_layer_v1.py:1694: UserWarning: `layer.apply` is deprecated and will be removed in a future version. Please use `layer.__call__` method instead.
  warnings.warn('`layer.apply` is deprecated and '
Exception in thread Thread-3:
  File "/miniconda/envs/upa/lib/python3.9/threading.py", line 980, in _bootstrap_inner
    self.run()
  File "/miniconda/envs/upa/lib/python3.9/threading.py", line 917, in run
    self._target(*self._args, **self._kwargs)
  File "/DeepLabCut/deeplabcut/pose_estimation_tensorflow/core/train.py", line 84, in load_and_enqueue
    batch_np = dataset.next_batch()
  File "/DeepLabCut/deeplabcut/pose_estimation_tensorflow/datasets/pose_multianimal_imgaug.py", line 538, in next_batch
    temp.append(array[mask[inds]])
IndexError: invalid index to scalar variable.
```

I tested with the current code on `main` and do not get this issue. Does this happen to anyone else?

jeylau:
@n-poulsen, I'm looking at it.

jeylau:
This PR one is ready to merge; I pushed a fix in a separate branch as it's unrelated to the GUI.

PR #2436, opened 2023-11-07
Fix video adaptation

jeylau:
I fixed the extra argument passed to `deeplabcut.train()` and the issue noted by Niels here https://github.com/DeepLabCut/DeepLabCut/pull/2431#issuecomment-1794686008


n-poulsen (approved):
Looks good to me! Thanks for the quick fix!

PR #2443, opened 2023-11-10
Fixes for testscript_3d.py

jesse-gilmer:
Hello DLC Team,

During testing of a fresh install of DLC on Ubuntu 20.04 I noticed that examples/testscript_3d.py was not completing due to various errors.

Here is a summary of the changes made:
1. > Clarification of the base directory in the code. I chose to use the 0th parent directory on line 37 and removed the reference to "examples" from line 42, but you could alternately add " config = glob.glob(os.path.join(basepath,"examples","TEST*", "config.yaml"))[-1] " as a replacement for line 104. Otherwise, line 104 will not be successful, as basepath points to /deeplabcut/, not /deeplabcut/examples/ where the config path is located.

2. > url = "http://www.vision.caltech.edu/bouguetj/calib_doc/htmls/stereo_example.zip" and its associated lines will not complete because that url does not exist any longer. The file exists (I presume this is the same file) at https://data.caltech.edu/records/jx9cx-fdh55 in calib_doc.zip at /htmls/stereo_example.zip but obviously this is not easy to access via url. I added the file to the examples folder and accessed it with "file_name = os.path.join(basepath,"stereo_example.zip")". I'm sure you will need to make some other correction that suits your project.

3. > Line 189 needs an assigned videotype to complete successfully, e.g. "deeplabcut.create_labeled_video_3d(path_config_file, [video_dir], start=5, end=10, videotype=".avi")"

I don't know git well enough to comprehend exactly what your pull request etiquette guide requires, so I hope this is sufficient. Always happy to clarify anything that is unclear.

I am tagging you for a review request per your guidelines, @MMathisLab.

All the best,
Jesse


examples/testscript_3d.py
```diff
@@ -138,11 +137,8 @@
     project_name = path_config_file.split(os.sep)[-2]
 
     os.chdir(os.path.join(project_name, "calibration_images"))
-    # Downloading the calibration images
-    url = "http://www.vision.caltech.edu/bouguetj/calib_doc/htmls/stereo_example.zip"
-    file_name = "stereo_example.zip"
-    with urllib.request.urlopen(url) as response, open(file_name, "wb") as out_file:
-        shutil.copyfileobj(response, out_file)
+
+    file_name = os.path.join(basepath,"stereo_example.zip")
```
MMathisLab, line 141: hi @jesse-gilmer dropping the download link on purpose?

examples/testscript_3d.py
```diff
@@ -186,7 +182,7 @@
     deeplabcut.triangulate(path_config_file, video_dir, save_as_csv=True)
 
     print("CREATING LABELED VIDEO 3-D")
-    deeplabcut.create_labeled_video_3d(path_config_file, [video_dir], start=5, end=10)
+    deeplabcut.create_labeled_video_3d(path_config_file, [video_dir], start=5, end=10, videotype=".avi")
```
MMathisLab, line 185: approve

examples/testscript_3d.py
jesse-gilmer: @MMathisLab I think it would be better to have a download link, but as far as I can tell the files needed are now embedded in another, much larger zip file (https://data.caltech.edu/records/jx9cx-fdh55) and it would take more work in the code to extract the stereo params files from caltech. The original link is dead. To use the fix as I have written it the user needs to download the zip manually, so I included it in the repo to reflect that. But, the file belongs to the caltech group so that probably isn't a usable solution.

MMathisLab (approved):

Issue #2449, opened 2023-11-17
 IndexError: list index out of range when using analyze_videos


trexbone:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Bug description

Hi!
This is my first time using python and DeepLabCut, so any suggestions or comments are greatly appreciated! When I was trying to analyze the video file, it was running fine until it failed halfway. Then I tried the suggestion in #1527 but the same error occured again. Due to the capacity of the gpu, I am running on batchsize=1.


### Operating System

Ubuntu 20.04.6 LTS

### DeepLabCut version

2.3.8

### DeepLabCut mode

multi animal

### Device type

gpu

### Steps To Reproduce

1. Enter conda environment by conda activate DEEPLABCUTAUD
2. Switch to ipython
3. import deeplabcut
4. Analyze videos
5. Error appears after analyzing two videos

### Relevant log output

```shell
In [28]: deeplabcut.analyze_videos('/home/fishlab4/Fish_audrey-Audrey-2023-10-31/config.yaml',['/home/fishlab4/Fish_audrey-Audrey-2023-10-31/videos'],videotype='.MP4',auto_track=True,allow_growth=True)
Using snapshot-500000 for model /home/fishlab4/Fish_audrey-Audrey-2023-10-31/dlc-models/iteration-0/Fish_audreyOct31-trainset95shuffle1
/home/fishlab4/anaconda3/envs/DEEPLABCUTAUD/lib/python3.9/site-packages/tensorflow/python/keras/engine/base_layer_v1.py:1694: UserWarning: `layer.apply` is deprecated and will be removed in a future version. Please use `layer.__call__` method instead.
  warnings.warn('`layer.apply` is deprecated and '
Activating extracting of PAFs
2023-11-15 11:33:07.319245: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1616] Created device /job:localhost/replica:0/task:0/device:GPU:0 with 1371 MB memory:  -> device: 0, name: Quadro P620, pci bus id: 0000:18:00.0, compute capability: 6.1
Analyzing all the videos in the directory...
Starting to analyze %  /home/fishlab4/Fish_audrey-Audrey-2023-10-31/videos/L.bicolor1_(Tank23)_Non-Resident1.MP4
Loading  /home/fishlab4/Fish_audrey-Audrey-2023-10-31/videos/L.bicolor1_(Tank23)_Non-Resident1.MP4
[...]

PR #2467, opened 2023-12-18
bug fix: empty predictions during video analysis

n-poulsen:
This PR addresses #2449

During video analysis, we do not check that there are detections for a frame before unpacking them. This PR addresses that in the same way that it's dealt with in the evaluation code


jeylau (approved):
Good catch, thanks Niels!

MMathisLab (approved):

Issue #2463, opened 2023-12-14
More informative error message when string instead of list is passed to add_new_videos()


tsievert:
If you accidentally pass a string for video locations to `add_new_videos()` you get an uninformative error and DLC tries to slice the string.
Another side effect is that it creates a series of single-letter symlinks (for `copy_videos=False`) and a bunch of single-letter folders in the `labelled-data` folder.

Codes that causes the error:
```
import deeplabcut

config_file_path = "/gpfs/projects/acad/behavior/DLC_model/primary_DeepLabCut_model/LassanceLab-main-model-2023-09-14/config.yaml"

video_file_path = "/gpfs/projects/acad/behavior/videos/2023-11-13_conspecific_odor_videos/2023-09-15_test videos/Basler_acA1440-220um__40264738__20230912_112719915.mp4"

VideoType = 'mp4'

deeplabcut.add_new_videos(
    config_file_path,
    video_file_path,
    copy_videos=True,
    coords=None,
    extract_frames=False,
)
```

Uninformative error message:
```
Copying the videos
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
Cell In[12], line 1
----> 1 deeplabcut.add_new_videos(
      2     config_file_path,
      3     video_file_path,
      4     copy_videos=True,
      5     coords=None,
      6     extract_frames=False,
      7 )

File ~/.local/lib/python3.10/site-packages/deeplabcut/create_project/add.py:80, in add_new_videos(config, videos, copy_videos, coords, extract_frames)
     78         else:
     79             print("Copying the videos")
---> 80             shutil.copy(os.fspath(src), os.fspath(dst))
     82 else:
     83     # creates the symlinks of the video and puts it in the videos directory.
     84     print("Attempting to create a symbolic link of the video ...")

File /gpfs/softs/easybuild/software/Python/3.10.4-GCCcore-11.3.0/lib/python3.10/shutil.py:417, in copy(src, dst, follow_symlinks)
    415 if os.path.isdir(dst):
    416     dst = os.path.join(dst, os.path.basename(src))
--> 417 copyfile(src, dst, follow_symlinks=follow_symlinks)
    418 copymode(src, dst, follow_symlinks=follow_symlinks)
    419 return dst

File /gpfs/softs/easybuild/software/Python/3.10.4-GCCcore-11.3.0/lib/python3.10/shutil.py:254, in copyfile(src, dst, follow_symlinks)
[...]

Issue #2461, opened 2023-12-13
Error in add_new_videos(extract_frames=True) but not extract_frames()


tsievert:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Bug description

I'm currently testing DLC on our HPC and noticed a strange behavior between `add_new_videos(extract_frames=True)` and `extract_frames()`.

When I try to extract frames via `add_new_videos(extract_frames=True)` I always get `Frame extraction failed. Video files must be corrupted.` but `extract_frames()` still works.
Switching between `copy_videos=False` and `copy_videos=True` has no effect.

### Operating System

Operating System: Red Hat Enterprise Linux 8.6 (Ootpa)
CPE OS Name: cpe:/o:redhat:enterprise_linux:8::baseos
Kernel: Linux 4.18.0-372.26.1.el8_6.x86_64

### DeepLabCut version

2.3.8

### DeepLabCut mode

single animal

### Device type

gpu

### Steps To Reproduce

This causes the error:
```
import deeplabcut

config_file_path = '/gpfs/projects/acad/behavior/DLC_model/primary_DeepLabCut_model/LassanceLab-main-model-2023-09-14/config.yaml'

video_file_path = ["/gpfs/projects/acad/behavior/videos/2023-11-13_conspecific_odor_videos/2023-09-15_test videos/cluster_test_vid.mp4"]

VideoType = 'mp4'

deeplabcut.add_new_videos(
    config_file_path,
    video_file_path,
    copy_videos=False,
    coords=None,
    extract_frames=True,
)
```

If you now run `deeplabcut.extract_frames(config_file_path, mode='automatic', algo='kmeans', userfeedback=False, crop=False)` the extraction suddently works.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/DeepLabCut/DeepLabCut/blob/master/CODE_OF_CONDUCT.md)

KonradDanielewski:
Does it help if in you move this [line](https://github.com/DeepLabCut/DeepLabCut/blob/0f657cde513dad207c68e9bb16eab8d5c351fde5/deeplabcut/create_project/add.py#L141) to be above `if extract_frames:` (line 130)?

[...]

PR #2468, opened 2023-12-18
Bug fix: `add_new_videos(..., extract_frames=True)`

n-poulsen:
Addresses issues #2461 and #2463

When running `add_new_videos(..., extract_frames=True)`, the configuration file is written after `extract_frames` is called. So when getting the videos for which to extract new frames, the new videos are not yet in the project configuration:

```python
    # deeplabcut.extract_frames, lines 269-272
    if videos_list is None:
        videos = cfg.get("video_sets_original") or cfg["video_sets"]
    else:  # filter video_list by the ones in the config file
        videos = [v for v in cfg["video_sets"] if v in videos_list]  # none of videos_list are in cfg["video_sets"] yet!
    ...

    has_failed = []
    for video in videos:
        ...

    if all(has_failed):  # True as the list is empty
        print("Frame extraction failed. Video files must be corrupted.")
```

Also deals with the case when users pass a single video to add as a string: wraps it in a list to avoid the annoying issue in #2463.

### DIFF
--- a/deeplabcut/create_project/add.py
+++ b/deeplabcut/create_project/add.py
@@ -58,6 +58,10 @@ def add_new_videos(
     # Read the config file
     cfg = auxiliaryfunctions.read_config(config)
 
+    # deal with user passing a single video to add
+    if isinstance(videos, str):
+        videos = [videos]
+
     video_path = Path(config).parents[0] / "videos"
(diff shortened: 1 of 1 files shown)


MMathisLab (approved):
Lgtm

AlexEMG (approved):

PR #2507, opened 2024-02-02
Add GUI help button back

jeylau:
Reintroduces the Help button in all GUI tabs, displaying core DeepLabCut function docstrings.


n-poulsen (approved):
Looks great to me!

PR #2508, opened 2024-02-05
Add feature to evaluate network for specfic, named snapshots.

### LINKED ISSUE #2356: Request snapshot number parameter in evaluate_network() as alternative to snapshotindex.
I want to `evaluate_network()` on two shuffles that have a different number of snapshots in their train folders. The differing number of snapshots saved is a result of exceptions or time-outs while training on colab. Both shuffles have a particular snapshot in common that I would like to use to compare the two. It is not currently intuitive to do this. 

I could run `evaluate_network()` for each shuffle individually specifying the `snapshotindex` in the config file but doing this is not especially intuitive as one has to determine what the `snapshotindex` actually is. I wasn't able to find an explanation in the DLC documentation although from what I can see in the repo it can be determined by listing the snapshots in the order of increasing iterations and determining the index corresponding the desired snapshot. I believe there is a simpler alternative to this. 

A parameter explicitly specifying the exact snapshot number (eg. 'snapshot-120000') in the `evaluate_network()` function would simplify this greatly. 

Alternatively adding documentation clarifying how the snapshotindex is defined (beyond the values of '-1' and 'all') could be helpful.

Tetra-quark:
Adds a feature to pass a list of snapshot names `snapshots_to_evaluate` to the `evaluate_network` function as an alternative to using the `snapshotindex` parameter in the main `config.yaml` file. 

The new parameter allows multiple, explicitly named snapshots to be evaluated where previously only a single snapshot or `"all"` could be chosen.

This pull request closes DeepLabCut/DeepLabCut#2356.

Overview of changes:
* Add `snapshots_to_evaluate` parameter to both single and multi-animal evaluation. 
* Function `get_available_requested_snapshots` to handle the `snapshots_to_evaluate` parameter. 
* Function `get_snapshots_by_index` to refactor existing `snapshotindex` from config file functionality. 
* Function `get_snapshots_from_folder` to refactor getting an ordered list of snapshots from the train folder.
* Docstring for new functions and parameters.
* Unit tests for all three new functions. 
* Update `testscript.py` to use `snapshots_to_evaluate` in `evaluate_network` (however in doing so I have disabled the `snapshotindex` method, both should be in the test ideally).
* General refactoring, renaming and cleaning. 

Example usage:
```python
deeplabcut.evaluate_network(
    config_path, 
    snapshots_to_evaluate=["snapshot-50000", "snapshot-100000"],
)
```


MMathisLab:
@n-poulsen and @jeylau let's prioritize this -- I think it's a great addition, and can be worked into DLC3.0

n-poulsen (approved):
Made some minor linting changes, tested on single + multi-animal projects. Looks good to me (as soon as tests pass)!

Great PR, thanks @Tetra-quark!

PR #2524, opened 2024-02-19
Use video crop coordinates appropriately in create_video_with_all_detections

jeylau:
If `cropping=True`, crop coordinates are parsed from the config.yaml and used to shift the raw detections so they align with the video.


MMathisLab:
can we also pass cropped coordinates as in analyze_videos; that is often useful for debugging without editing main config.yaml

MMathisLab (approved):

PR #2525, opened 2024-02-19
Avoid infinite loop in `mine()` when there are no overlapping tracklets

jeylau:


deeplabcut/refine_training_dataset/stitch.py
```diff
@@ -606,6 +606,9 @@ def compute_max_gap(tracklets):
         return max_gap
 
     def mine(self, n_samples):
+        if not self._lu_overlap:
+            raise ValueError("No overlapping tracklets found.")
```
yeshaokai, line 610: Maybe add more information suggesting it's because only one animal is found from the assembly?

jeylau: We cannot be certain about it though: very sparse (bad) tracklets could also cause that. This warning if nm_animal==1 could be shown much higher up before creating a tracklet dataset. What do you think?

yeshaokai: Good point! Then I don't have other questions

yeshaokai (approved):

MMathisLab (approved):

PR #2538, opened 2024-03-13
Video analysis on cropped videos: bug fixes for `extract_outlier_frames` and `create_video_with_all_detections`

n-poulsen:
When running video analysis with cropping, the predicted poses are stored in the cropped image space. To get the coordinates of the pose in the space of the full video, the data stored in the `_metadata.pickle` can be used.

The `extract_outlier_frames` and `create_video_with_all_detections` did not take these offsets into account, meaning the video with all detections had its keypoints shifted, as did the machine labels.

This PR fixes both of these issues.

As `create_video_with_all_detections` has a `cropping`argument, the behavior is set such that:
1. If `cropping` is given as an argument, that value is used (overriding the value stored in the metadata)
2. If `cropping` is not given then metadata["data"]["cropping"] and metadata["data"]["cropping_parameters"] are used


MMathisLab (approved):
Lgtm!

Issue #2433, opened 2023-11-01
Check benchmark submissions for completeness


stes:
**Is your feature request related to a problem? Please describe.**

The deeplabcut benchmark scripts currently do not check if a submission is complete, e.g. if predictions for all test images are returned, as noted by @n-poulsen .

**Describe the solution you'd like**

A good solution would be to add a few lines to [this part](https://github.com/DeepLabCut/DeepLabCut/blob/25f8c5026a03168358360e70c98039a7d7ed1b32/deeplabcut/benchmark/base.py#L91) of the evaluation code, sth along the lines of

```python
predictions = self.get_predictions(name)
self._validate_predictions(name)
...

def _validate_predictions(self, predictions):
  # check if predictions contains too many images not contained in the ground truth
  # check if predictions contains not all images in the ground truth
  # (other potential tests)
  pass
```

MMathisLab:
@n-poulsen is this completed?

PR #2570, opened 2024-04-30
bug fix: missing images during evaluation

n-poulsen:
During the computation of benchmark scores, the train/test images are loaded from the assemblies file (predictions made by the user) using the train/test indices in the dataset documentation file. This isn't robust, as the order of images passed by the user might be different to the ones in the ground truth file.

This PR updates the code to use the ground truth training dataset file to obtain the paths of the training/test images to use for evaluation, and find those images in the assemblies file. If there are predictions for test images that are missing, a warning is raised a warning and evaluation is made as if no predictions are made for that image.

The same issue happened when computing OKS in DeepLabCut - missing predictions can lead to a shift in the indices. This has been fixed in `deeplabcut/pose_estimation_tensorflow/lib/crossvalutils.py`.

### DIFF
--- a/deeplabcut/benchmark/base.py
+++ b/deeplabcut/benchmark/base.py
@@ -23,6 +23,7 @@ into this evaluation framework, please feel free to extend the base classes
 
 import abc
 import dataclasses
+import warnings
 from typing import Iterable
 from typing import Tuple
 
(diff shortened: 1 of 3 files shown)


n-poulsen:
Addresses #2433

MMathisLab (approved):
Lgtm thanks!

Issue #2581, opened 2024-05-16
Matlplotlib after installed by CONDA missing matplotlib.cm.cmap


chz31:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Bug description

Hi, I am trying the [demo ](https://colab.research.google.com/github/DeepLabCut/DeepLabCut/blob/master/examples/COLAB/COLAB_DEMO_mouse_openfield.ipynb#scrollTo=RiDwIVf5-3H_). I used conda to install DeepLabCut.

However, I got this error of not finding `cmap `in `matplotlib` when I finished training and ran the evaluation  `deeplabcut.evaluate_network(path_config_file,plotting=True)`:

```
Results for 10000  training iterations: 95 1 train error: 4.09 pixels. Test error: 5.71  pixels.
With pcutoff of 0.4  train error: 4.09 pixels. Test error: 5.71 pixels
Thereby, the errors are given by the average distances between the labels by DLC and the scorer.
Plotting...
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/chz31/DeepLabCut/deeplabcut/pose_estimation_tensorflow/core/evaluate.py", line 1019, in evaluate_network
    Plotting(
  File "/home/chz31/DeepLabCut/deeplabcut/pose_estimation_tensorflow/core/evaluate.py", line 201, in Plotting
    colors = visualization.get_cmap(len(comparisonbodyparts), name=cfg["colormap"])
  File "/home/chz31/DeepLabCut/deeplabcut/utils/visualization.py", line 36, in get_cmap
    return plt.cm.get_cmap(name, n)
AttributeError: module 'matplotlib.cm' has no attribute 'get_cmap'
```
After activating `conda activate DEEPLABCUT`, the Python used is 3.9. 19 in the conda environment. In the base environment, matplotlib.cm.cmap still exists with Python 3.11 but the method appeared to be deprecated.

Tensorflow 2.10 also could not find CUDA. The CUDA installed in my machine is 11.8, which appears to not be compatible with Tensorflow, but I think it's not relevant to this issue.

### Operating System

WSL2 (Ubuntu) under Windows 11

### DeepLabCut version

dlc version 2.3.9

### DeepLabCut mode

single animal

### Device type

gpu NVIDIA Geforce GTX 1070 Max Q
nvidia-smi returns NVIDIA-SMI 550.76.01              Driver Version: 552.22         CUDA Version: 11.8

[...]

PR #2583, opened 2024-05-20
pin matplotlib version

n-poulsen:
Addresses issue #2581. Replicated in Google Colab and locally. We use this method in [many functions](https://github.com/DeepLabCut/DeepLabCut/blob/fbaf36aee285564f18f78384bb94cc20c116297c/deeplabcut/utils/visualization.py#L33)

With `matplotlib==3.8.4`:

```
>>> import matplotlib as mpl
>>> print(mpl.__version__)
3.8.4
>>> import matplotlib.pyplot as plt
>>> plt.cm.get_cmap
<function _get_cmap at 0x11e9f4430>
```

With `matplotlib==3.9.0`:

```
>>> import matplotlib as mpl
>>> print(mpl.__version__)
3.9.0
>>> import matplotlib.pyplot as plt
>>> plt.cm.get_cmap
  File "<stdin>", line 1, in <module>
AttributeError: module 'matplotlib.cm' has no attribute 'get_cmap'
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ setuptools.setup(
         "imgaug>=0.4.0",
         "imageio-ffmpeg",
         "numba>=0.54",
-        "matplotlib>=3.3,,<3.9,!=3.7.0,!=3.7.1",
+        "matplotlib>=3.3,<3.9,!=3.7.0,!=3.7.1",
         "networkx>=2.6",
         "numpy>=1.18.5",
         "pandas>=1.0.1,!=1.5.0",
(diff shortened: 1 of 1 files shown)


MMathisLab (approved):

NadjaBrun:
Hi! Thanks for working on this!

I use google colab and just tried to start evaluating. However, the error message remains: 
AttributeError: module 'matplotlib.cm' has no attribute 'get_cmap' 

Any help is much appreciated!

MMathisLab:
yes, matplotlib just changed, so you will need to explicitly call and older version by running in Colab: `!pip install matplotlib==(older version)` -- or install from source our package `!pip install git+https://github.com/DeepLabCut/DeepLabCut.git`; DeepLabCut is not updated on pypi  only the repo here

NadjaBrun:
The evaluation step worked! Thank you so much!

PR #2586, opened 2024-05-23
Update label and warn users of the absence of videos

### LINKED ISSUE #2584: DeepLabCut extract frames GUI feature
Improving the frame extraction interface in the GUI and make it more user friendly. 

When the user wants to extract frames manually it is necessary to select a video. Currently the video selection has the title: Optional frame extraction from a video subset. While this title is informative is not linked to the manual extraction. 
For example: Frame Extraction from a Video Subset (Mandatory Step for Manual Extraction)

In edition when the user tries to extract frames manually without choosing a video first there is an error that is not clear
'Traceback (most recent call last):
  File "/Users/tomdeadman/miniforge3/envs/DEEPLABCUT_M1/lib/python3.9/site-packages/deeplabcut/gui/tabs/extract_frames.py", line 197, in extract_frames
    _ = launch_napari(list(self.video_selection_widget.files)[0])
IndexError: list index out of range'

We could add some information. An example:
No video file selected. Please upload a video to proceed with manual frame extraction.

jeylau:
Fixes #2584

### DIFF
--- a/deeplabcut/gui/tabs/extract_frames.py
+++ b/deeplabcut/gui/tabs/extract_frames.py
@@ -201,10 +201,13 @@ class ExtractFrames(DefaultTab):
                 QtWidgets.QMessageBox.critical(
                     self,
                     "Error",
-                    "Please select at least one video to extract frames from.",
+                    "Please select exactly one video to extract frames from.",
                 )
                 return
-            video_path_in_folder = self._check_symlink(videos[0])
+            first_video = videos[0]
+            if len(videos) > 1:
+                self.root.writer.write(f"Only the first video ({first_video}) will be opened.")
+            video_path_in_folder = self._check_symlink(first_video)
             _ = launch_napari(str(video_path_in_folder))
             return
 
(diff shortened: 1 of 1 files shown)


deeplabcut/gui/tabs/extract_frames.py
```diff
@@ -197,6 +197,13 @@ def extract_frames(self):
         mode = self.extraction_method_widget.currentText()
         if mode == "manual":
             videos = list(self.video_selection_widget.files)
+            if not videos:
+                QtWidgets.QMessageBox.critical(
+                    self,
+                    "Error",
+                    "Please select at least one video to extract frames from.",
```
n-poulsen: I would change the wording here to "... exactly one ..." as only we will only open frame extraction for one video.

We could print a warning when more than 1 video is selected (saying "only the first video ({video-name}) will be opened to manually extract frames from")

n-poulsen (approved):
Looks good to me, thanks @jeylau!

Issue #2590, opened 2024-06-02
Update the YAML file for Apple Silicon (M1, M2, M3) Macs, new working environment file below.


slagathor69:
**Is your feature request related to a problem? Please describe.**
The current yaml file does not correctly install on Apple Silicon Macs using anaconda. There are pandas, tables, tensorflow, and other errors when trying to run.

**Describe the solution you'd like**
I have created an updated yaml file that works on Apple Silicon Macs that I recommend to be added to installation instructions with the added bonus of working with Python 3.10.

**Describe alternatives you've considered**
I tried original way and after a lot of troubleshooting was able to create a functional, simpler yaml environment file for installing DeepLabCut seamlessly on Apple Silicon Macs (I have not tested it however on any Intel Macs).

**Additional context**
Updated YAML File below:

```
name: deeplabcut
channels:
  - conda-forge
  - defaults
dependencies:
  - python=3.10
  - pip
  - ipython
  - jupyter
  - python.app
  - ffmpeg
  - pytables
  - pandas
  - pip:
    - tensorflow
    - tensorflow-macos==2.9.0
    - tensorflow-metal==0.5.0
    - "deeplabcut[gui,apple_mchips]"
    - deeplabcut==2.3.9
```

MMathisLab:
thanks! Also note this month we are switching to pytorch, making this pain all go away!

MMathisLab:
I made a PR here, if you can check it @slagathor69 https://github.com/DeepLabCut/DeepLabCut/pull/2595

fffllllxxxxxxxx:
I kept failing to install it on my M3 Pro macbook
i tried this .YAML, but i got error again.
NEED SOME HELP.

<img width="1497" alt="image" src="https://github.com/user-attachments/assets/b9b4a975-4af3-4b4e-9e4e-6dfe0ccb72a1">

Scientific1:
updated everything just 'slightly' and now it should work (because of different dependencies) on apple m4pro chips. Hope it helps someone:

name: deeplabcut
channels:
  - conda-forge
  - defaults
dependencies:
  - python=3.10
  - pip
  - ipython
  - jupyter
  - python.app
  - ffmpeg
  - pytables
  - pandas
  - pip:
     - tensorflow-macos==2.12.0
     - tensorflow-metal
     - deeplabcut[apple-mchips,gui]==2.3.9

Scientific1:
[...]

PR #2595, opened 2024-06-05
Create DEEPLABCUT_M2.yaml

MMathisLab:
- user-supplied conda file, see https://github.com/DeepLabCut/DeepLabCut/issues/2590
- note, this will be depreciated once pytorch version is available

### DIFF
--- a/conda-environments/DEEPLABCUT_M2.yaml
+++ b/conda-environments/DEEPLABCUT_M2.yaml
@@ -1,0 +1,47 @@
+# DEEPLABCUT_M2.yaml
+# user supplied cond file, see issue: https://github.com/DeepLabCut/DeepLabCut/issues/2590
+
+#DeepLabCut2.0 Toolbox (deeplabcut.org)
+#© A. & M.W. Mathis Labs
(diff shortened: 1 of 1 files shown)


slagathor69:
My mistake, one last thing is to remove "tensorflow" from pip packages and only have tensorflow-macos and tensorflow-metal with those specified versions

MMathisLab:
@slagathor69 looking at this more, isn't the best way to just change the pin version in setup file? As now what is happenin is you install  tf-macos, tf-metal - with a pinned version, then you install DLC and over-write this with out apple_mchips versions... then re-install DLC again:

```bash
    - tensorflow-macos==2.9.0
    - tensorflow-metal==0.5.0
    - "deeplabcut[gui,apple_mchips]"
    - deeplabcut>=2.3.9
  ```

The best would be to just change the pins in setup.py to 2.9.0 and 0.5.0 if that is the only version that works for you? (note this is not true on my M2 chip macos).

MMathisLab:
@slagathor69 I am going to close, as it will be rapidly outdated by the end of the weekend most likely; see incoming PyTorch docs/code, i.e #2605 💜🔥

PR #2600, opened 2024-06-07
Auto check for dlclibrary & napari-deeplabcut updates when the GUI opens

jeylau:
Quite a number of users seem to have up-to-date DLC versions, but outdated labeling plugin. This verification would now be automatic (instead of manual via the Help menu).


MMathisLab (approved):

MMathisLab:
lgtm!

PR #2664, opened 2024-07-11
[resolved] Fix typo in pytorch data transforms

stes:
Fix python syntax in `transforms.py`. Note that I only fixed the typo and didn't check the resulting function for correctness.


stes:
sorry, nevermind, seems like this is indeed intended and the `:=` op was meant --- the file was flagged by my code formatter (which didnt seem to support that syntax yet).

Closing.

Issue #2667, opened 2024-07-15
DLC3 MA, hrnet_w18, epoch predictions broke with unique_bodyparts


Zelin2001:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Bug description

I installed with conda yaml file (pip install "git+https://github.com/DeepLabCut/DeepLabCut.git@pytorch_dlc#egg=deeplabcut[gui,modelzoo,wandb]"), and started my first multi-animal project, with hrnet_w18 model type.

As I train the network, I saw "Training for epoch 1 done, starting evaluating", followed by a fail and Error message:

```
File /home/username/anaconda3/envs/deeplabcut-2.3.10/lib/python3.10/site-packages/deeplabcut/pose_estimation_pytorch/apis/train.py:189, in train(loader, run_config, task, device, gpus, logger_config, snapshot_path, transform, inference_transform, max_snapshots_to_keep)
    186 else:
    187     logging.info("\nStarting pose model training...\n" + (50 * "-"))
--> 189 runner.fit(
    190     train_dataloader,
    191     valid_dataloader,
    192     epochs=run_config["train_settings"]["epochs"],
    193     display_iters=run_config["train_settings"]["display_iters"],
    194 )

File /home/username/anaconda3/envs/deeplabcut-2.3.10/lib/python3.10/site-packages/deeplabcut/pose_estimation_pytorch/runners/train.py:181, in TrainingRunner.fit(self, train_loader, valid_loader, epochs, display_iters)
    179 with torch.no_grad():
    180     logging.info(f"Training for epoch {e} done, starting evaluation")
--> 181     valid_loss = self._epoch(
    182         valid_loader, mode="eval", display_iters=display_iters
    183     )
    184     if self._print_valid_loss:
    185         msg += f", valid loss {float(valid_loss):.5f}"

File /home/username/anaconda3/envs/deeplabcut-2.3.10/lib/python3.10/site-packages/deeplabcut/pose_estimation_pytorch/runners/train.py:221, in TrainingRunner._epoch(self, loader, mode, display_iters)
    219 loss_metrics = defaultdict(list)
    220 for i, batch in enumerate(loader):
--> 221     losses_dict = self.step(batch, mode)
    222     epoch_loss.append(losses_dict["total_loss"])
    224     for key in losses_dict.keys():

[...]

PR #2668, opened 2024-07-15
Update train.py: avoids error on 'unique_bodyparts'

Zelin2001:
- See Issue #2667.

This is a possible solution that works smoothly in my environment. However, I have not figured out why it is necessary to reset `keypoints[i][3]` to non-negative values so I left the process alone.


jeylau:
Thanks @Zelin2001, I find your solution very elegant.
I suppose it could be further simplified to:
```python
if name == 'unique_bodyparts':
     keypoints[-1] = max(0, keypoints[-1])
```
but that's really a detail.
Thanks a lot for the PR 😊

jeylau (approved):

PR #2678, opened 2024-07-19
Default PAF config update

n-poulsen:
This pull request updates the default PAF configuration file to match the tensorflow version:

- Plateau generator instead of Gaussian
- BCE instead of MSE heatmap loss
- Bug fix: PAF model stride computation


jeylau (approved):
Looking good, thanks @n-poulsen!

Issue #2631, opened 2024-06-21
Evaluation error with PAF heads: `ValueError: matrix contains invalid numeric entries`


BBrianZhang:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Bug description

Hello,
while using Deeplabcut with a Pytorch engine, I encountered an issue with the model dlcrnet_stride16_ms5. During the training process, a ValueError occurred stating "matrix contains invalid numeric entries." No matter how much I reduce the learning rate or adjust the training batch size, I cannot resolve this problem.

### Operating System

Ubuntu 18.04

### DeepLabCut version

DLC 3.0.0rc1

### DeepLabCut mode

multi animal

### Device type

gpu

### Steps To Reproduce

1.creating a training dataset
2.pytorch_config.yaml as follows

```
data:
  colormode: RGB
  inference:
    normalize_images: true
  train:
    affine:
      p: 0.5
      rotation: 30
      scaling:
      - 1.0
      - 1.0
      translation: 0
    collate:
      type: ResizeFromDataSizeCollate
      min_scale: 0.4
      max_scale: 1.0
      min_short_side: 128
      max_short_side: 1152
      multiple_of: 32
      to_square: false
    covering: false
    gaussian_noise: 12.75
    hist_eq: false
    motion_blur: false
    normalize_images: true
device: auto
metadata:
  project_path: /mnt/Data16Tb/Data/boyang/pose/MMVISV3-BRIAN-2024-06-19
  pose_config_path: 
    /mnt/Data16Tb/Data/boyang/pose/MMVISV3-BRIAN-2024-06-19/dlc-models-pytorch/iteration-3/MMVISV3Jun19-trainset70shuffle0/train/pose_cfg.yaml
  bodyparts:
  - Front
  - Right
  - Middle
  - Left
  - FL1
  - BL1
  - FR1
  - BR1
  - BL2
  - BR2
  - FL2
  - FR2
  - Body1
  - Body2
  - Body3
  unique_bodyparts: []
  individuals:
  - MARMOSET_1
  - MARMOSET_2
  with_identity: true
method: bu
model:
  backbone:
    type: DLCRNet
    model_name: resnet50
    pretrained: true
    output_stride: 16
  backbone_output_channels: 2304
  pose_model:
    stride: 8
  heads:
    bodypart:
      type: DLCRNetHead
      predictor:
        type: PartAffinityFieldPredictor
        num_animals: 2
        num_multibodyparts: 15
[...]

PR #2679, opened 2024-07-19
evaluation refactor

n-poulsen:
# Evaluation refactor

Improvements to the evaluation code, as well as new tests to ensure that mAP scores match the pycocotools implementation.

Change list:
- Moved all metric computation code to a `deeplabcut/core/metrics` folder (as metrics are computed with `numpy`)
- Cleaned metric computation code so the prediction/ground truth matching always happens
  - Refactored in a way such that no OOM errors should occur, even on very large datasets (>60k images)  
- Multi-animal RMSE: only compute RMSE using (ground-truth, detection) matches with non-zero RMSE
- Add `compute_detection_rmse` to compute "detection" RMSE, matching the DeepLabCut 2.X implementation
- Fixed the bug for PAF models documented in #2631


MMathisLab (approved):
lgtm, although I did not test

deeplabcut/core/metrics/api.py
```diff
@@ -0,0 +1,159 @@
...
+) -> dict:
+    """Computes pose estimation performance metrics
+
+    Given ground truth pose labels and predictions on a dataset, computes RMSE and pose
+    mAP/mAR using OKS.
+
+    The image paths in the ground_truth dict must be the same as the ones in the
+    predictions dict.
+
+    Single animal RMSE is computed by simply calculating the distance between each
```
MMathisLab: I would say Euclidean distance here @n-poulsen

deeplabcut/core/metrics/api.py
```diff
@@ -0,0 +1,159 @@
...
+    ground truth keypoint and the corresponding prediction.
+
+    Multi-animal RMSE is computed differently: predictions are first matched to ground
+    truth individuals using greedy OKS matching. RMSE is then computed only between
+    predictions and the ground truth pose they are matched to, only when the OKS is
```
MMathisLab, line 44: define OKS

deeplabcut/core/metrics/api.py
```diff
@@ -0,0 +1,159 @@
...
+    )
+    results = dict(rmse=rmse, rmse_pcutoff=rmse_pcutoff, **oks_scores)
+
+    if not single_animal:
+        det_rmse, det_rmse_p = distance_metrics.compute_detection_rmse(data, pcutoff)
+        results["rmse_detections"] = det_rmse
+        results["rmse_detections_pcutoff"] = det_rmse_p
+
+    if unique_bodypart_gt is not None:
+        # TODO: Should we integrate unique bodyparts to main RMSE?
```
MMathisLab: re:TODO yes

MMathisLab (approved):
but see suggested changes to docstrings

MMathisLab (approved):
maybe in main docs we need to add information about these new metrics as well @n-poulsen

PR #2699, opened 2024-07-31
Location Refinement for Top-Down models

n-poulsen:
Location refinement is also need for top-down models, especially ResNets - otherwise sub-pixel accuracy cannot be obtained (which makes the pose estimation task too difficult for models like a stride 16 ResNet50).


yeshaokai (approved):
LGTM

PR #2701, opened 2024-08-01
Allow processing of videos in MKV containers

ximion:
Hi!

Most of our videos that we want to analyze with DLC are in the [MKV/Matroska](https://en.wikipedia.org/wiki/Matroska) format, and we use the options MKV gives with regards to adding extra metadata quite a lot.
MKV is supported natively by FFmpeg (and pretty much anywhere else), so, allowing it in DeepLabCut is just a matter of allowing it in the filter dialogs for the GUI.

We have been testing this for months now, and it works like a charm!

Thank you for considering this patch!
Best,
    Matthias


MMathisLab (approved):

ximion:
Thank you! :-)

MMathisLab:
Thanks for your contribution! 🔥

PR #2708, opened 2024-08-07
batched inference

n-poulsen:
Implemented batched inference for video analysis.

- updated the InferenceRunner classes to support batched inference
- added a `detector_batch_size` attribute to the project `config.yaml` (with default value 1)
- added `batch_size` and `detector_batch_size` parameters to `video_inference_superanimal`
- added `batch_size` and `detector_batch_size` parameters to `analyze_videos ` (with default value `None`, which means that the batch sizes used are the ones defined in the `config.yaml`)
- added tests to ensure all frames are processed for top-down and bottom-up models


AlexEMG (commented):
Perfect!

PR #2709, opened 2024-08-08
Bug fixes: `find_closest_neighbors` + ellipse tracking

n-poulsen:
This pull request makes two bug fixes:

In `deeplabcut/core/crossvalutils.py` - `find_closest_neighbors`
- The [cKDTree.query](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.cKDTree.query.html) method returns `n` (the size of the input data) for missing nieghbors (_"Missing neighbors are indicated with `self.n`."_). This wasn't taken into account, so this value could be returned in the index list, leading to an out-of-bounds error.

In `deeplabcut/core/trackingutils.py` - `calc_similarity_with`
- Zero division as we didn't check the height/width of ellipses was greater than 0


AlexEMG (commented):
Great!

PR #2716, opened 2024-08-13
Benchmarking user guide

dikraMasrour:
User Guide for model performance benchmarking across different architectures, DLC versions, and frameworks (TensorFlow and PyTorch for instance)


MMathisLab:
this looks great, we should add it into the toc.yml file 🔥

MMathisLab (approved):

PR #2718, opened 2024-08-14
Fix minimalnumberofconnections

segashin:
I've noticed that convert_detections_to_tracklets always uses the default value for min_n_links. Based on the comment on inference_cfg.yaml sample, the value at minimalnumberofconnctions should be used instead.
https://github.com/DeepLabCut/DeepLabCut/blob/dd0ef5a2189eb1a2e8c29762784a5df259a6c3b8/deeplabcut/inference_cfg.yaml#L17-L18

This pull request makes the following change.
- Read `minimalnumberofconnections` from `inference_cfg.yaml` and use it as `min_n_links` when creating the Assembler.


n-poulsen (approved):
LGTM! Thanks for the fix, @segashin!

Issue #2734, opened 2024-09-04
Sleap to DLC


agosztolai:
Hello! 

It's good to see that DLC GUI has improved so much since I last used it a few years ago!

Do you have a functionality to convert projects created in Sleap to DLC? Before I write my own code, I wanted to ask if you have already done it:)

MMathisLab:
hi @agosztolai thanks! Nice to see you on here. We have not written any loading code, but we have a guide about formatting ones data (which I should move into the proper docs!) https://github.com/DeepLabCut/DeepLabCut/wiki/Using-labeled-data-in-DeepLabCut-that-was-annotated-elsewhere-(or-merge-across-labelers) but would be very happy for your contribution!

agosztolai:
Hi Mackenzie, I have written a python script that you can run in a terminal to convert a .slp file in SLEAP (containing the annotations, images, and skeleton metadata) into a DLC project. What would be the best way to share?

agosztolai:
I'll just post it here! 

```
#!/usr/bin/env python3
# -*- coding: utf-8 -*-
"""
Written by: Adam Gosztolai

Script to convert SLEAP project to DLC project

Usage:
$ python convertSLP2DLC.py --slp_file /path/to/<project>.pkg.slp --dlc_dir /path/to/dlc/dir

Arguments:
--slp_file    Path to the SLEAP project file (.pkg.slp)
--dlc_dir      Path to the output DLC project directory

Once done, run deeplabcut.convertcsv2h5('config.yaml', userfeedback=False) on the
config file to obtain the .h5 files that Deeplabcut needs.

"""

import argparse
import io
import json
import yaml
import os
import time

import h5py
import numpy as np
import pandas as pd

import PIL.Image as Image
import cv2


def extract_frames_from_pkg_slp(file_path, base_output_dir, project_name='converted_project', scorer='me'):

    config = {'Task': project_name,
              'scorer': scorer,
              'date': time.strftime("%Y-%m-%d"),
              'identity': None,
              'project_path': base_output_dir,
              'engine': 'pytorch',
              'video_sets': {},
              'start': 0,
              'stop': 1,
              'numframes2pick': 20,
              'skeleton_color': 'black',
              'pcutoff': 0.6,
              'dotsize': 12,
              'alphavalue': 0.6,
              'colormap': 'rainbow',
              'TrainingFraction': [0.95],
              'iteration': 0,
              'default_net_type': 'resnet_50',
[...]

PR #2736, opened 2024-09-08
adding docs for using data labeled elsewhere (formatting)

MMathisLab:
- This addresses only in part #2734. At least the guide to format data is in the docs then! Could be expanded if a new function in DLC is added!


n-poulsen (approved):
LGTM!

Issue #2733, opened 2024-09-03
evaluate_network method possibly searches for hardcoded config.yaml filename?


anthonyattacks:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Bug description

Hello,

Running through my pipeline, I've had one specific method fail despite every other method using the proper (in our case: d1.config.yaml) file. Here's my code, and here's the error:

code:
```
# %%capture --no-stderr evalModel

# Track / log time, start
timeStart, startCalc = datetime.now().time(), time.time()
print(f"Time began: {timeStart.strftime('%H:%M:%S')}")

deeplabcut.evaluate_network(configPath, plotting = True)
# print(configPath)

# Track / log time, end, elapsed
timeStop, stopCalc = datetime.now().time(), time.time()
print(f"Time completed: {timeStop.strftime('%H:%M:%S')}")
print(f"Time elapsed: {round((stopCalc - startCalc)/60, 2)} minutes")

# Config path prints correctly; DLC is doing something to the path in the background.

# Times:
# All Default Settings: 
```

I've checked with print, the variable the config file name is stored in is still correct. It looks like the method is looking for "config.yaml" instead of what I have input as the config filename. This doesn't happen with any other method in my pipeline. Looking at the error, it looks like it's hardcoded to look for config.yaml instead of *config.yaml? We have an interest in preserving config.yaml for individual camera names irrespective of directory path, so just changing the name of config.yaml to match DLC convention wouldn't solve the problem given our use-case. Let me know if there are any thoughts. We're using the conda environment, which (having updated the environment) doesn't appear to be on the most recent version of DLC / gives an update prompt, so it's possible that this was also fixed in the most recent version.

### Operating System

operating system: Ubuntu 24.04

### DeepLabCut version

dlc version: 2.2.2, I believe?

### DeepLabCut mode

single animal

### Device type

gpu: Nvidia - 3060

### Steps To Reproduce

In a conda environment, run:

deeplabcut.evaluate_network(configPath, plotting = True)

where configPath ends with a file that isn't exactly config.yaml (eg. d1.config.yaml)

### Relevant log output

```shell
{
	"name": "FileNotFoundError",
	"message": "Config file at /home/happy/programming/deepLabCut/finches/dlcProject/d1/finchD1-happy-2024-09-02/config.yaml not found. Please make sure that the file exists and/or that you passed the path of the config file correctly!",
	"stack": "---------------------------------------------------------------------------
[...]

Issue #2703, opened 2024-08-02
Error with encountering NaNs in dlc_pytorch


AnnaStuckert:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Bug description

When running model training using the dlc_pytorch branch, I got the following error:

```
Traceback (most recent call last):
  File "/media1/data/anna/script_facemap_full.py", line 61, in <module>
    main()
  File "/media1/data/anna/script_facemap_full.py", line 45, in main
    deeplabcut.train_network(config, shuffle=shuffle) #step 2 - to edit training parameters such as max epochs to run, edit thepytorch_config.yaml file under dlc-models
  File "/media1/data/anna/DeepLabCut/deeplabcut/compat.py", line 245, in train_network
    return train_network(
  File "/media1/data/anna/DeepLabCut/deeplabcut/pose_estimation_pytorch/apis/train.py", line 339, in train_network
    train(
  File "/media1/data/anna/DeepLabCut/deeplabcut/pose_estimation_pytorch/apis/train.py", line 189, in train
    runner.fit(
  File "/media1/data/anna/DeepLabCut/deeplabcut/pose_estimation_pytorch/runners/train.py", line 177, in fit
    valid_loss = self._epoch(
  File "/media1/data/anna/DeepLabCut/deeplabcut/pose_estimation_pytorch/runners/train.py", line 217, in _epoch
    losses_dict = self.step(batch, mode)
  File "/media1/data/anna/DeepLabCut/deeplabcut/pose_estimation_pytorch/runners/train.py", line 321, in step
    self.logger.log_images(batch, outputs, target, step=self.current_epoch)
  File "/media1/data/anna/DeepLabCut/deeplabcut/pose_estimation_pytorch/runners/logger.py", line 326, in log_images
    images = self._prepare_images(inputs, outputs, targets)
  File "/media1/data/anna/DeepLabCut/deeplabcut/pose_estimation_pytorch/runners/logger.py", line 227, in _prepare_images
    image_logs[f"{base}.input"] = self._prepare_image(
  File "/media1/data/anna/DeepLabCut/deeplabcut/pose_estimation_pytorch/runners/logger.py", line 198, in _prepare_image
    image = draw_keypoints(
  File "/media1/data/anna/miniconda/envs/dlc/lib/python3.10/site-packages/torch/utils/_contextlib.py", line 116, in decorate_context
    return func(*args, **kwargs)
[...]

PR #2747, opened 2024-10-01
pytorch_dlc - Small updates and bug fixes

n-poulsen:
This pull requests makes some small updates and bug fixes:

Addresses #2703 to deal with NaNs when logging images to WandB
- The issue comes from NaNs being converted to integers 
- With `torchvision>=0.18`, we can use a visibility flag. I think that's too restrictive for now, added a comment and we can pin torchvision in the future.
- Tests were added to ensure this works.

Addresses #2733 for modified project configuration filenames
- This is for consistency in the codebase. It is not regular DeepLabCut workflow to have a project configuration file named anything other than `config.yaml`, and no other method is guaranteed to work.

Fixes a bug when calling `create_training_dataset` with a non-default detector for multi-animal projects:
- `detector_type=detector_type,` was not passed to `create_multianimaltraining_dataset`

When calling video inference with cropping=True, updates the metadata for the predictions correctly:
- So the keypoints are displayed correctly when calling `create_labeled_video` and analysis is correct

PAF predictor - Fixes a bug when there are no detections
- The assembler returns None instead of an empty list when no assemblies are found
- This deals with it by checking that the assemblies aren't None

Heatmap target generation - Fixes a bug where gradients might be masked if there is missing data
- also adds gradient masking configuration to the configuration files

### DIFF
--- a/deeplabcut/pose_estimation_pytorch/config/base/head_bodyparts.yaml
+++ b/deeplabcut/pose_estimation_pytorch/config/base/head_bodyparts.yaml
@@ -11,6 +11,7 @@ target_generator:
   num_heatmaps: "num_bodyparts"
   pos_dist_thresh: 17
   heatmap_mode: KEYPOINT
+  gradient_masking: false
   generate_locref: true
   locref_std: 7.2801
 criterion:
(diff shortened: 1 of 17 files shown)


deeplabcut/pose_estimation_pytorch/config/base/head_bodyparts.yaml
```diff
@@ -11,6 +11,7 @@ target_generator:
   num_heatmaps: "num_bodyparts"
   pos_dist_thresh: 17
   heatmap_mode: KEYPOINT
+  gradient_masking: false
```
MMathisLab, line 14: do we want it false by default?

MMathisLab (approved):
actually, i see the logic now -- lgtm!!

Issue #2744, opened 2024-09-28
The code doesn't seem to know when user specifies "gpus: [0, 1, 2, 3]" in the pytorch_config.yaml (PyTorch engine)


thuann2cats:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Bug description

Hi DLC team, I’m using the latest DLC version on the PyTorch engine. There seems to be an error with training this DLC engine on multiple GPUs.
Let me show you further:

If my pytorch_config.yaml is like this:
```
…
runner:
  type: PoseTrainingRunner
  gpus:
  - 0
  - 1
  - 2
  - 3
  key_metric: test.mAP
  key_metric_asc: true
  eval_interval: 300
…

```
I can verify that when the code executes to here: (at line 73 <DLC conda env folder>/lib/python3.10/site-packages/deeplabcut/pose_estimation_pytorch/apis/train.py), the run_config object has read the [0, 1, 2, 3] array of GPU IDs as seen below:

run_config
```
…
'model':{'backbone': {'type': 'ResNet', 'model_name': 'resnet50_gn', 'output_stride': 16, 'freeze_bn_stats': True, 'freeze_bn_weights': False}, 'backbone_output_channels': 2048, 'heads': {'bodypart': {...}}}
'net_type':'resnet_50'
'runner':{'type': 'PoseTrainingRunner', 'gpus': [0, 1, 2, 3], 'key_metric': 'test.mAP', 'key_metric_asc': True, 'eval_interval': 300, 'optimizer': {'type': 'AdamW', 'params': {...}}, 'scheduler': {'type': 'LRListScheduler', 'params': {...}}, 'snapshots': {'max_snapshots': 5, 'save_epochs': 25, 'save_optimizer_state': False}}
'train_settings':{'batch_size': 8, 'dataloader_workers': 0, 'dataloader_pin_memory': False, 'display_iters': 100, 'epochs': 200, 'seed': 42}
len():8
```

The code eventually executes to here, line 554 at this file: <DLC conda env folder>/lib/python3.10/site-packages/deeplabcut/pose_estimation_pytorch/runners/train.py, where the code builds a runner. It’s supposed to know that “gpus” is an array of GPU IDs. However, “gpus” remains None and thus the runner would not initiate the DataParallel().
```
def build_training_runner(
    runner_config: dict,
    model_folder: Path,
    task: Task,
    model: nn.Module,
    device: str,
    gpus: list[int] | None = None,
[...]

PR #2764, opened 2024-10-23
Fix reading gpus config

maximpavliv:
When not passed by argument to the train function, the gpus parameter is read from the pytorch_config.yaml file 
(Fix to issue [#2744](https://github.com/DeepLabCut/DeepLabCut/issues/2744))


AlexEMG (approved):

AlexEMG:
Fantastic, welcome to DLC!

Issue #2757, opened 2024-10-10
GUI option/feature to continue training from a saved snapshot


hkosiara:
I have been using deeplabcut for about 3 weeks, and after asking around on forums, I was told that DLC GUI does not have a way to continue training the network from a saved snapshot. That is, I have a snapshot of a network at 650 epochs, and I would like to train it for an additional 100 epochs, to 750. There is a way to do this in the terminal, but I use the GUI, and it seems like it's an important feature to have in both interfaces. There used to be a way to do it in the pose.cfg file with tensorflow, but I tried something similar with the new pytorch version to no avail. I think there should be an option to load a snapshot file path and continue training after that snapshot so that users can improve upon network training rather than having to start from 0 epochs each time more training is desired.

n-poulsen:
Hi @hkosiara! This is a great feature request, and I'll get to work on it now

austin2118ace:
@n-poulsen 
> Hi @hkosiara! This is a great feature request, and I'll get to work on it now

Is this feature available in the CLI version of DLC 3.0, or do we have to start from the pretrained weights each time? I looked through the config files as @hkosiara did and also didn't notice anywhere that I could load a snapshot.

n-poulsen:
@austin2118ace this is already available through the CLI for `DeepLabCut 3.0`. When calling [`train_network`](https://github.com/DeepLabCut/DeepLabCut/blob/17d72054a0382ecbabb6ba268da54539e5296bc9/deeplabcut/compat.py#L60) for PyTorch models, any additional parameters for the PyTorch implementation of [`train_network`](https://github.com/DeepLabCut/DeepLabCut/blob/17d72054a0382ecbabb6ba268da54539e5296bc9/deeplabcut/pose_estimation_pytorch/apis/train.py#L194) (found in `deeplabcut/pose_estimation_pytorch/apis/train.py`) can be passed through the [`**torch_kwargs`]((https://github.com/DeepLabCut/DeepLabCut/blob/17d72054a0382ecbabb6ba268da54539e5296bc9/deeplabcut/compat.py#L76)). So currently you can call:

```python
import deeplabcut

deeplabcut.train_network(
    "/path/to/my/dlc-project",
    shuffle=12,
    snapshot_path="/path/to/resume/training/from/snapshot-100.pt",
)
```

austin2118ace:
[...]

PR #2774, opened 2024-11-06
Resume training from existing snapshots option in GUI

maximpavliv:
Add widgets to GUI giving possibility to resume training from existing snapshots, both for pose model and detector.

This Pull Request adds the feature requested by [#2757](https://github.com/DeepLabCut/DeepLabCut/issues/2757) .

### DIFF
--- a/deeplabcut/gui/components.py
+++ b/deeplabcut/gui/components.py
@@ -184,6 +184,81 @@ class VideoSelectionWidget(QtWidgets.QWidget):
         self.root.logger.info(f"Cleared selected videos")
 
 
+class SnapshotSelectionWidget(QtWidgets.QWidget):
+    def __init__(
+        self,
+        root: QtWidgets.QMainWindow,
+        parent: QtWidgets.QWidget,
+        margins: tuple,
+        select_button_text: str,
+    ):
+        super(SnapshotSelectionWidget, self).__init__(parent)
+
+        self.root = root
+        self.parent = parent
+
+        self.selected_snapshot = None
+
+        self._init_layout(margins, select_button_text)
+
+    def _init_layout(self, margins, select_button_text):
+        layout = _create_horizontal_layout(margins=margins)
+
+        # Select videos
+        self.select_snapshot_button = QtWidgets.QPushButton(select_button_text)
+        self.select_snapshot_button.setMaximumWidth(200)
+        self.select_snapshot_button.clicked.connect(self.select_snapshot)
+
+        # Selected snapshot text
+        self.selected_snapshot_text = QtWidgets.QLabel(
+            ""
+        )  # updated when snapshot is selected
+
+        # Clear snapshot selection
+        self.clear_snapshot_button = QtWidgets.QPushButton("Clear selection")
+        self.clear_snapshot_button.clicked.connect(self.clear_selected_snapshot)
+        self.clear_snapshot_button.hide()
+
+        layout.addWidget(self.select_snapshot_button)
+        layout.addWidget(self.selected_snapshot_text)
+        layout.addWidget(self.clear_snapshot_button, alignment=Qt.AlignRight)
+
+        self.setLayout(layout)
+
+    def _update_selected_snapshot_display(self):
+        if self.selected_snapshot is None:
+            self.selected_snapshot_text.setText("")
+            self.clear_snapshot_button.hide()
+        else:
+            self.selected_snapshot_text.setText(
+                f"{os.path.basename(self.selected_snapshot)}"
+            )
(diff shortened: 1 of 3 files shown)


deeplabcut/gui/tabs/train_network.py
```diff
@@ -67,11 +68,43 @@ def _on_engine_change(self, engine: Engine) -> None:
...
+
+    def _update_snapshot_selection_widgets_visibility(self):
+        if self.root.engine == Engine.PYTORCH:
+            self.resume_from_snapshot_label.show()
+            self.snapshot_selection_widget.show()
+            self.detector_snapshot_selection_widget.show()
+        else:
+            self.resume_from_snapshot_label.hide()
+            self.snapshot_selection_widget.hide()
+            self.detector_snapshot_selection_widget.hide()
```
n-poulsen: The `detector_snapshot_selection_widget` should only be shown if the model is top-down (as is done line 252 for the `Detector parameters` - see `show_when_cfg` line 342)

deeplabcut/gui/tabs/train_network.py
```diff
@@ -67,11 +68,43 @@ def _on_engine_change(self, engine: Engine) -> None:
...
 
     def _set_page(self):
         self.main_layout.addWidget(_create_label_widget("Attributes", "font:bold"))
         self._generate_layout_attributes()
-        self.main_layout.addWidget(_create_label_widget(""))  # dummy label
+
+        self.resume_from_snapshot_label = _create_label_widget(
+            "Resume training from existing snapshots?",
```
n-poulsen: Here I'd add a note to say that it's optional; so maybe update the text to `[Optional]: Select a snapshot to resume training from`.

I'd also add a `tooltip` around this area that contains more information, e.g.:

`If you've already trained a model on this shuffle, you can continue training it instead of starting from scratch again. For top-down models, you can also choose a detector to resume training from.`

deeplabcut/gui/components.py
```diff
@@ -183,6 +181,70 @@ def clear_selected_videos(self):
...
+
+        self.root = root
+        self.parent = parent
+
+        self.selected_snapshot = None
+
+        self._init_layout(select_button_text)
+
+    def _init_layout(self, select_button_text):
+        layout = _create_horizontal_layout()
```
n-poulsen: Here I'd edit the margins a bit to make it look better (so the buttons are slightly to the right of the title):

```
layout = _create_horizontal_layout(margins=(30, 0, 0, 0))
``` 

The margins could also be a parameter passed to the `SnapshotSelectionWidget` so it's easier to choose based on where the `SnapshotSelectionWidget` is placed.

n-poulsen (changes requested):
Looks great to me! A few comments with a few changes to make, but almost ready to go!

deeplabcut/gui/tabs/train_network.py
maximpavliv: In order to have this working I had to move `self._pose_cfg_change()` init call, see corresponding commit. Normally this shouldn't break anything, but please double-check to be sure.

deeplabcut/gui/tabs/train_network.py
maximpavliv: I rephrased it to _[...]. When using top-down models, you can [...]_ .
So that the message still makes sense when the user is not in Top-Down mode. Is that OK for you?

deeplabcut/gui/tabs/train_network.py
```diff
@@ -67,13 +68,39 @@ def _on_engine_change(self, engine: Engine) -> None:
...
 
-        self._pose_cfg_change(self._shuffle_display.pose_cfg)
+        self.resume_from_snapshot_label = _create_label_widget(
+            "[Optional]: Select a snapshot to resume training from", "font:bold"
+        )
+        self.resume_from_snapshot_label.setToolTip(
+            "<span style='font-weight:normal; white-space:nowrap;'>"
+            "If you've already trained a model on this shuffle, you can continue training it instead of starting "
+            "from scratch again."
+            "</span>"
```
maximpavliv: @n-poulsen Had to do this CSS formatting thing in order to have the ToolTip displayed in regular fond (and not bold). Doesn't look nice, I know. Apparently there could be an other approach, `QToolTip.setFont()`, but I think it affects all ToolTips in the GUI, which is not what we want...
So any opinions welcome here.

n-poulsen: As this works I think we can keep it as-is

n-poulsen (approved):
Looks good to me, thanks @maximpavliv!

Issue #2749, opened 2024-10-03
Could not load a saved snapshot due to mismatched state_dict keys


thuann2cats:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Bug description

Hi DLC team, I’m running this evaluate function: 
`deeplabcut.evaluate_network(config_path, plotting=True)`

But I ran into an error with loading the state_dict on this line:
Line 67: `<environment folder>/lib/python3.10/site-packages/deeplabcut/pose_estimation_pytorch/runners/base.py`
```
@staticmethod
    def load_snapshot(
        snapshot_path: str | Path,
        device: str,
        model: ModelType,
        optimizer: torch.optim.Optimizer | None = None,
    ) -> int:
        """
        Args:
            snapshot_path: the path containing the model weights to load
            device: the device on which the model should be loaded
            model: the model for which the weights are loaded
            optimizer: if defined, the optimizer weights to load

        Returns:
            the number of epochs the model was trained for
        """
        snapshot = torch.load(snapshot_path, map_location=device)
        model.load_state_dict(snapshot['model'])
        if optimizer is not None and 'optimizer' in snapshot:
            optimizer.load_state_dict(snapshot["optimizer"])

        return snapshot.get("metadata", {}).get("epoch", 0)
```

right before `model.load_state_dict()`

model has these state dict keys:
```
model.state_dict()
[...]

Issue #2663, opened 2024-07-10
Automatic save model snapshots with best test error in DLC3.0


tom21100227:
**Is your feature request related to a problem? Please describe.**
In the new dlc3.0.0 release with PyTorch, some metrics are displayed after every epoch (test error / test recall/precision). But the model snapshots are saved solely on iteration/step/epoch numbers. When going through the logs I can see (for example) the metrics after epoch 178 is performing really well, but the only snapshots of models I have is 175 or 200. 

**Describe the solution you'd like**
Having an additional option during `train_network` that allows DLC to save the 5 model snapshots with the best `metrics/test.rmse` or the best `metrics/test.mAP_pcutoff`. 

**Describe alternatives you've considered**
I could make save epoch being 1 and just save all model snapshots after every epoch to cherry pick. But that would take up a lot of storage space and still kind of frustrating to navigate through a sea of snapshots. 

**Additional context**
N/A

n-poulsen:
Hi @tom21100227! I think that's a good idea - we're working on adding this to the code

PR #2775, opened 2024-11-07
Improvements to the training UX

n-poulsen:
Many minor improvements to the user experience when training models. Overview of changes (see below for description of each change):

- Patch pycocotools printing during bounding box evaluation (so we don't get annoying prints)
- Same mAP scale for pose and object detection metrics (0 to 100, not 0 to 1)
- Typing fixes
- Logging to `learning_stats.csv` (log detector and pose model to different files)
- Non-zero starting epoch (bug fix)
- Detector training - don't show evaluation loss
- Improved look of printed metrics during training
- #2663
- Add option to select weights to resume training from the `pytorch_config.yaml` file
- Fix: running PAF predictors on MPS
- #2749

### Patch pycocotools printing during bounding box evaluation

Evaluating object detection performance with `pycocotools` led to useless/confusing lines being printed:

```
creating index...
index created!
Loading and preparing results...
Converting ndarray to lists...
(2198, 7)
...
```

The print functions inside of `pycocotools` have been patched so these lines are no longer printed.

### Same mAP scale for pose and object detection metrics

Object detection mAP was reported between 0 and 1 (the default though pycocotools), while pose mAP was reported between 0 and 100. Both bounding box mAP and pose mAP is now reported between 0 and 100.

### Typing fixes

When calling `train_network`, both `snapshot_path` and `detector_path` can be strings, paths or None.

### Logging to `learning_stats.csv`

Both detector and pose model stats were logged to `learning_stats.csv`, so one would overwrite the other. This is no longer the case, with pose model stats logged to `learning_stats.csv` and detector stats are logged to `learning_stats_detector.csv`.

### Non-zero starting epoch

When continuing to train a model, if the epochs given were larger than the starting epoch (the number of epochs for which the given weights were trained), the model was only trained for `epochs - starting_epochs`. So in the example below, the 2nd call to `train_network` would only train for 5 extra epochs. This was so the model is always trained for the number of epochs passed as an argument (so that in the example below, the model would be trained for 10 extra epochs and the last snapshots output would be `snapshot-015.pt` and `snapshot-detector -015.pt`)

```python
import deeplabcut
from deeplabcut.pose_estimation_pytorch import DLCLoader

deeplabcut.train_network("dlc-project/config.yaml", shuffle=0, epochs=5, detector_epochs=5)
loader = DLCLoader("dlc-project/config.yaml", shuffle=0) 
deeplabcut.train_network(
    config="dlc-project/config.yaml",
    shuffle=0,
    epochs=10,
    detector_epochs=10,
    snapshot_path=loader.model_folder / f"snapshot-005.pt",
    detector_path=loader.model_folder / f"snapshot-detector-005.pt",
)
```

### Detector training - evaluation loss

Torchvision [object detection models cannot return both loss and predictions: it's one or the other](https://github.com/pytorch/vision/blob/868a3b42f4bffe29e4414ad7e4c7d9d0b4690ecb/torchvision/models/detection/generalized_rcnn.py#L39). When evaluating during training, the predictions are used to obtain mAP/mAR metrics, so the loss is `nan`. To avoid any confusion, the validation loss (which is NaN as we don't have it) is no longer printed.

### Printing metrics during training

Some visual improvements were made when logging metrics to the console during training.

When training detectors:

```
# Previous
...
Training for epoch 20 done, starting evaluation
creating index...
index created!
...
Accumulating evaluation results...
DONE (t=0.00s).
Epoch 20 performance:
metrics/test.mAP@50:95:0.134
metrics/test.mAP@50:0.514
metrics/test.mAP@75:0.042
metrics/test.mAR@50:95:0.300
metrics/test.mAR@50:0.833
metrics/test.mAR@75:0.167
Epoch 20/200 (lr=0.0001), train loss 4.69560
...

# ** New **
...
Epoch 20/200 (lr=0.0001), train loss 4.69560
Model performance:
  metrics/test.mAP@50:95:  13.37
  metrics/test.mAP@50:     51.37
  metrics/test.mAP@75:      4.21
  metrics/test.mAR@50:95:  30.00
  metrics/test.mAR@50:     83.33
  metrics/test.mAR@75:     16.67
...
```

When training pose estimation models:

```
# Previous
Epoch 193 performance:
metrics/test.rmse:  5.50
metrics/test.rmse_pcutoff:5.27
metrics/test.mAP:   100.000
metrics/test.mAR:   100.000
metrics/test.rmse_detections:23.134
metrics/test.rmse_detections_pcutoff:14.039
Epoch 193/200 (lr=0.0001), train loss 0.00039, valid loss 0.00027

# ** New **
Epoch 193/200 (lr=0.0001), train loss 0.00039, valid loss 0.00027
Model performance:
  metrics/test.rmse:                      5.50
  metrics/test.rmse_pcutoff:              5.27
  metrics/test.mAP:                     100.00
  metrics/test.mAR:                     100.00
  metrics/test.rmse_detections:           5.50
  metrics/test.rmse_detections_pcutoff:   5.27
```

### Saving the best snapshot

Addresses #2663 to save the best snapshot during training. The best snapshot will be saved as `snapshot-best-XYZ.pt`, where `XYZ` is the number of epochs for which it was trained.

### Resuming Training from a Given Snapshot 

Adds an option to the `pytorch_config.yaml` to resume training from an existing snapshot, with:

```yaml
...
detector:
    # weights from which to resume training the detector
    resume_training_from: /Users/john/.../train/snapshot-detector-250.pt 
...
# weights from which to resume training the pose model
resume_training_from: /Users/john/.../train/snapshot-100.pt
```

### Fix PAF predictor running on MPS

The PAF predictor would fail when running on MPS (macOS GPU), as torch.round(...) is not yet implemented. An easy fix was to run scripts with `PYTORCH_ENABLE_MPS_FALLBACK=1` set as an environment variable. This changes the operation to run with numpy so this fix is no longer needed.

### Multi-GPU training: fix saving the state dict

Addresses issue #2749.


docs/pytorch/pytorch_config.md
```diff
@@ -396,6 +396,20 @@ You can also log images as they are seen by the model to `wandb`
 with the `image_log_interval`. This logs a random train and test image, as well as the 
 targets and heatmaps for that image.
 
+### Restarting Training at a Specific Checkpoint
+
+If you wish to restart the training at a specific checkpoint, you can specify the
+full path of the checkpoint to the `init_weights` variable, as shown below. In this
```
maximpavliv:
```suggestion
full path of the checkpoint to the `resume_training_from` variable, as shown below. In this 
```

docs/pytorch/pytorch_config.md
```diff
@@ -449,3 +463,19 @@ detector that brings enough performance. The recommended variants are the follow
 | fasterrcnn_mobilenet_v3_large_fpn |                             32.8 |                           19.4M |                     4.49 |
 | fasterrcnn_resnet50_fpn           |                               37 |                           41.8M |                   134.38 |
 | fasterrcnn_resnet50_fpn_v2        |                             46.7 |                           43.7M |                   280.37 |
+
+
+### Restarting Training of an Object Detector at a Specific Checkpoint
+
+If you wish to restart the training of a detector at a specific checkpoint, you can
+specify the full path of the checkpoint to the detector's `init_weights` variable, as
```
maximpavliv:
```suggestion
specify the full path of the checkpoint to the detector's `resume_training_from` variable, as
```

deeplabcut/pose_estimation_pytorch/runners/snapshots.py
```diff
@@ -68,18 +75,21 @@ class TorchSnapshotManager:
...
     task: Task
     model_folder: Path
     key_metric: str | None = None
     key_metric_asc: bool = True
     max_snapshots: int = 5
     save_epochs: int = 25
     save_optimizer_state: bool = False
+    _best_model_epochs: int = -1
+    _best_metric: float | None = None
+    _key: str = field(init=False)
```
maximpavliv, line 88: No docstring because these are "private", is that right?

maximpavliv (approved):
Beside the variable names in the pytorch_config.md, this looks good to me!

deeplabcut/pose_estimation_pytorch/runners/snapshots.py
n-poulsen: exactly 👍

PR #2781, opened 2024-11-11
Use architecture-specific snapshots prefixes in video adaptation

maximpavliv:
When doing video adaptation with one model architecture and then re-doing with an other, video_inference_superanimal() was trying to load existing trained model with previous architecture, which resulted in a crash.

This PR uses architecture-specific prefixes to save snapshots when doing video adaptation, which resolves the issue.

### DIFF
--- a/deeplabcut/pose_estimation_pytorch/runners/train.py
+++ b/deeplabcut/pose_estimation_pytorch/runners/train.py
@@ -597,7 +597,10 @@ def build_training_runner(
     optim_cls = getattr(torch.optim, optim_cfg["type"])
     optimizer = optim_cls(params=model.parameters(), **optim_cfg["params"])
     scheduler = build_scheduler(runner_config.get("scheduler"), optimizer)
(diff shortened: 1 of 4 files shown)


deeplabcut/pose_estimation_pytorch/runners/train.py
```diff
@@ -597,11 +597,15 @@ def build_training_runner(
     optim_cls = getattr(torch.optim, optim_cfg["type"])
     optimizer = optim_cls(params=model.parameters(), **optim_cfg["params"])
     scheduler = build_scheduler(runner_config.get("scheduler"), optimizer)
+    # if no custom snapshot prefix is defined, use the deafult one
```
n-poulsen:
```suggestion
    # if no custom snapshot prefix is defined, use the default one
```

n-poulsen (changes requested):
Looks good to me! Tiny typo to fix

n-poulsen (approved):

PR #2782, opened 2024-11-11
Timeout retrieving latest package version from the Web

maximpavliv:
This morning, I had deeplabcut GUI stuck in [utils.is_latest_deeplabcut_version()](https://github.com/DeepLabCut/DeepLabCut/blob/48c5969408b10f66bd9ec337d1c4b7d2358e9e2b/deeplabcut/gui/window.py#L45C15-L45C31) and in [misc.is_latest_version()](https://github.com/DeepLabCut/DeepLabCut/blob/48c5969408b10f66bd9ec337d1c4b7d2358e9e2b/deeplabcut/gui/window.py#L46). After some investigation, it seems like 

```
url = "https://pypi.org/pypi/deeplabcut/json"
contents = urllib.request.urlopen(url).read()
```
was stuck because of some firewall or DNS in Python, but it was working when I was getting the data from the CLI (using `curl`) or in a web browser.
This case is not handeled by the [except URLError block of _check_for_updates()](https://github.com/DeepLabCut/DeepLabCut/blob/48c5969408b10f66bd9ec337d1c4b7d2358e9e2b/deeplabcut/gui/window.py#L47), this PR adds a timeout to trying to read the latest available package versions.

### DIFF
--- a/deeplabcut/gui/utils.py
+++ b/deeplabcut/gui/utils.py
@@ -8,7 +8,10 @@
 #
 # Licensed under GNU Lesser General Public License v3.0
 #
+from typing import Tuple
+
 from PySide6 import QtCore
+import re
 
 
 class Worker(QtCore.QObject):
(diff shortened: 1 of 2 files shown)


n-poulsen (approved):
Great fix thanks Maxim (really nice to not have the update message when on an RC version)

Issue #2771, opened 2024-10-30
DLC 3.0.0rc - video analysis with identity prediction memory issues


SGasperoni:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Bug description

Hi,
I'm running the analysis step of multi-animal dlc, in videos with 4 mice and 8 bodyparts per mouse. The frame size is 500 x 450, and videos last 120 minutes. After few minutes of the analysis step, I run out of memory. 
When I check on nvidia-smi, I see that GPU seems to be used, but not its memory, resulting in overuse of RAM memory, leading to the issue.
![image](https://github.com/user-attachments/assets/ca0115ee-e4cb-4c2c-aaff-339ee6c16144)

I had previously run larger videos with the same setup, in DLC2.3.3, and never encountered this problem.

* I have followed the instructions to have GPU activated when using pytorch engine instead of tensorflow. 

Any help will be greatly appreciated.

### Operating System

Windows 11

### DeepLabCut version

dlc 3.0.0rc

### DeepLabCut mode

multi animal

### Device type

NVIDIA GeForce RTX 4090

### Steps To Reproduce

Config file:
![image](https://github.com/user-attachments/assets/62e442d9-a76f-4822-be57-5d2256e63038)
![image](https://github.com/user-attachments/assets/43bc869b-8455-436b-8471-cabfefca94a6)

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/DeepLabCut/DeepLabCut/blob/master/CODE_OF_CONDUCT.md)

n-poulsen:
Hi @Sere-98, I haven't experienced this issue during testing. Could you let me know the exact DeepLabCut version that you're running (you can report the output of `pip list`)

SGasperoni:
Hi @n-poulsen, here it is:

```
Package                       Version
----------------------------- --------------
alabaster                     0.7.16
albumentations                1.4.3
anyio                         4.6.2.post1
app-model                     0.2.8
appdirs                       1.4.4
argon2-cffi                   23.1.0
argon2-cffi-bindings          21.2.0
arrow                         1.3.0
asttokens                     2.4.1
async-lru                     2.0.4
attrs                         24.2.0
Babel                         2.14.0
beautifulsoup4                4.12.3
bleach                        6.1.0
Brotli                        1.1.0
build                         1.2.2.post1
cached-property               1.5.2
cachey                        0.2.1
certifi                       2024.8.30
[...]

PR #2785, opened 2024-11-13
Video Analysis with Identity - Lowering memory footprint

n-poulsen:
This PR fixes a memory issue when running video analysis using a model with an identity head, as reported in #2771.

**Improvements**

- `PadOutputs`: instead of needing to set `"key": 0` for outputs we don't want to pad (such as `identity_heatmap`), check if each output is present in the `max_individuals` dict to see if padding is needed.
- `PredictKeypointIdentities`: Models with identity heads would output the identity heatmaps, which leads to memory issues when analyzing large videos. A `keep_id_maps=False` parameter is added, only keep identity heatmaps in the output when explicitly requested. 
- `assign_identity`: use the pose confidence to weight identity scores when assigning identity, as done in the [`TensorFlow` implementation](https://github.com/DeepLabCut/DeepLabCut/blob/e98edcfe34ae13570c813a33aa410bea8369e58f/deeplabcut/pose_estimation_tensorflow/lib/inferenceutils.py#L150)

**Bug fixes**

- `make_pytorch_test_config` when there are unique bodyparts for a project. The unique bodyparts need to be included in `all_joint_names` and `all_joints`, otherwise video analysis fails.

### DIFF
--- a/deeplabcut/pose_estimation_pytorch/data/postprocessor.py
+++ b/deeplabcut/pose_estimation_pytorch/data/postprocessor.py
@@ -67,7 +67,6 @@ def build_bottom_up_postprocessor(
         keys_to_rescale.append("unique_bodyparts")
 
     if with_identity:
-        # TODO: do we really want to return the heatmaps?
         keys_to_concatenate["identity_heatmap"] = ("identity", "heatmap")
         empty_shapes["identity_heatmap"] = (1, 1, max_individuals)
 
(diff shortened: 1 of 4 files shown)


deeplabcut/pose_estimation_pytorch/data/postprocessor.py
```diff
@@ -404,39 +404,52 @@ def __call__(
 class PredictKeypointIdentities(Postprocessor):
     """Assigns predicted identities to keypoints
 
+    The identity maps have shape (h, w, num_ids).
+
     Attributes:
-        identity_key:
-        identity_map_key: shape (h, w, num_ids)
-        pose_key:
+        identity_key: The key with which to add predicted identities in the predictions
```
maximpavliv:
```suggestion
        identity_key: The key with which to add predicted identities in the predictions dict
```

maximpavliv (approved):

Issue #2784, opened 2024-11-13
How can users specify the initial weights of the network in DeepLabCut 3.0?（pytorch engine）


tagodong:
I tried using a method similar to the one in DeepLabCut 2.x, but the network didn't work. After reviewing the explanation document for the pytorch_config.yaml file (https://deeplabcut.github.io/DeepLabCut/docs/pytorch/pytorch_config.html#dlc3-pytorch-config), I still couldn't find any relevant information on this feature. Does this operation not yet have support in the current version?

tagodong:
I also tried
`deeplabcut.pose_estimation_pytorch.apis.train.train_network(config_path, shuffle=3, gputouse=3, snapshot_path='RootPath/snapshot-4500.pt')`

I found that the iteration count did indeed continue from the previous checkpoint. However, the results show a significant difference in the loss, even though the dataset and split are the same.
Before retrainning, the test loss is 1.089
After retrainning, the test loss is 1.221(Only reached the level of 170 epochs of training with the initial weights.)
Why is this happening?

```
Before retraining：
2024-11-13 09:28:51 Training for epoch 4500 done, starting evaluation
2024-11-13 09:28:54 Epoch 4500 performance:
2024-11-13 09:28:54 metrics/test.rmse:  1.089
2024-11-13 09:28:54 metrics/test.rmse_pcutoff:1.089
2024-11-13 09:28:54 metrics/test.mAP:   60.922
2024-11-13 09:28:54 metrics/test.mAR:   74.576
2024-11-13 09:28:54 metrics/test.rmse_detections:1.089
2024-11-13 09:28:54 metrics/test.rmse_detections_pcutoff:1.089
2024-11-13 09:28:54 Epoch 4500/20000 (lr=1e-06), train loss 0.00039, valid loss 0.00023
2024-11-13 09:29:09 Epoch 4501/20000 (lr=1e-06), train loss 0.00037
2024-11-13 09:29:24 Epoch 4502/20000 (lr=1e-06), train loss 0.00037
2024-11-13 09:29:38 Epoch 4503/20000 (lr=1e-06), train loss 0.00039
2024-11-13 09:29:53 Epoch 4504/20000 (lr=1e-06), train loss 0.00039
2024-11-13 09:30:08 Epoch 4505/20000 (lr=1e-06), train loss 0.00036
2024-11-13 09:30:22 Epoch 4506/20000 (lr=1e-06), train loss 0.00037
[...]

PR #2788, opened 2024-11-14
Load scheduler state when resuming training

n-poulsen:
This pull request addresses learning rate schedulers when resuming training. Currently, the schedulers are simply built from the config file. So when continuing to train from an existing snapshot, the scheduler will restart at epoch 0, and the learning rate won't be adapted (as mentioned in #2784).

In this pull request, the code is updated to:

- save the scheduler state dicts in snapshots
- when resuming training, try to load the state dict for the scheduler
  - if successful, set the optimizers learning rate to the last learning rate from the scheduler
- a `load_scheduler_state_dict` key is added to the runner configuration
  - the default is `True`: loading a snapshot with a saved scheduler to continue training will load the scheduler state dict
  - however, users might edit the scheduler's parameters and want an updated learning rate to continue training
  - in that case, they need to set `load_scheduler_state_dict: false` so the state dict doesn't overwrite their edited parameters
  - the `self.starting_epoch` value is used to set the `last_epoch` in the scheduler; so the way the learning rate will be schedule matches the epochs printed in the logs

Tests were added to validate that the expected learning rates are applied.


deeplabcut/pose_estimation_pytorch/runners/train.py
```diff
@@ -89,27 +99,34 @@ def __init__(
         self._print_valid_loss = True
 
         if self.snapshot_path is not None and self.snapshot_path != "":
```
maximpavliv:
```suggestion
        if self.snapshot_path:
```

deeplabcut/pose_estimation_pytorch/runners/train.py
```diff
@@ -45,33 +45,43 @@ class TrainingRunner(Runner, Generic[ModelType], metaclass=ABCMeta):
     def __init__(
```
maximpavliv, line 68: I think the docstring of the class is wrong (it is the same as the base Runner's one)

The InferenceRunner's docstring should probably be updated as well

maximpavliv (approved):

PR #2790, opened 2024-11-18
DeepLabCut 3.0 - Video Analysis with `use_shelve=True`

n-poulsen:
Implements shelving data during video analysis for the PyTorch engine. 

**DeepLabCut 3.0 new API features - `analyze_videos`**

- addition of the `use_shelve` 
- addition of the `robust_nframes` parameter for the `analyze_videos`

**Changes to the code**

- implementation of a `ShelfReader` and a `ShelfWriter` in `deeplabcut.pose_estimation_pytorch.runners.shelving`
- as the `ShelfWriter` now creates the `_full.pickle` file on-the-fly, `assign_identity` is called in a new `AssignIndividualIdentities` post-processor instead of in `video_inference` (which is an improvement on the previous implementation)

**Notes**

When reading from the `ShelfReader`, the order of keys returned does not necessarily match the order in which the keys were inserted. This could cause the predictions for frames to be shuffled when iterating over the frame keys. To solve this issue, the `nframes` parameter is now used to iterate over the number of frames in the video (in `_generate_assemblies_file` and in `convert_detections2tracklets`), and a "key_str_width" parameter is added to the metadata, storing the `str_width` parameter so that frame keys can be recreated from the frame index.


deeplabcut/pose_estimation_pytorch/apis/analyze_videos.py
```diff
@@ -233,11 +251,13 @@ def analyze_videos(
...
-            Note that the same cropping parameters will then be used for all videos.
-            If different video crops are desired, run ``analyze_videos`` on individual
-            videos with the corresponding cropping coordinates.
+        cropping: List of cropping coordinates as [x1, x2, y1, y2]. Note that the same
+            cropping parameters will then be used for all videos. If different video
+            crops are desired, run ``analyze_videos`` on individual videos with the
+            corresponding cropping coordinates.
+        save_as_df: Cannot be used when `use_shelve` is True. Saves the video
+            predictions (before tracking results) to an H5 file containing a pandas
+            DataFrame.
```
maximpavliv: Maybe we should also say that if `save_as_csv==True` it also saves the predictions? (here or in `save_as_csv`  argument description)

maximpavliv (approved):
Great!! 😃

PR #2791, opened 2024-11-20
Plot detector's bboxes if Top-Down architecture

maximpavliv:
This PR introduces the plotting of the detector's bounding boxes, if the architecture is Top-Down.

In particular, the bounding boxes are:
- Saved in the *_full.pickle file during video analysis
- Plotted by `evaluate_network()` when passing argument `plotting=True`
- Plotted by `create_video_with_all_detections()`, unless passing argument `plot_bboxes=False`
- Plotted by `create_labeled_video()`, unless passing argument `plot_bboxes=False`
- Plotted by `video_inference_superanimal() `, unless passing argument `plot_bboxes=False`
- Plotted by `visualize_predictions() `, unless passing argument `plot_bboxes=False`


deeplabcut/pose_estimation_pytorch/modelzoo/inference.py
```diff
@@ -181,6 +187,9 @@ def _video_inference_superanimal(
             bbox=bbox,
             cmap=colormap,
             output_path=str(output_video),
+            plot_bboxes=plot_bboxes,
+            bboxes_list=bboxes_list,
+            bboxes_pcutoff=bboxes_pcutoff,
```
maximpavliv, line 187: Hey, can we do something about this method's import? This `_create_labeled_video()` is defined in another module and is supposed to be private, so not imported here... I haven't found what would be the best way to resolve this

deeplabcut/utils/make_labeled_video.py
```diff
@@ -152,12 +156,36 @@ def CreateVideo(
...
+                    x, y = bbox[0], bbox[1]
+                    x += x1
+                    y += y1
+                    w, h = bbox[2], bbox[3]
+                    confidence = bbox_scores[i]
+                    if confidence < bboxes_pcutoff:
+                        continue
+                    rect_coords = rectangle_perimeter(start=(y, x), extent=(h, w))
+
+                    set_color(
```
maximpavliv, line 185: I was wondering why we are using two different libraries (`pyplot` or `skimage`) to mark images? These libraries are used to do basically the same thing, but the produced results look different (for example the added markers shapes look different, the line widths of rectangles are not the same...).
Pyplot seems to offer richer features (we can do dashed lines for examples, different lines widths, etc), so would be great if we could switch completely to it...

Do we need both because `skimage` works faster on videos (fastmode) than `pyplot`? But why do we keep pyplot then? Is it to have different labels (+, x, .) for network evaluation for example?

deeplabcut/utils/auxfun_multianimal.py
```diff
@@ -210,36 +186,33 @@ def SaveFullMultiAnimalData(data, metadata, dataname, suffix="_full"):
     return data_path, metadata_path
 
 
-def LoadFullMultiAnimalData(dataname):
-    """Save predicted data as h5 file and metadata as pickle file; created by predict_videos.py"""
-    data_file = dataname.split(".h5")[0] + "_full.pickle"
+def load_multianimal_full_meta_data(
```
maximpavliv: I have the feeling that we have multiple methods doing the same thing.
For loading the data saved in *_full.pickle and *_meta.pickle, we have this method, but there is also one in _deeplabcut/utils/auxiliaryfunctions.py_ called `load_video_metadata()` (and I added also `load_video_full_data()`)..
Should we unify these, to have them defined only once?

deeplabcut/utils/make_labeled_video.py
```diff
@@ -1168,15 +1266,42 @@ def create_video_with_all_detections(
...
+                                continue
+                            rect_coords = rectangle_perimeter(start=(y, x), extent=(h, w))
+
+                            set_color(
+                                frame,
+                                rect_coords,
+                                bboxes_color,
+                            )
+
+                    # Draw detected bodyparts
```
maximpavliv: We have this plotting of labels (and bboxes) on frames at least 4 times:
- In _deeplabcut/utils/make_labeled_video.py_:
    - in `create_video()`
    - in `create_video_slow()`
    - in `create_video_with_all_detections()`
- In _deeplabcut/utils/visualization.py_:
    - in `make_multianimal_labeled_image()`
    
Even if we need 1 version based on Pyplot and another based on skimage, can we reduce the number of these functions to 2? And ideally to one?

deeplabcut/pose_estimation_pytorch/modelzoo/inference.py
n-poulsen: I agree that it's not great to have this. There are different ways to do this

- add `create_labeled_video` method which calls the private method
- rename the method and add `_create_labeled_video = create_labeled_video` for backwards compatibility (note: do we need to? a private method should not be used externally)
- call `CreateVideo` instead (not good; there would likely be code duplication)

n-poulsen: Agreed that this should be fixed. I think we could either:

- rename the method to `make_video` or something like that (as the method was private, do we need to add `_ create_labeled_video = create_labeled_video`? normally not, but I think it might be best to do it anyways)
- add a `make_video ` wrapper around `_create_labeled_video` (not great)
- call `CreateVideo` instead (not great either; likely code duplication)

deeplabcut/utils/visualization.py
```diff
@@ -105,21 +114,62 @@ def make_labeled_image(
...
+    Plots groundtruth labels and predictions onto the matplotlib's axes, with the specified graphical parameters.
+
+    Args:
+        frame: image
+        coords_truth: groundtruth labels
+        coords_pred: predictions
+        probs_pred: prediction probabilities
+        colors: colors for poses
+        dotsize: size of dot
+        alphavalue:
```
n-poulsen: Missing docstring for `alphavalue` (it's the transparency for the keypoints)

deeplabcut/utils/visualization.py
```diff
@@ -394,6 +444,8 @@ def plot_evaluation_results(
     dot_size: int = 12,
     alpha_value: float = 0.7,
     p_cutoff: float = 0.6,
+    bounding_boxes: dict = {},
```
n-poulsen: A `dict` shouldn't be used as a default argument as it's mutable. Should set:

```python
def make_multianimal_labeled_image(
    ...
    bounding_boxes: dict | None = None,
    ...
) -> plt.Axes:
    if bounding_boxes is None:
        bounding_boxes = {}
```

n-poulsen (changes requested):
Some minor changes (mostly the mutable default argument), otherwise looks good to me!

deeplabcut/utils/visualization.py
maximpavliv: Right!

deeplabcut/pose_estimation_pytorch/modelzoo/inference.py
maximpavliv: To be sure I understand you right, the first option is to rename `_create_labeled_video()` to `make_video()` [for example] ?
And then you were wondering if we should add `_create_labeled_video` = `make_video` (that is what you meant, right)? I would say that it is not needed

n-poulsen: Yes that's what I meant - by not needed, do you mean because `_create_labeled_video` is private and thus nobody should be calling it from outside that file?

I agree that in principle that should be the case, however I think there's a chance other people were using it as it's a useful method that shouldn't have been made private, and adding this mapping would ensure outside scripts don't break

maximpavliv: Ok, yes that was my reasoning, but ok I will add the link then!

deeplabcut/pose_estimation_pytorch/apis/evaluate.py
```diff
@@ -250,13 +252,25 @@ def visualize_predictions(
             visible_pred.append(visible_points)
         visible_pred = np.stack(visible_pred)  # Shape: [N, num_visible_joints, 3]
 
+        if plot_bboxes:
+            bboxes = predictions[image_path].get("bboxes", None)
+            bbox_scores = predictions[image_path].get("bbox_scores", None)
+            bounding_boxes = (
+                (bboxes, bbox_scores)
+                if bbox_scores is not None and bbox_scores is not None
```
n-poulsen:
```suggestion
                if bboxes is not None and bbox_scores is not None
```

n-poulsen (changes requested):
Small typo to fix but otherwise LGTM!

n-poulsen (approved):

PR #2793, opened 2024-11-21
Fix: correct the early return error when save_epochs=1 and optimize the clean up way

xiu-cs:
- Fixed an issue causing an incorrect early return when save_epochs=1.
- clean up old snapshots if the number of existing snapshots exceeds the maximum allowed (executed at the end of the process).

Special thanks to @n-poulsen  for your invaluable guidance! 🙏

### DIFF
--- a/deeplabcut/pose_estimation_pytorch/runners/snapshots.py
+++ b/deeplabcut/pose_estimation_pytorch/runners/snapshots.py
@@ -130,17 +130,15 @@ class TorchSnapshotManager:
                     current_best.path.rename(new_name)
                 else:
                     current_best.path.unlink(missing_ok=False)
-        else:
+        elif last or epoch % self.save_epochs == 0:
             # Save regular snapshot if needed
-            should_save = last or epoch % self.save_epochs == 0
-            if should_save:
-                save_path = self.snapshot_path(epoch=epoch)
-                parsed_state_dict = {
-                    k: v
-                    for k, v in state_dict.items()
-                    if self.save_optimizer_state or k != "optimizer"
-                }
-                torch.save(parsed_state_dict, save_path)
+            save_path = self.snapshot_path(epoch=epoch)
+            parsed_state_dict = {
+                k: v
+                for k, v in state_dict.items()
+                if self.save_optimizer_state or k != "optimizer"
+            }
+            torch.save(parsed_state_dict, save_path)
 
         # Clean up old snapshots if needed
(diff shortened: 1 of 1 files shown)


deeplabcut/pose_estimation_pytorch/runners/snapshots.py
```diff
@@ -113,41 +113,43 @@ def update(self, epoch: int, state_dict: dict, last: bool = False) -> None:
         ):
             current_best = self.best()
             self._best_metric = metrics[self._key]
+            
+            # Save the new best model
             save_path = self.snapshot_path(epoch, best=True)
             parsed_state_dict = {
-                k: v
+                k: v
```
n-poulsen: You should run [black](https://github.com/psf/black) on the changes you made (with `black deeplabcut/pose_estimation_pytorch/runners/snapshots.py`) to lint the code (so there aren't these extra spaces)

deeplabcut/pose_estimation_pytorch/runners/snapshots.py
```diff
@@ -113,41 +113,43 @@ def update(self, epoch: int, state_dict: dict, last: bool = False) -> None:
...
+            # Save regular snapshot if needed
+            should_save = last or epoch % self.save_epochs == 0
+            if should_save:
+                save_path = self.snapshot_path(epoch=epoch)
+                parsed_state_dict = {
+                    k: v
+                    for k, v in state_dict.items()
+                    if self.save_optimizer_state or k != "optimizer"
+                }
+                torch.save(parsed_state_dict, save_path)
```
n-poulsen: I think we can be more concise here:

```suggestion
        elif last or epoch % self.save_epochs == 0:
            # Save regular snapshot if needed
            save_path = self.snapshot_path(epoch=epoch)
            parsed_state_dict = {
                k: v
                for k, v in state_dict.items()
                if self.save_optimizer_state or k != "optimizer"
            }
            torch.save(parsed_state_dict, save_path)
```

n-poulsen (changes requested):
Awesome thank you @xiu-cs! Just requested a few minor changes and some linting :)

deeplabcut/pose_estimation_pytorch/runners/snapshots.py
xiu-cs: Thanks! Next time I will run black before I submit the changes.

image

n-poulsen (approved):
Looks good to me, thanks @xiu-cs!

PR #2800, opened 2024-12-03
DLC 3.0 - Function to export PyTorch models

n-poulsen:
Implements the `export_model` API method for models trained with the PyTorch engine. Tests are added for the different features of the `export_model` method.

The following modifications are made to implement the feature:

- The `DLCLoader` lazy-loads the DataFrames and train/test splits
- We can instantiate the `DLCLoader` with a project configuration directly (a `dict`) instead of having to pass a filepath
  - This makes it possible to change the project `iteration` dynamically when exporting models

### DIFF
--- a/deeplabcut/compat.py
+++ b/deeplabcut/compat.py
@@ -53,7 +53,7 @@ def get_available_aug_methods(engine: Engine) -> tuple[str, ...]:
     if engine == Engine.TF:
         return "imgaug", "default", "deterministic", "scalecrop", "tensorpack"
     elif engine == Engine.PYTORCH:
-        return ("albumentations", )
+        return ("albumentations",)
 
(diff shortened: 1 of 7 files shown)


maximpavliv (approved):
LGTM!

hg-aus-g:
Does the export_model() function works with pytorch models? Unfortunately I got only one .pt file. But I need .pb file for use witch dlc-live-gui!

Thanks!

maximpavliv:
Hello @hg-aus-g ,

_.pb_ is the format used by Tensorflow, Pytorch uses the _.pt_ format to save models, so this is perfectly normal.
 
The thing is that pytorch support has not been merged on the DeepLabCut-live side yet, although it is almost ready (see [PR](https://github.com/DeepLabCut/DeepLabCut-live/pull/121)). It should be merged very soon :)

Issue #2769, opened 2024-10-28
Changing batch size causes an error: `KeypointAwareCrop`


vonaviv:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Bug description

I use DLC with PyTorch engine and when I set the batch size for training in `pytorch_config.yaml` to anything except of 1 I get the following error message:

```
Loading DLC 3.0.0rc5...
DLC loaded in light mode; you cannot use any GUI (labeling, relabeling and standalone GUI)
Traceback (most recent call last):
  File "/user/vladyslav.ivanov01/u11170/dlc/ExR_18cam-smg-2024-10-25/scripts/train.py", line 13, in <module>
    deeplabcut.train_network(config_path, max_snapshots_to_keep=20, shuffle=shuffle)
  File "/usr/local/lib/python3.10/dist-packages/deeplabcut/compat.py", line 245, in train_network
    return train_network(
  File "/usr/local/lib/python3.10/dist-packages/deeplabcut/pose_estimation_pytorch/apis/train.py", line 338, in train_network
    train(
  File "/usr/local/lib/python3.10/dist-packages/deeplabcut/pose_estimation_pytorch/apis/train.py", line 186, in train
    runner.fit(
  File "/usr/local/lib/python3.10/dist-packages/deeplabcut/pose_estimation_pytorch/runners/train.py", line 166, in fit
    train_loss = self._epoch(
  File "/usr/local/lib/python3.10/dist-packages/deeplabcut/pose_estimation_pytorch/runners/train.py", line 216, in _epoch
    for i, batch in enumerate(loader):
  File "/usr/local/lib/python3.10/dist-packages/torch/utils/data/dataloader.py", line 701, in __next__
    data = self._next_data()
  File "/usr/local/lib/python3.10/dist-packages/torch/utils/data/dataloader.py", line 757, in _next_data
    data = self._dataset_fetcher.fetch(index)  # may raise StopIteration
  File "/usr/local/lib/python3.10/dist-packages/torch/utils/data/_utils/fetch.py", line 52, in fetch
    data = [self.dataset[idx] for idx in possibly_batched_index]
  File "/usr/local/lib/python3.10/dist-packages/torch/utils/data/_utils/fetch.py", line 52, in <listcomp>
    data = [self.dataset[idx] for idx in possibly_batched_index]
[...]

PR #2801, opened 2024-12-03
`KeypointAwareCrop`: uniform sampling when no keypoints are present

n-poulsen:
Addresses issue #2769. When all keypoints labeled in an image are augmented outside of the image (e.g. they are close to the edge and a rotation is applied), no keypoints are left. Then, density-based sampling of the crop center fails as there are no keypoints to sample from.

Address a few other typos when it comes to image augmentation:
- When creating the `Affine` transform, the `translation` parameter must be given as `(-translation, translation)`
- Keypoints should be padded with `0`s (not visible) and not `-1` (not defined)


MMathisLab (commented):
Good catch!

PR #2805, opened 2024-12-04
Fix version check timeout

maximpavliv:
In #2782, the `call_with_timeout()` method was not timing out correctly.

This PR fixes this method so that the latestest version retrieval from the web times out if caught in the firewall.

I also added some tests to be sure `call_with_timeout()`  works correctly this time.


n-poulsen (approved):
LGTM! Thanks @maximpavliv

PR #2816, opened 2024-12-16
Implement missing arguments for PyTorch models 

n-poulsen:
This pull request implements features for API methods that were not yet available for PyTorch models.

 ##  `deeplabcut.analyze_images`

The `analyze_time_lapse_frames` method is added to the PyTorch API, such that it can be called with PyTorch models in the same way as it was for TensorFlow models. Below the hood, it calls the `analyze_images` method.

The `analyze_images` (which can only be called for PyTorch models - `analyze_time_lapse_frames` should still be used for TensorFlow models) is added to the `compat.py` so that it can be called as:

```python
import deeplabcut

deeplabcut.analyze_images(
    config="path/to/my/project/config.yaml",
    images="/my/image/dir",
    shuffle=12,
)
```

New features are also added to the `analyze_images` method as well:
- Predictions can directly be plot onto images through the `plotting` parameter. Predictions are plot in the same format as when called with `evaluate_network`, the only difference being there is no ground truth to plot. The plotting parameters are taken from the project configuration file, and `pcutoff`, `bbox_pcutoff` and `plot_skeleton` allow to dynamically edit some of the parameters controlling plotting. 
- The predictions are now stored as a pandas `DataFrame`, as with other outputs in DeepLabCut. The predictions can also be saved in `CSV` files with the `save_as_csv` parameter.

## `deeplabcut.analyze_videos`

The `in_random_order` and `n_tracks` parameters are implemented for video analysis with PyTorch models.

Dynamic cropping (through the `dynamic` argument) is now available for `analyze_videos ` with PyTorch models. This function is implemented through a `DynamicCropper` cropper class, which is fully tested.

The `greedy`, `calibrate` and `window_size` arguments still need to be implemented. They are only valid for PAF-models, and require that more information is added to the `_full.pickle` file when running video analysis. This will be done in a subsequent pull request.

## `deeplabcut.train_network`

The `keepdeconvweights` argument is implemented, which allows to continue. Two additional arguments are added to the method: `epochs` and `save_epochs`. These arguments could already be passed, but making them explicit makes it clearer for users how they can be used.


deeplabcut/compat.py
```diff
@@ -120,6 +122,20 @@ def train_network(
...
+
+    epochs: optional, default=None
+        Only for the PyTorch engine (equivalent to the `maxiters` parameter for the
+        TensorFlow engine). The maximum number of epochs to train the model for. If
+        None, the value will be read from the `pytorch_config.yaml` file. An epoch is a
+        single pass through the training dataset, which means your model has seen each
+        training image exactly once. So if you have 64 training images for your network,
+        an epoch is 64 iterations with batch size 1 (or 32 iterations with batch size 2,
+        16 with batch size 4, etc.).
+
```
maximpavliv:
```suggestion
    epochs: optional, default=None
        Only for the PyTorch engine (equivalent to the `maxiters` parameter for the
        TensorFlow engine). The maximum number of epochs to train the model for. If
        None, the value will be read from the `pytorch_config.yaml` file. An epoch is a
        single pass through the training dataset, which means your model has seen each
        training image exactly once. So if you have 64 training images for your network,
        an epoch is 64 iterations with batch size 1 (or 32 iterations with batch size 2,
        16 with batch size 4, etc.).

    save_epochs: optional, default=None
        Only for the PyTorch engine (equivalent to the `saveiters` parameter for the
        TensorFlow engine). The number of epochs between each snapshot save. If
        None, the value will be read from the `pytorch_config.yaml` file.
```

deeplabcut/compat.py
```diff
@@ -944,26 +961,201 @@ def create_tracking_dataset(
...
+        The cutoff score when plotting bounding box predictions. Must be
+        None or in (0, 1). If None, it is read from the project configuration file.
+
+    plot_skeleton: bool, default=False
+        If a skeleton is defined in the project's config.yaml, whether
+        to plot the skeleton connecting the predicted bodyparts on the images.
+
+    modelprefix: str, optional
+        Directory containing the deeplabcut models to use when running image analysis.
+        By default, the models are assumed to exist in the project folder.
```
maximpavliv:
```suggestion
    save_as_csv: bool, optional
        Saves the predictions in a .csv file. The default is ``False``; if provided it
        must be either ``True`` or ``False``.

    modelprefix: str, optional
        Directory containing the deeplabcut models to use when running image analysis.
        By default, the models are assumed to exist in the project folder.

    plotting: bool, str, default=False
        Plots the predictions made by the model on the analyzed images. Results will be
        stored in a folder named `LabeledImages_{scorer}`, where scorer is the name
        of the model used to analyze the images. This folder will be in the same
        directory as the file containing the predictions (either the given `destfolder`,
        or the folder containing the first image to analyze).

        If provided it must be either ``True``, ``False``, ``"bodypart"``, or
        ``"individual"``. Setting to ``True`` defaults as ``"bodypart"`` for
        multi-animal projects. If a detector is used, the predicted bounding boxes
        will also be plotted.

    pcutoff: float, optional, default=None
        The cutoff score when plotting pose predictions. Must be None or in
        (0, 1). If None, the pcutoff is read from the project configuration file.

    bbox_pcutoff: float, optional, default=None
        The cutoff score when plotting bounding box predictions. Must be
        None or in (0, 1). If None, it is read from the project configuration file.

    plot_skeleton: bool, default=False
        If a skeleton is defined in the project's config.yaml, whether
        to plot the skeleton connecting the predicted bodyparts on the images.
```

deeplabcut/compat.py
```diff
@@ -1020,6 +1230,18 @@ def analyze_time_lapse_frames(
             save_as_csv=save_as_csv,
             modelprefix=modelprefix,
         )
+    elif engine == Engine.PYTORCH:
+        from deeplabcut.pose_estimation_pytorch import analyze_images
+        analyze_images(
```
maximpavliv:
```suggestion
        return analyze_images(
```

deeplabcut/compat.py
```diff
@@ -944,26 +961,201 @@ def create_tracking_dataset(
...
+    Output: The labels are stored as MultiIndex Pandas Array, which contains the name
+    of the network, body part name, (x, y) label position in pixels, and the likelihood
+    for each frame per body part. These arrays are stored in an efficient Hierarchical
+    Data Format (HDF) in the same directory, where the video is stored. However, if the
+    flag save_as_csv is set to True, the data can also be exported in comma-separated
+    values format (.csv), which in turn can be imported in many programs, such as
+    MATLAB, R, Prism, etc.
 
     This function is only implemented for tensorflow models/shuffles, and will throw
     an error if called with a PyTorch shuffle.
```
maximpavliv:
```suggestion
```

deeplabcut/pose_estimation_pytorch/apis/convert_detections_to_tracklets.py
```diff
@@ -42,9 +42,9 @@ def convert_detections2tracklets(
     ignore_bodyparts: Optional[List[str]] = None,
     inferencecfg: Optional[dict] = None,
     modelprefix="",
-    greedy=False,  # TODO: Unused, remove
-    calibrate=False,  # TODO: Unused, remove
-    window_size=0,  # TODO: Unused, remove
+    greedy: bool = False,  # TODO: Unused, remove
+    calibrate: bool = False,  # TODO: Unused, remove
+    window_size: int = 0,  # TODO: Unused, remove
```
maximpavliv: These arguments are planned to be added, right? Then maybe we should update the comments, to say that it isn't implemented yet, but will in the future

docs/pytorch/user_guide.md
```diff
@@ -76,13 +76,13 @@ parameters are not valid for the DLC 3.0 PyTorch API.
...
-| `evaluate_network`             |     🟠      | `rescale`                                                                                           |                                                     |
+| `evaluate_network`             |     🟢      |                                                                                                     |                                                     |
 | `return_evaluate_network_data` |     🔴      |                                                                                                     | `TFGPUinference`, `allow_growth`                    |
-| `analyze_videos`               |     🟠      | `in_random_order`, `dynamic`, `n_tracks`, `calibrate`                                               |                                                     |
+| `analyze_videos`               |     🟠      | `calibrate`                                                                                         |                                                     |
```
maximpavliv:
```suggestion
| `analyze_videos`               |     🟠      | `greedy`, `calibrate`, `window_size`                                                                |                                                     |
```

maximpavliv (changes requested):

deeplabcut/pose_estimation_pytorch/apis/convert_detections_to_tracklets.py
n-poulsen: Yes those will be added in a subsequent pull request.

```suggestion
    greedy: bool = False,  # TODO(niels): implement greedy assembly during video analysis
    calibrate: bool = False,  # TODO(niels): implement assembly calibration during video analysis
    window_size: int = 0,  # TODO(niels): implement window size selection for assembly during video analysis
```

maximpavliv (approved):
LGTM!

Issue #2725, opened 2024-08-27
pose_estimation_pytorch torch.load warning


dlhagger:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Bug description

When running analyze_videos on rc4, a future warning that seems reasonable to update is presenting:

FutureWarning: You are using `torch.load` with `weights_only=False` (the current default value), which uses the default pickle module implicitly. It is possible to construct malicious pickle data which will execute arbitrary code during unpickling (See https://github.com/pytorch/pytorch/blob/main/SECURITY.md#untrusted-models for more details). In a future release, the default value for `weights_only` will be flipped to `True`. This limits the functions that could be executed during unpickling. Arbitrary objects will no longer be allowed to be loaded via this mode unless they are explicitly allowlisted by the user via `torch.serialization.add_safe_globals`. We recommend you start setting `weights_only=True` for any use case where you don't have full control of the loaded file. Please open an issue on GitHub for any issues related to this experimental feature.
  snapshot = torch.load(snapshot_path, map_location=device)

Is it possible to add the files to the allowlist by default or load with weights_only in the future? Many of us are on security managed compute by our institutes or corporations, so might be worth while to address before a public release?

Thanks!

David

### Operating System

operating system

MacOS 14.6.1

### DeepLabCut version

dlc version

pytorch rc4

### DeepLabCut mode

single animal

### Device type

gpu

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/DeepLabCut/DeepLabCut/blob/master/CODE_OF_CONDUCT.md)

n-poulsen:
Hi @dlhagger! I think this is a good point and I'll address it before the full release (this warning was [just added in PyTorch 2.4](https://github.com/pytorch/pytorch/releases/tag/v2.4.0) :) so we'll make the change before PyTorch changes the default).

Generally, there should be no security issue as long as you only load checkpoints that you created for your models (and not load a random checkpoint found on the internet).

Meanwhile, in `torch>=2.4`, if you want to ensure that the weights are loaded safely, you can set the environment variable `TORCH_FORCE_WEIGHTS_ONLY_LOAD=1`, which will ensure that `weights_only` will be used, no matter what's written in the code. In a Python script, you can do so with the following code:

```python
import os

os.environ["TORCH_FORCE_WEIGHTS_ONLY_LOAD"] = "1"
```

You can also set environment variables in your shell directly.

[...]

PR #2818, opened 2024-12-18
Set `weights_only=True` as a default when loading snapshot weights.

n-poulsen:
Addresses #2725. The `torch.load` method is now called with `weights_only=True` as a default. If loading snapshot weights fails, users can set a `load_weights_only: False` key in their `pytorch_config.yaml` file to set the `weights_only` parameter to False. This is explained in the error message when loading weights fails. This should only be done when users trust the content of the snapshot they're loading, as explained in [the `torch.load` docs](https://pytorch.org/docs/stable/generated/torch.load.html).

Tests were added to ensure `torch.load` was called with the correct parameters.

To set `weights_only=False`, users can edit their `pytorch_config.yaml` files as:

```yaml
detector:
    ...
    runner:
        load_weights_only: false  # to load detector weights with `weights_only=False`

runner:
    load_weights_only: false  # to load pose model weights with `weights_only=False`
```

### DIFF
--- a/tests/pose_estimation_pytorch/modelzoo/test_load_superanimal_models.py
+++ b/tests/pose_estimation_pytorch/modelzoo/test_load_superanimal_models.py
@@ -1,0 +1,31 @@
+#
+# DeepLabCut Toolbox (deeplabcut.org)
+# © A. & M.W. Mathis Labs
+# https://github.com/DeepLabCut/DeepLabCut
+#
+# Please see AUTHORS for contributors.
+# https://github.com/DeepLabCut/DeepLabCut/blob/main/AUTHORS
+#
+# Licensed under GNU Lesser General Public License v3.0
+#
+import dlclibrary
+import pytest
+import torch
+
+from deeplabcut.pose_estimation_pytorch.modelzoo import get_super_animal_snapshot_path
+
+
+@pytest.mark.skip(reason="require-models")
+def test_load_superanimal_models_weights_only():
+    super_animal_names = dlclibrary.get_available_datasets()
+    for super_animal in super_animal_names:
+        print(f"\nTesting {super_animal}")
+        for detector in dlclibrary.get_available_detectors(super_animal):
+            print(super_animal, detector)
+            path = get_super_animal_snapshot_path(super_animal, detector)
+            snapshot = torch.load(path, map_location="cpu", weights_only=True)
+
+        for pose_model in dlclibrary.get_available_models(super_animal):
+            print(super_animal, pose_model)
+            path = get_super_animal_snapshot_path(super_animal, pose_model)
+            snapshot = torch.load(path, map_location="cpu", weights_only=True)
--- a/deeplabcut/pose_estimation_pytorch/apis/utils.py
+++ b/deeplabcut/pose_estimation_pytorch/apis/utils.py
@@ -516,6 +516,9 @@ def get_inference_runners(
                 postprocessor=build_detector_postprocessor(
                     max_individuals=max_individuals,
                 ),
+                load_weights_only=model_config["detector"]["runner"].get(
+                    "load_weights_only", True,
+                ),
             )
 
     pose_runner = build_inference_runner(
@@ -526,6 +529,7 @@ def get_inference_runners(
         batch_size=batch_size,
         preprocessor=pose_preprocessor,
         postprocessor=pose_postprocessor,
+        load_weights_only=model_config["runner"].get("load_weights_only", True),
     )
     return pose_runner, detector_runner
 
@@ -575,6 +579,7 @@ def get_detector_inference_runner(
         batch_size=batch_size,
         preprocessor=preprocessor,
         postprocessor=postprocessor,
+        load_weights_only=det_cfg["runner"].get("load_weights_only", True),
     )
 
     if not isinstance(runner, DetectorInferenceRunner):
@@ -650,6 +655,7 @@ def get_pose_inference_runner(
         batch_size=batch_size,
         preprocessor=pose_preprocessor,
         postprocessor=pose_postprocessor,
+        load_weights_only=model_config["runner"].get("load_weights_only", True),
     )
     if not isinstance(runner, PoseInferenceRunner):
         raise RuntimeError(f"Failed to build PoseInferenceRunner for {model_config}")
--- a/deeplabcut/pose_estimation_pytorch/runners/base.py
+++ b/deeplabcut/pose_estimation_pytorch/runners/base.py
@@ -10,7 +10,6 @@
 #
 from __future__ import annotations
 
-import logging
 import pickle
 from abc import ABC
 from pathlib import Path
--- a/deeplabcut/pose_estimation_pytorch/runners/base.py
+++ b/deeplabcut/pose_estimation_pytorch/runners/base.py
@@ -19,7 +19,6 @@ from typing import Generic, TypeVar
 import torch
 import torch.nn as nn
 
-
 ModelType = TypeVar("ModelType", bound=nn.Module)
 
 
@@ -87,7 +86,9 @@ class Runner(ABC, Generic[ModelType]):
         """
         try:
             snapshot = torch.load(
-                snapshot_path, map_location=device, weights_only=weights_only,
+                snapshot_path,
+                map_location=device,
+                weights_only=weights_only,
             )
         except pickle.UnpicklingError as err:
             print(
--- a/deeplabcut/pose_estimation_pytorch/runners/base.py
+++ b/deeplabcut/pose_estimation_pytorch/runners/base.py
@@ -10,6 +10,8 @@
 #
 from __future__ import annotations
 
+import logging
+import pickle
 from abc import ABC
 from pathlib import Path
 from typing import Generic, TypeVar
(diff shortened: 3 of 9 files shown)


tests/pose_estimation_pytorch/modelzoo/test_load_superanimal_models.py
```diff
@@ -0,0 +1,31 @@
...
+# Licensed under GNU Lesser General Public License v3.0
+#
+import dlclibrary
+import pytest
+import torch
+
+from deeplabcut.pose_estimation_pytorch.modelzoo import get_super_animal_snapshot_path
+
+
+@pytest.mark.skip(reason="require-models")
```
maximpavliv, line 18: Is there a way to force running tests decorated like this? I tried looking that up but haven't found anything that could be used without modifying the code.
Otherwise what is the goal of tests that are never ran?

maximpavliv (approved):

tests/pose_estimation_pytorch/modelzoo/test_load_superanimal_models.py
n-poulsen: I ran these locally - but as it requires that all SuperAnimal models be downloaded locally, that would be too slow to run when the CI/CD pipeline is ran. You're correct that there's no CLI argument we can give to run this test. So the only way to run it again would be to remove the skip parameter.

The goal of this code was to test that the SuperAnimal models could all be loaded correctly with the `weights_only=True` option set. Now that it's be run once for all models it should be good (though I agree, it would be better to have a command-line option to be able to run it periodically). But `pytest` isn't great with that, so there isn't an easy way to do it.

MMathisLab:
this seems to break this fucntion now @n-poulsen 

```
superanimal_analyze_images(superanimal_name,
                           model_name,
                           detector_name= 'fasterrcnn_mobilenet_v3_large_fpn',
                           images = in_image_folder,
                           max_individuals = max_individuals,
                           out_folder = out_image_folder,
                           pose_threshold = 0.5,
                           bbox_threshold = 0.6,
                           plot_skeleton = True,
                           customized_model_config=customized_model_config,
                           customized_pose_checkpoint=customized_pose_checkpoint,
                           customized_detector_checkpoint=customized_detector_checkpoint)
```
output (on weights that worked yesterday!)

```
Failed to load the snapshot: /content/drive/MyDrive/.../snapshot-detector-best-070.pt.
If you trust the snapshot that you're trying to load, you can try calling `Runner.load_snapshot` with `weights_only=False`. See the message below for more information and warnings.
You can set the `weights_only` parameter in the model configuration (the content of the pytorch_config.yaml), as:

runner:
  load_weights_only: False


---------------------------------------------------------------------------
UnpicklingError                           Traceback (most recent call last)
[<ipython-input-15-12fba7293495>](https://localhost:8080/#) in <cell line: 1>()
----> 1 superanimal_analyze_images(superanimal_name,
      2                            model_name,
      3                            detector_name= 'fasterrcnn_mobilenet_v3_large_fpn',
      4                            images = in_image_folder,
      5                            max_individuals = max_individuals,

9 frames
[/usr/local/lib/python3.10/dist-packages/deeplabcut/pose_estimation_pytorch/apis/analyze_images.py](https://localhost:8080/#) in superanimal_analyze_images(superanimal_name, model_name, detector_name, images, max_individuals, out_folder, progress_bar, device, pose_threshold, bbox_threshold, plot_skeleton, customized_model_config, customized_pose_checkpoint, customized_detector_checkpoint)
    181         config["detector"]["model"]["box_score_thresh"] = bbox_threshold
    182 
--> 183     predictions = analyze_image_folder(
    184         model_cfg=config,
    185         images=images,

[/usr/local/lib/python3.10/dist-packages/deeplabcut/pose_estimation_pytorch/apis/analyze_images.py](https://localhost:8080/#) in analyze_image_folder(model_cfg, images, snapshot_path, detector_path, frame_type, device, max_individuals, progress_bar)
    435         device = resolve_device(model_cfg)
    436 
--> 437     pose_runner, detector_runner = get_inference_runners(
    438         model_config=model_cfg,
    439         snapshot_path=snapshot_path,

[/usr/local/lib/python3.10/dist-packages/deeplabcut/pose_estimation_pytorch/apis/utils.py](https://localhost:8080/#) in get_inference_runners(model_config, snapshot_path, max_individuals, num_bodyparts, num_unique_bodyparts, batch_size, device, with_identity, transform, detector_batch_size, detector_path, detector_transform, dynamic)
    521                 detector_config["pretrained"] = False
    522 
--> 523             detector_runner = build_inference_runner(
    524                 task=Task.DETECT,
    525                 model=DETECTORS.build(detector_config),

[/usr/local/lib/python3.10/dist-packages/deeplabcut/pose_estimation_pytorch/runners/inference.py](https://localhost:8080/#) in build_inference_runner(task, model, device, snapshot_path, batch_size, preprocessor, postprocessor, dynamic, load_weights_only)
    370                 f"detection. Please turn off dynamic cropping."
    371             )
--> 372         return DetectorInferenceRunner(**kwargs)
    373 
    374     if task != Task.BOTTOM_UP:

[/usr/local/lib/python3.10/dist-packages/deeplabcut/pose_estimation_pytorch/runners/inference.py](https://localhost:8080/#) in __init__(self, model, **kwargs)
    291             **kwargs: Inference runner kwargs.
    292         """
--> 293         super().__init__(model, **kwargs)
    294 
    295     def predict(self, inputs: torch.Tensor) -> list[dict[str, dict[str, np.ndarray]]]:

[/usr/local/lib/python3.10/dist-packages/deeplabcut/pose_estimation_pytorch/runners/inference.py](https://localhost:8080/#) in __init__(self, model, batch_size, device, snapshot_path, preprocessor, postprocessor, load_weights_only)
     68 
     69         if self.snapshot_path is not None and self.snapshot_path != "":
---> 70             self.load_snapshot(
     71                 self.snapshot_path,
     72                 self.device,

[/usr/local/lib/python3.10/dist-packages/deeplabcut/pose_estimation_pytorch/runners/base.py](https://localhost:8080/#) in load_snapshot(snapshot_path, device, model, weights_only)
     84             The content of the snapshot file.
     85         """
---> 86         snapshot = attempt_snapshot_load(snapshot_path, device, weights_only)
     87         model.load_state_dict(snapshot["model"])
     88         return snapshot

[/usr/local/lib/python3.10/dist-packages/deeplabcut/pose_estimation_pytorch/runners/base.py](https://localhost:8080/#) in attempt_snapshot_load(path, device, weights_only)
    121             "  load_weights_only: False\n```\n"
    122         )
--> 123         raise err
    124 
    125     return snapshot

[/usr/local/lib/python3.10/dist-packages/deeplabcut/pose_estimation_pytorch/runners/base.py](https://localhost:8080/#) in attempt_snapshot_load(path, device, weights_only)
    109     """
    110     try:
--> 111         snapshot = torch.load(path, map_location=device, weights_only=weights_only)
    112     except pickle.UnpicklingError as err:
    113         print(

[/usr/local/lib/python3.10/dist-packages/torch/serialization.py](https://localhost:8080/#) in load(f, map_location, pickle_module, weights_only, mmap, **pickle_load_args)
   1357                         )
   1358                     except pickle.UnpicklingError as e:
-> 1359                         raise pickle.UnpicklingError(_get_wo_message(str(e))) from None
   1360                 return _load(
   1361                     opened_zipfile,

UnpicklingError: Weights only load failed. This file can still be loaded, to do so you have two options, do those steps only if you trust the source of the checkpoint. 
	(1) Re-running `torch.load` with `weights_only` set to `False` will likely succeed, but it can result in arbitrary code execution. Do it only if you got the file from a trusted source.
	(2) Alternatively, to load with `weights_only=True` please check the recommended steps in the following error message.
	WeightsUnpickler error: Unsupported global: GLOBAL numpy.core.multiarray.scalar was not an allowed global by default. Please use `torch.serialization.add_safe_globals([scalar])` to allowlist this global if you trust this class/function.

Check the documentation of torch.load to learn more about types accepted by default with weights_only https://pytorch.org/docs/stable/generated/torch.load.html.
```

and if I edit the pytorch config, adding `load_weights_only: False` does not solve it.

MMathisLab:
@n-poulsen I can confirm if I roll back to this commit, it's working, so I think its this PR that is causing issues cc @maximpavliv !pip install --force-reinstall "git+https://github.com/DeepLabCut/DeepLabCut.git@198fc74#egg=deeplabcut[modelzoo]"

n-poulsen:
@MMathisLab Sorry about that - I should have made the error message more clear. This might happen on weights that were previously trained in DeepLabCut. The issue is with `numpy` float64 values being stored in the metadata for snapshots (some metric functions returned `numpy` floats instead of native python floats). The metric functions now all return native python floats, so this should not be an issue in the future.

If we don't want users who already have trained weights to have these issues, we could:
- Temporarily set the `load_weights_only` parameter to `true` as default. This is the current behavior with PyTorch if `weights_only` is not explicitly set to False
- Add a helper function to "fix" snapshots, converting all values to `python` floats instead of `numpy` floats

Do you see one of those two options as a good way to deal with this issue?

I was able to successfully load such weights by adding the `load_weights_only: false` parameter. Here it seems that it's the detector snapshot that's failing to load, so the `load_weights_only` key should be added under the `detector`'s runner in the `pytorch_config.yaml` file. I'll update the error message to make it clear that the `load_weights_only` key should be added for the detector or pose snapshot specifically

```yaml
...
detector:
  runner:
    load_weights_only: false
...
runner:
  load_weights_only: false
```

PR #2819, opened 2024-12-19
Add missing positional arguments to SuperAnimal Colab

maximpavliv:
The Superanimal colab was missing positional arguments, so I updated it


maximpavliv:
@n-poulsen `model_name` is a positional argument of `video_inference_superanimal()` and was missing in this Colab.
I am not sure if I added the best default one, can you double-check and confirm please?

n-poulsen (approved):
Thanks @maximpavliv! Looks good to me!

PR #2820, opened 2024-12-20
Fix check version timeout Version 2 (compatible with Windows and MacOS)

maximpavliv:
In my [previous attempt](https://github.com/DeepLabCut/DeepLabCut/pull/2805) to add a timeout to `deeplabcut.gui.window._check_for_updates()`, the tests were failing on Windows and MacOS due to a difference in the `multiprocessing` package workflow across these different OSes. This attempt has therefore [been reverted](https://github.com/DeepLabCut/DeepLabCut/pull/2807).

This Pull Requests proposes an addition of timeout to `_check_for_updates()` with tests that should be compatible with all 3 OS families (tested on Ubuntu 24.04 and Windows, @n-poulsen could you confirm this also works on MacOS please?).

### DIFF
--- a/deeplabcut/gui/window.py
+++ b/deeplabcut/gui/window.py
@@ -17,7 +17,6 @@ from pathlib import Path
 from typing import List
 from urllib.error import URLError
 import qdarkstyle
-import multiprocessing
 
 import deeplabcut
 from deeplabcut import auxiliaryfunctions, VERSION, compat
@@ -25,6 +24,7 @@ from deeplabcut.core.engine import Engine
 from deeplabcut.gui import BASE_DIR, components, utils
 from deeplabcut.gui.tabs import *
 from deeplabcut.gui.widgets import StreamReceiver, StreamWriter
+from deeplabcut.utils.multiprocessing import call_with_timeout
 from napari_deeplabcut import misc
 from PySide6.QtWidgets import (
     QMessageBox,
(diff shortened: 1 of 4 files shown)


maximpavliv:
For info:
The difference in `multiprocessing` between Linux and Windows or MacOS is how it starts new processes (by default):

- On Linux, the default method is `"fork"`
- On Windows and MacOS, the default method is `"spawn"`

This can be verified by doing
```
import multiprocessing
multiprocessing.get_start_method()
```
With `"fork"`, the child process inherits the memory space, file descriptors, and environment from the parent process. With `"spawn"`, a fresh Python interpreter is started, and only the necessary objects (like the target function) are passed to the child process.
Therefore, to successfully run the tests on all OSes, the methods passed to multiprocessing cannot be defined in a local scope, because they cannot be pickled if it is the case, which then is not compatible with the `"spawn"` method.
Another consequence is that, in practice, `deeplabcut` needs to be reimported in each child process if using the `"spawn"` method (both when running `test_call_with_timeout()` and when running `_check_for_updates()`), which might take some time. This is not the case if new processes are started using `"fork"` method (Linux).
 @n-poulsen do you see any potential issue with this, that I didn't think of?

MMathisLab (approved):

maximpavliv:
@n-poulsen I increased the timeout from 1s to 5s in version checks, what do you think about this?

Issue #2796, opened 2024-11-28
Unable to Create Labeled Video after Analyzation


alexgaudineeralexgaudineer:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Bug description

After Following steps from the Single Animal User Guide, I was able to Extract & Label Frames, train a network, and finally analyze a video. When trying to create a labeled video though, I get and error - "No unfiltered data file found in C:\dir for video and scorer".
This happens right after a successful output from the analyze function.

### Operating System

Win11

### DeepLabCut version

3.0.0rc6

### DeepLabCut mode

single animal

### Device type

gpu
NVIDIA-SMI 556.12                 Driver Version: 556.12         CUDA Version: 12.5 

### Steps To Reproduce

Followed Steps from Single animal User Guide in Project documentation,

Created a new project
Extracted Frames
Labeled Frames
Created Training Dataset
Trained Network
Analyzed Videos
Create videos

### Relevant log output

```shell
Analyzing videos with C:\conda_envs\Cat-004-Midas-2024-11-26\dlc-models-pytorch\iteration-0\Cat-004Nov26-trainset95shuffle1\train\snapshot-best-020.pt
C:\conda_envs\dlcdlc\lib\site-packages\deeplabcut\pose_estimation_pytorch\runners\base.py:80: FutureWarning: You are using `torch.load` with `weights_only=False` (the current default value), which uses the default pickle module implicitly. It is possible to construct malicious pickle data which will execute arbitrary code during unpickling (See https://github.com/pytorch/pytorch/blob/main/SECURITY.md#untrusted-models for more details). In a future release, the default value for `weights_only` will be flipped to `True`. This limits the functions that could be executed during unpickling. Arbitrary objects will no longer be allowed to be loaded via this mode unless they are explicitly allowlisted by the user via `torch.serialization.add_safe_globals`. We recommend you start setting `weights_only=True` for any use case where you don't have full control of the loaded file. Please open an issue on GitHub for any issues related to this experimental feature.
  snapshot = torch.load(snapshot_path, map_location=device)
Starting to analyze C:\conda_envs\Cat-004-Midas-2024-11-26\videos\camera-2-vid1_30fps.mp4
Video metadata:
  Overall # of frames:    614
  Duration of video [s]:  20.47
  fps:                    30.0
  resolution:             w=1920, h=1080

Running pose prediction with batch size 8
[...]

PR #2826, opened 2025-01-07
`create_labeled_video` fails with single animal TensorFlow models

n-poulsen:
As mentioned in [this comment from issue 2796](https://github.com/DeepLabCut/DeepLabCut/issues/2796#issuecomment-2574104417), `create_labeled_video` fails to create a labeled video for TensorFlow single-animal projects, as it could not find the "full_data" file which is not created for such models. The content of the file was used to plot bounding boxes in the video for top-down PyTorch models.

This pull request fixes the issue by catching the `FileNotFoundError` when the file is not present. The functional test `examples/testscript.py` is updated to check that the videos have all videos have been created successfully. This updated test fails with the when the bug is in the code, and succeeds with the fix above.

### DIFF
--- a/deeplabcut/utils/make_labeled_video.py
+++ b/deeplabcut/utils/make_labeled_video.py
@@ -888,9 +888,6 @@ def proc_video(
             df, filepath, _, _ = auxiliaryfunctions.load_analyzed_data(
                 destfolder, vname, DLCscorer, filtered, track_method
             )
-            full_data = auxiliaryfunctions.load_video_full_data(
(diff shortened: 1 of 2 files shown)


maximpavliv (approved):
Great, thank you @n-poulsen ! I think we can merge this

MHRosenberg:
Not fixed as far as I can tell, despite a successful pip upgrade. https://github.com/DeepLabCut/DeepLabCut/issues/2796#issuecomment-2593703565

PR #2833, opened 2025-01-10
Improve `deeplabcut.pose_estimation_pytorch` imports

n-poulsen:
Improves the way the most useful classes/methods from `deeplabcut.pose_estimation_pytorch` can be imported.

Additionally, makes the following improvements/fixes:

- a new `dlc_torch.config.make_basic_project_config` for COCO projects, to make the creation of model configuration files easier (see `deeplabcut/pose_estimation_pytorch/README.md` for an example)
- use `get_pose_inference_runner` and `get_detector_inference_runner` in `analyze_image_folder` (instead of `get_inference_runners`) to decouple detector and pose model loading.
- `video_inference` no longer needs to be given a `task`, as the task was not useful (if a detector runner is given, object detection will be run before pose estimation)
- `load_head_weights` was not correctly given to `train` in `train_network`
- `Loader.ground_truth_bboxes` now also returns image sizes, which makes evaluation of detectors easier


maximpavliv (approved):
LGTM, thanks @n-poulsen !
Btw, I added a couple of comments on the diagram on Figma

Issue #2835, opened 2025-01-11
How to specify snapshot to restart training from in command line?


mikeumo:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Bug description

I am trying to restart training session form a command line. I can specify the snapshot to restart training from in DLC 3.0 GUI. However, how to specify the same in a command line is unclear. The older documentation refers to setting an entry in pose_cfg.yaml but this file seems to have disappeared in the most recent DLC.

 So, how do I specify the initial model weights that are non-default?

### Operating System

MacOS Sequoia 15.2

### DeepLabCut version

3.0.0rc6, so, PyTorch engine

### DeepLabCut mode

single animal

### Device type

built-in gpu

### Steps To Reproduce

N/A

### Relevant log output

```shell
N/A
```

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/DeepLabCut/DeepLabCut/blob/master/CODE_OF_CONDUCT.md)

maximpavliv:
Hello,

To specify the snapshots to restart training from when using Pytorch engine, you can pass kwargs arguments to the `train_network()` method, specifically `"snapshot_path"` for the pose model snapshot and `"detector_path"` for the detector snapshot if you have one:
```
deeplabcut.train_network(
            '/analysis/project/reaching-task/config.yaml',
            shuffle=1,
            batch_size=8,
            epochs=100,
            save_epochs=10,
            display_iters=50,
            snapshot_path="/path/to/pose_snapshot/to/resume/training/from",
            detector_path="/path/to/detector_snapshot/to/resume/training/from",
        )
```
It is true that these arguments are missing in the docstrings, we will add them for more clarity.

In DLC3 (Pytorch), the _pose_cfg.yaml_ file has been replaced by the _pytorch_config.yaml_ file, you can also specify the snapshot to resume training from in that file with the `resume_training_from` key, see [here](https://deeplabcut.github.io/DeepLabCut/docs/pytorch/pytorch_config.html#restarting-training-at-a-specific-checkpoint).

mikeumo:
Great! Thank you!Mikhail On Jan 13, 2025, at 02:28, maximpavliv ***@***.***> wrote:﻿
Hello,
To specify the snapshots to restart training from when using Pytorch engine, you can pass kwargs arguments to the train_network() method, specifically "snapshot_path" for the pose model snapshot and "detector_path" for the detector snapshot if you have one:
[...]

PR #2836, opened 2025-01-13
Improve train_network() docstring

maximpavliv:
Adds `snapshot_path` and `detector_path` to kwargs arguments examples to the `train_network()` method docstring.

I decided to add this after #2835 .


n-poulsen (approved):
LGTM, thanks @maximpavliv

PR #2843, opened 2025-01-23
Reading apply_prob for augmentation pipeline from config

aiyeraditya:
The image augmentation pipeline always uses a value 0.5 for apply_prob.

This change allows the augmentation building pipeline for single animal projects to use the probability of applying augmentations (apply_prob) from the pose config file.

Curiously, this issue only exists for the single animal DLC. pose_multianimal_aug.py reads the apply_prob from the pose_config file.

OS Tested - Windows 10 Enterprise Python 3.10.4
[testscript_log.txt](https://github.com/user-attachments/files/18521303/testscript_log.txt)


maximpavliv (approved):
Amazing, thanks a lot @aiyeraditya !

PR #2844, opened 2025-01-23
Fix reading filtered state

maximpavliv:
In the Create Video tab in the GUI, unfiltered videos could not be created because of an incorrect reading of the "Use filtered data" checkbox state.
This PR fixes this bug


n-poulsen (approved):
@maximpavliv Looks good to me, thanks! Did you check if there's anywhere else in the code that we're using `checkState()`?

maximpavliv:
@n-poulsen You're right, this error was present in a few other places. Fixed them as well and refactored a bit, can you review again please?

n-poulsen (approved):
Looks good to me, thanks!

PR #2846, opened 2025-01-24
Remove pickle5 dependency from madlc Colab

maximpavliv:
Remove pickle5 dependency from examples/COLAB/COLAB_maDLC_TrainNetwork_VideoAnalysis.ipynb
It is useless there and creates dependency conflicts.


n-poulsen (approved):
Thanks @maximpavliv!

PR #2858, opened 2025-01-30
fixes to label_frames, addition of docstring

n-poulsen:
Improves the `deeplabcut.label_frames(...)` method:

- Fixes the call to `launch_napari` so an image folder is opened correctly.
- Option to not pass any arguments to the GUI so it just opens and drag-and-drop can be used
- Add a docstring indicating how the method can be used.

### DIFF
--- a/deeplabcut/gui/tabs/label_frames.py
+++ b/deeplabcut/gui/tabs/label_frames.py
@@ -92,7 +92,7 @@ def label_frames(
         else:
             image_dir = data_dir / image_folder
 
-        files = [str(image_dir), config_path]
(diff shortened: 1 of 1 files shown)


maximpavliv (approved):
Very nice, thanks a lot @n-poulsen !!

maximpavliv:
I think I found how to make it work in a simple python shell. Please wait a bit before merging, maybe this fix should be added too

maximpavliv (approved):

maximpavliv:
> I think I found how to make it work in a simple python shell. Please wait a bit before merging, maybe this fix should be added too

Let's address this in a future PR

PR #2859, opened 2025-02-03
Dynamic update of pytorch config with dot notation

maximpavliv:
In this Pull Request, in order to facilitate the dynamic update of pytorch_config.yaml (when calling `deeplabcut.train_network()`), I:

- Implement the `update_config_by_dotpath()` method. This method is easier to use than the previous `update_config()`. With `update_config()`, you needed to pass a nested dictionary as argument to target nested arguments in the pytorch_config.yaml, and it wasn't explicit how the nested dict was processed. With `update_config_by_dotpath()`, you target nested arguments by using a single-level dictionary with dot-notation path access. For example to update the `gpus` list in the `runner` section, you pass `{"runner.gpus": [0,1]}` (instead of `dict(runner=dict(gpus=[0,1]))`)
- Pass updates to the pytorch configuration with the `pytorch_cfg_updates: dict` argument instead of using the `kwargs` in `train_network()`.
- Make explicit all the other arguments that can be passed to `train_network()` (that were passed through the `kwargs` before) for more transparency, and remove the `kwargs` argument from `train_network()`.
- Implement a test for the new `update_config_by_dotpath()` method, update integration tests.

### DIFF
--- a/deeplabcut/pose_estimation_pytorch/apis/prune_paf_graph.py
+++ b/deeplabcut/pose_estimation_pytorch/apis/prune_paf_graph.py
@@ -142,8 +142,7 @@ def benchmark_paf_graphs(
             print()
 
         # update the edges to keep in the PyTorch configuration file
-        head_update = dict(predictor=dict(edges_to_keep=best_edges))
-        loader.update_model_cfg(dict(model=dict(heads=dict(bodypart=head_update))))
+        loader.update_model_cfg({"model.heads.bodypart.predictor.edges_to_keep": best_edges})
 
         # update the edges indices
         test_config = loader.model_folder.parent / "test" / "pose_cfg.yaml"
--- a/deeplabcut/pose_estimation_pytorch/apis/prune_paf_graph.py
+++ b/deeplabcut/pose_estimation_pytorch/apis/prune_paf_graph.py
@@ -78,7 +78,12 @@ def benchmark_paf_graphs(
 
     gt_train = loader.ground_truth_keypoints("train")
     best_paf_edges, _ = get_n_best_paf_graphs(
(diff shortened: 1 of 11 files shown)


deeplabcut/pose_estimation_pytorch/apis/prune_paf_graph.py
```diff
@@ -143,7 +143,7 @@ def benchmark_paf_graphs(
 
         # update the edges to keep in the PyTorch configuration file
         head_update = dict(predictor=dict(edges_to_keep=best_edges))
-        loader.update_model_cfg(dict(model=dict(heads=dict(bodypart=head_update))))
+        loader.update_model_cfg({"model.heads.bodypart": head_update})
```
n-poulsen: I believe this should be updated to:

```python
loader.update_model_cfg({"model.heads.bodypart.predictor.edges_to_keep": best_edges})
```

n-poulsen (changes requested):
Just one minor change needed - otherwise looks good to me!

deeplabcut/pose_estimation_pytorch/apis/prune_paf_graph.py
maximpavliv: You're right, thanks!

n-poulsen (changes requested):
The `examples/utils.py` file also needs to be updated; currently both the `testscript_pytorch_single_animal.py` and the `testscript_pytorch_multi_animal.py` fail

maximpavliv:
> The `examples/utils.py` file also needs to be updated; currently both the `testscript_pytorch_single_animal.py` and the `testscript_pytorch_multi_animal.py` fail

You're right, now I solved this!

n-poulsen (approved):
Looks good to me! Thanks @maximpavliv!

PR #2865, opened 2025-02-07
fix broken `train_network` argument `display_iters` in the GUI

n-poulsen:
Fixes the `displayiters` parameter when `train_network` is called from the GUI.

### DIFF
--- a/deeplabcut/gui/tabs/train_network.py
+++ b/deeplabcut/gui/tabs/train_network.py
@@ -320,7 +320,7 @@ def get_train_attributes(engine: Engine) -> list[TrainAttributeRow]:
                 attributes=[
                     IntTrainAttribute(
                         label="Display iterations",
-                        fn_key="display_iters",
+                        fn_key="displayiters",
                         default=1_000,
                         min=1,
                         max=100_000,


maximpavliv (approved):
LGTM!

mikeumo:
I have a error with this parameter. So, could you tell me which way is it? "displayiters" or "display_iters"?

Using rc7 on MacOS.

I had display_iters = 1000 parameter in my train_network() call, I started running my code without changing anything and I got error about "display_iters". I go to documentation and see that it is now "displayiters". Ok, but the example in the same help page uses "display_iters = 1000". Very confusing. So, I get rid of the parameter in the train_network() call and change pytorch_config.yaml to "displayiters: 1000". Get an error about 
File "/scratch/user/.../.conda/envs/dlc3/lib/python3.10/site-packages/deeplabcut/pose_estimation_pytorch/apis/training.py", line 196, in train display_iters=run_config["train_settings"]["display_iters"],
KeyError: 'display_iters'

Changed "displayiters" back to "display_iters" in the PyTorch_config.yaml.
Seems to be running.


So, which way is it?

maximpavliv:
Hello,

Sorry this is indeed confusing, I agree with you.

Basically:
- When there  was only Tensorflow as DeepLearning engine (before DLC 3.0), the naming was not consistent: the `train_network()` method exposed this argument as `displayiters`, but inside the method itself (and in the model configuration file, _pose_cfg.yaml_), the variable is named `display_iters`.
- When Pytorch support has been added (DLC 3.0), the variable has been consistently named `display_iters` inside the `pose_estimation_pytorch` submodule. This means that inside _pytorch_config.yaml_, it is named `display_iters`, and if you call the Pytorch-specific `train_network()` method implementation (`deeplabcut.pose_estimation_pytorch.train_network()`), you specify `display_iters`.
- To route [API-level methods](https://deeplabcut.github.io/DeepLabCut/docs/quick-start/single_animal_quick_guide.html) calls to their implementation with the right Deep Learning Engine, there is the _compat.py_ module. For example, when you call `deeplabcut.train_network()`, you actually call `deeplabcut.compat.train_network()`, [which itself then calls](https://github.com/DeepLabCut/DeepLabCut/blob/main/deeplabcut/compat.py#L61-L309) either `deeplabcut.pose_estimation_tensorflow.train_network.py` if you have a Tensorflow-based model, or `deeplabcut.pose_estimation_pytorch.train_network.py` if you have a Pytorch-based model. And in order to stay backwards-compatible with the previous DeepLabCut versions, we had to name the argument `displayiters` in `deeplabcut.compat.train_network()` (so that, if a user had a script that was doing `deeplabcut.train_network(displayiters=9999)` with DLC 2.0, they can still run their script after updating their DLC version to 3.0).

To summarize:
- When you do `deeplabcut.train_network()` - pass `displayiters`
- In your _pytorch_config.yaml_, specify `display_iters`
- If you use the pytorch module directly (`import deeplabcut.pose_estimation_pytorch as dlc_torch`, and then `dlc_torch.train_network()`), pass `display_iters`.

Thanks for spotting the error in the `deeplabcut.train_network()` docstring, I opened [a PR](https://github.com/DeepLabCut/DeepLabCut/pull/2939) to fix it, and sorry for the confusing naming. Unfortunately we cannot refactor to make it consistent everywhere, because ensuring full backwards compatibility is a strict requirement of DeepLabCut.

mikeumo:
Thank you for the thorough explanation!

PR #2871, opened 2025-02-07
Updates to the docs

n-poulsen:
Updates documentation to be compatible with DeepLabCut 2.0 and 3.0, for both the PyTorch and TensorFlow engine. Also updates:

- the `publish-book.yml` file to build the docs with Python 3.10
- the `_config.yml` so the docstrings build correctly
- removes the `DEEPLABCUT_M1.yaml` conda environment file, which is no longer needed for the PyTorch engine


maximpavliv (approved):
Perfect, thanks @n-poulsen !

PR #2872, opened 2025-02-10
fix evaluation for ma models, following change in 'deeplabcut/core/metrics/api.py'

Zelin2001:
This is a supplemental fix for Commit 40465e1 "fix - evaluation during training for ma models".

Following the principles of the previous fix, this minor change passed "compute_detection_rmse" argument to "compute_metrics()". To me, this fixed the assertion error in my multi-animal project which was comparable to what caused the 40465e1 fix.


n-poulsen (approved):
Hi @Zelin2001 - thanks for your fix! Looks good to me!

PR #2879, opened 2025-02-13
Fix assert in compute_detection_rmse()

maximpavliv:
Bug reported by a [post](https://forum.image.sc/t/deeplabcut-colab-installation-issues/107559/15) on the forum:

[This assertion](https://github.com/DeepLabCut/DeepLabCut/blob/5b6c22a89a9115203d277d9c570ed5c7ad750888/deeplabcut/core/metrics/distance_metrics.py#L366) is too restrictive, since the unique bodyparts might be missing in the ground truth annotations. 

This PR fixes this and adds a test for this case.

### DIFF
--- a/deeplabcut/core/metrics/distance_metrics.py
+++ b/deeplabcut/core/metrics/distance_metrics.py
@@ -363,14 +363,15 @@ def compute_detection_rmse(
 
     if data_unique is not None:
         for image_gt, image_pred in data_unique:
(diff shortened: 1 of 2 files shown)


maximpavliv:
@n-poulsen , before merging, I have a doubt: can we `assert len(image_pred) == 1` (in opposition to `len(image_gt)`, which can be 0 or 1)?

n-poulsen (approved):
LGTM! Thanks @maximpavliv

PR #2863, opened 2025-02-07
fix entrypoint for dlc

hmaarrfk:

PR #2883, opened 2025-02-17
Fix entrypoint - Requested changes

maximpavliv:
Following https://github.com/DeepLabCut/DeepLabCut/pull/2863:
- Change CLI command name from `dlc` to `deeplabcut`
- Create `main()` function in _deeplabcut/\_\_main\_\_.py_ (somehow in https://github.com/DeepLabCut/DeepLabCut/pull/2863 it was working, but I'm not sure why, as `deeplabcut.__main__:main` is calling the main() function, which didn't exist).

### DIFF
--- a/deeplabcut/__main__.py
+++ b/deeplabcut/__main__.py
@@ -9,20 +9,25 @@
 # Licensed under GNU Lesser General Public License v3.0
 #
 
-try:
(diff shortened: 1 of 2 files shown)


n-poulsen (approved):
LGTM!

PR #2889, opened 2025-02-18
Set pcutoff per-bodypart when evaluating PyTorch models

n-poulsen:
Adds an option to select the `pcutoff` for each bodypart individually for models trained with the PyTorch engine.

The `pcutoff` can be set as a `float`, a `list` or a `dict`. Examples for a project where there are 5 bodyparts: `["snout", "left_ear", "right_ear", "mid", "tail_base"]`.

Example 1: Using the same pcutoff for all bodyparts

```python
deeplabcut.evaluate_network(config, ..., pcutoff=0.6),
```

Example 2: Using a list to set the `pcutoff`

```python
deeplabcut.evaluate_network(config, ..., pcutoff=[0.6, 0.8, 0.8, 0.6, 0.6]),
```

Example 3: Changing the `pcutoff` for the ears to `0.8`, and otherwise keeping the same values as set in the project `config.yaml`: 

```python
deeplabcut.evaluate_network(config, ..., pcutoff={"left_ear": 0.8, "right_ear": 0.8})
```

Tests were added for evaluation.

### DIFF
--- a/deeplabcut/compat.py
+++ b/deeplabcut/compat.py
@@ -386,6 +386,7 @@ def evaluate_network(
     modelprefix: str = "",
     per_keypoint_evaluation: bool = False,
     snapshots_to_evaluate: list[str] | None = None,
+    pcutoff: float | list[float] | dict[str, float] | None = None,
     engine: Engine | None = None,
     **torch_kwargs,
 ):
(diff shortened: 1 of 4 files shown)


MMathisLab:
Awesome! this is a super useful feature for videos and evaluation, etc! Can we add into the DOC PR the changes too?

PR #2891, opened 2025-02-18
Facelift for codespell config: newer codespell, annotation to PRs, codespell config centralization

yarikoptic:
Individual commit messages might say more


AlexEMG (approved):
Thanks!

Issue #2732, opened 2024-09-03
Create projects from SuperAnimal models


codikzng:
Dear DeepLabCut Team,

Thank you for your tremendous work on this invaluable open-source tool. I truly appreciate your efforts and dedication.

I have a suggestion regarding the use of pre-trained superanimal models from the modelzoo for inference. Currently, when using these models purely for inference (without fine-tuning or transfer learning), a project directory and config.yaml file are not created, unlike the standard DeepLabCut projects. This makes it difficult to use the results for further analysis.

It would be very helpful if the inference process with pre-trained superanimal models could also generate a project directory with the necessary files, such as config.yaml. This would ensure consistency across all DeepLabCut projects and make downstream analysis more practical and efficient.

Thank you for considering this suggestion. I hope it contributes to enhancing the usability of DeepLabCut for everyone.

Best regards,

n-poulsen:
Hi @codikzng! As described in the [COLAB ModelZoo example Notebook](https://github.com/DeepLabCut/DeepLabCut/blob/main/examples/COLAB/COLAB_DLC_ModelZoo.ipynb), we have a [`create_pretrained_project`](https://github.com/DeepLabCut/DeepLabCut/blob/e499671fd13a68a71cb606453bc22989bc8d87e9/deeplabcut/create_project/modelzoo.py#L98) method to create projects from ModelZoo models.

I don't think this will work out-of-the-box with SuperAnimal models, but I think it would be a great feature to have as well! We'll work on adding such a method for SuperAnimal models, but if you want to write the code yourself and PR it into DeepLabCut that would definitely speed up development ;)

maximpavliv:
This new feature has just been added with https://github.com/DeepLabCut/DeepLabCut/pull/2897 !

codikzng:
That's too cool!

PR #2881, opened 2025-02-14
[WIP] DLC 3.0 - Review documentation

maximpavliv:
Updated documentation for the upcoming release of DLC 3.0

PR #2897, opened 2025-02-21
Implement create_pretrained_project() for Pytorch engine

maximpavliv:
**This Pull Request adds Pytorch support for the `create_pretrained_project()` method (and closes https://github.com/DeepLabCut/DeepLabCut/issues/2732)**

Specifically:
- `create_pretrained_project()` becomes a `compat.py`-like method, and calls the specific method implementation based on the specified `engine` parameter. 
The old `create_pretrained_project()` is renamed to `create_pretrained_project_tensorflow()` and a new `create_pretrained_project_pytorch()` method with the Pytorch-shuffle logic is added.
The existing usages of `create_pretrained_project()` are also adapted.
- Both `create_pretrained_project_*` implementations call `_create_training_datasets_metadata()`. 
This method has been cherry-picked from #2881(where a first version, `_create_and_save_fake_metadata()`, has been implemented as a temporary fix), and adapted to work with both Tensorflow and Pytorch. It creates the `metadata.yaml ` file in the training-datasets directory, and fills it with a first shuffle corresponding to the downloaded modelzoo model.
- The video processing logic (that is optionally done after the pretrained project has been created) is extracted to 
`_process_videos()`, and called by both `create_pretrained_project_*` implementations.
- A new `individuals` parameter is added to `create_new_project()` and to `create_pretrained_project()`, so that the list of individuals can be automatically filled in the different configuration files of the new project.
- An integration test script for `create_pretrained_project(engine=Engine.PYTORCH)` is added.

**This Pull Request also does some fixes to issues that have been spotted during its development:**
- Fixes `list_videos_in_folder()` : in original (Tensorflow-implementation) methods (f.ex. `analyze_videos()`), `videotype` argument's value by default was `""` (instead of `None`), and this default argument value was not treated properly.
- `get_super_animal_project_cfg()` was using `read_config()` to read the superanimal project configuration template embedded into the DeepLabCut repository, which then modified the file in the DeepLabCut codebase. Now, the method reads the template with `read_config_as_dict()`, which doesn't modify the file.
- The previous fix created a circular-import issue, that has been resolved by moving the `read_config_as_dict()`, `write_config()` and `pretty_print()` methods to `deeplabcut.core.config_utils.py` module.
- The other superanimal integration test scripts (`testscript_superanimal_adaptation.py`, `testscript_superanimal_inference.py` and `testscript_superanimal_transfer_learning.py`) have been fixed.


maximpavliv:
The Tensorflow implementation of `create_pretrained_project()` fills the `"default_net_type"` field in the project configuration file with the chosen modelzoo's model architecture. 
In the Pytorch implementation of `create_pretrained_project()` I wasn't sure how to do this, because we would also then need a "default detector type" field (particularly since Superanimal models are all Top-Down currently), but such a field doesn't exist. Therefore I left the `"default_net_type"` field  emtpy for now. 
Any suggestions welcome on this point.

deeplabcut/create_project/new.py
```diff
@@ -58,6 +59,9 @@ def create_new_project(
     multianimal: bool, optional. Default: False.
         For creating a multi-animal project (introduced in DLC 2.2)
 
+    individuals: list[str]|None = None,
+        todo
+
```
n-poulsen: We could add a docstring here

n-poulsen (approved):
LGTM @maximpavliv, thanks! just one minor comment

deeplabcut/create_project/new.py
maximpavliv: Right, I forgot!

PR #2901, opened 2025-02-25
Fixes failing tests

n-poulsen:
Fixes tests that were failing due to sub-optimal file naming. Checks for `float` and `int` pcutoff values. Fixes warnings raised in tests:

- rename `np.product` to `np.prod`
- removes unnecessary cast to torch tensor (when the returned value was already a torch tensor)


maximpavliv (approved):
Thanks @n-poulsen !

Issue #2787, opened 2024-11-13
Save x,y coordinates from predicted keypoints after evaluation


ValeriePineauNoel:
**Is your feature request related to a problem? Please describe.**
I have a quest where I want to compare the performances of a few pipelines for pose estimation (including DLC). To do so, I am looking for metrics that I can compute to compare quantitatively pipelines together. One of the things that I wish DLC would provide is the x,y coordinates of the predicted keypoints, so that I have the liberty of computing the metrics however I want afterwards. I know that DLC provides the euclidien distances between predicted and ground-truth keypoints, but providing the x,y coordinates would also be useful I think. 

**Describe the solution you'd like**
In my case, since I am training a multianimal pipeline, I think I would just modify the file _evaluate_multianimal.py_. At line 453, I would convert the coords_pred to a dataframe (just like you already do with the distances and the confidence) and then joint this dataframe to df_joint. The csv would save the coordinates as well as the distances and the confidence. 

**Describe alternatives you've considered**
Are there any? If there is an easier way to get the predicted coordinates of the keypoints, I would be very happy to know them. 

**Additional context**

maximpavliv:
The feature has been added in https://github.com/DeepLabCut/DeepLabCut/pull/2902!

PR #2613, opened 2024-06-13
🔥 New main backend Deep Learning Library (Engine) for DeepLabCut: PyTorch

MMathisLab:
## New main backend Deep Learning Library (Engine) for DeepLabCut: PyTorch 🔥

This post serves to provide the source code of DeepLabCut 3.0, which marks the first major release change since 2 (in late 2018). With this change, we move to PyTorch.

PyTorch combines the GPU-accelerated backend libraries from [Torch](http://torch.ch/) with an intuitive Python frontend that focuses on rapid prototyping, readable code, and support for the widest possible variety of deep learning models.

It is high-performance, easier to use, easier to install, and developer-friendly.

img-1
Image reference https://pytorch.org/features/

## Why:

We wanted to modernize our stack and also make it more user-friendly for all our stakeholders. Namely, the stakeholders in DeepLabCut are:
- end users in life sciences
- computer vision researchers
- animal pose estimation developers

## What is new:

We have dedicated documentation that walks you through the new options
(note, these links point to a PR, but will be in main docs upon merger)

- [User Guide & RoadMap](https://github.com/DeepLabCut/DeepLabCut/blob/pytorch_docs/docs/pytorch/user_guide.md)

- [PyTorch configuration file](https://github.com/DeepLabCut/DeepLabCut/blob/pytorch_docs/docs/pytorch/pytorch_config.md)

- [Neural Networks](https://github.com/DeepLabCut/DeepLabCut/blob/pytorch_docs/docs/pytorch/architectures.md)

## What to expect:

**Importantly, the high-level API stays the same 🪄.**
The dev team worked very hard to keep both DeepLabCut PyTorch and (for the time) DeepLabCut TensorFlow for you and keep the workflow identical. Truly, you can use the same 12 steps you know and love! ❤️

### In short:
Screenshot 2024-06-12 at 16 12 29 (1)
Screenshot 2024-06-12 at 16 12 54

**You have the GUI, with the same launch code:**
```bash
python -m deeplabcut
```
**And at the command line interface you have [the same 12 core steps!](https://deeplabcut.github.io/DeepLabCut/docs/single_animal_quick_guide.html)**

## Get started by testing this PR:

First, this is not yet on `pypi`, so you will need to test this branch 💜. This is straightforward to do:

(1) We recommend building a new `conda` environment. In your anaconda terminal, please type:

```
conda create -n deeplabcut3 python=3.10
```
[...]

PR #2902, opened 2025-02-25
Save predicted poses in evaluate_multianimal_full()

maximpavliv:
This PR addresses https://github.com/DeepLabCut/DeepLabCut/issues/2787 , and saves the poses used in network evaluation of a multi-animal project with TensorFlow engine.

In network evaluation of a single-animal project with TensorFlow engine, the poses were already saved as a pandas Dataframe into a `.h5` file. The same format is also already used for both single and multi-animal projects using Pytorch as engine, on the [pytorch_dlc branch](https://github.com/DeepLabCut/DeepLabCut/pull/2613). 
In order to match this workflow as closely as possible, this PR saves the poses as a pandas Dataframe into a `.h5` file as well. 
If needed, the `.h5` file can then be easily read into a pandas dataframe using
```python
df = pd.read_hdf(h5_file_path, key="df_with_missing")
```
and then saved into a `.csv` format file using
```python
df.to_csv(csv_file_path)
```


AlexEMG (changes requested):
I'd suggest we add a flag to set this to true/false, but make it false by default?

maximpavliv:
> I'd suggest we add a flag to set this to true/false, but make it false by default? 

I'm not sure, because all the other scenarios (TF single animal and Torch single and multi animal) all save the predicted poses during evaluation into a `.h5` file in any case.

So i could add the flag for the specific case of multi-animal TF models evaluation, and keep the others as they are? Or add the flag and use it in all cases, thus modifying the current behavior of the other scenarios?

I think it would be best to match the style of the other network evaluation scenarios, and just save the poses into an `.h5` file in any case, without adding any flags.

n-poulsen (approved):

AlexEMG:
That's right @maximpavliv!

AlexEMG (approved):

AlexEMG:
@maximpavliv, can you please also merge into the 2.*LTS branch?

maximpavliv:
> @maximpavliv, can you please also merge into the 2.*LTS branch?

Yes, good idea!

[edit] Just opened a PR: https://github.com/DeepLabCut/DeepLabCut/pull/2925

PR #2903, opened 2025-02-26
Improve tracklet creation for ID-only tracking

n-poulsen:
This pull request fixes tracklet creation with `identity_only=True`. It

- only keeps valid predictions to assign identity to
- removes poses with confidence below the inference `pcutoff`
- updates the way the softmax is computed to assign each assembly (total score across keypoints with a given identity instead of average score)

It also fixes identity metric computation when NaNs are in the predictions.


maximpavliv (approved):
LGTM, thanks @n-poulsen !

PR #2905, opened 2025-02-26
Tracking benchmarking - Updates

maximpavliv:
This Pull Request updates few parts related to tracking:
Mainly:
- Fixes `create_video()` 's `bbox` argument's default value case, which was causing an `Exception`
- Adds docstrings to methods used for tracking
- Fixes `SORTBox` constructor args (it was constructed using `oks_threshold` parameter instead of `iou_threshold`)
- Makes `convert_detections2tracklets()` lighter by moving the tracklets building part into its own `build_tracklets()` method, and makes `build_tracklets()` lighter by moving the tracklets header creation into its own `_create_tracklets_header()` method


AlexEMG (changes requested):
Fantastic, can we also add unit-tests for those functions in case they are missing?

n-poulsen (approved):
LGTM! Thanks @maximpavliv

AlexEMG (approved):

PR #2907, opened 2025-02-27
NMS for DEKR models

n-poulsen:
Implements pose non-maximum suppression for DEKR as done in the official [DEKR implementation](https://github.com/HRNet/DEKR/blob/7a303139e92bdf3eab8d899415ccac37374285a4/lib/core/nms.py#L62). Some slight changes were made to the code so it could be applied on batches of predictions.


deeplabcut/pose_estimation_pytorch/models/predictors/dekr_predictor.py
```diff
@@ -358,51 +365,84 @@ def get_heat_value(
...
+        # compute the difference between keypoints
+        pose_diff = xy_[:, None] - xy_
+        pose_diff.pow_(2)
+
+        # put batch first again
+        pose_diff = pose_diff.transpose(2, 0)
+
+        # Compute error between people pairs
+        pose_dist = pose_diff.sum(dim=-1)
+        pose_dist.sqrt_()
```
maximpavliv, line 410: I think we can compute the distance more efficiently:
```python
pose_dist = torch.norm(xy[:, :, None] - xy[:, None, :], dim=-1)
```
(not tested)

deeplabcut/pose_estimation_pytorch/models/predictors/dekr_predictor.py
```diff
@@ -358,51 +365,84 @@ def get_heat_value(
...
+                suppressed = (
+                    nms_pose[batch_idx, person_idx]
+                    .nonzero()
+                    .detach()
+                    .reshape(-1)
+                    .tolist()
+                )
+                for to_suppress in suppressed:
+                    if to_suppress not in kept_indices:
+                        ignored_indices.add(to_suppress)
```
maximpavliv, line 439: I think this can be optimized (code-length-wise, and speed-wise):
```python
kept = torch.ones((batch_size, num_people), dtype=torch.bool, device=poses.device)
for person_idx in range(num_people):
    suppress_mask = nms_pose[:, person_idx] & kept
    suppress_mask[:, person_idx] = False  # Keep the current person
    kept &= ~suppress_mask
```
(not tested)

maximpavliv (changes requested):

maximpavliv (approved):
LGTM, thanks @n-poulsen !

PR #2908, opened 2025-02-27
_create_video_from_tracks fix

KonradDanielewski:
Fixes two things:

1. in `make_labeled_video.py` adjust the shape of the data used (slice without identity column) as it was causing: 
```python
ValueError: cannot reshape array of size ## into shape (3)
```
also removes extracted frames used for video creation, they take up a lot of space.

2. in `pose_estimation_pytorch/apis/tracklets.py`

change frame key naming to be backwards compatible (was named with just index now `frame000..`)


AlexEMG (approved):
We should also push the make_labeled_videos fix into the 2.* version!

KonradDanielewski:
Maybe one thing could be changed for safety if users set the `destfolder` to an existing directory that has other `png` files. Instead of removing based on `.png` remove files with `frame` in name. Let me know what you think.

AlexEMG:
That's a good idea!

n-poulsen (approved):
LGTM! Thanks @KonradDanielewski!

This can be merged once tests pass.

PR #2915, opened 2025-03-03
Fix testscript_superanimal_create_pretrained_project

maximpavliv:
_testscript_superanimal_create_pretrained_project.py_ was using a video in _examples/openfield-Pranav-2018-10-30/videos/_ that is originally not present in the directory, and created by another testscript.

This PR fixes this by using the original video that is always present in _examples/openfield-Pranav-2018-10-30/videos/_ .


MMathisLab (approved):
lgtm :D

PR #2917, opened 2025-03-07
Fix circular import

maximpavliv:
Fix circular import introduced in https://github.com/DeepLabCut/DeepLabCut/pull/2897, which was logging
`DLC loaded in light mode; you cannot use any GUI (labeling, relabeling and standalone GUI)`
when launching the DLC GUI.

### DIFF
--- a/deeplabcut/create_project/modelzoo.py
+++ b/deeplabcut/create_project/modelzoo.py
@@ -22,8 +22,8 @@ from dlclibrary.dlcmodelzoo.modelzoo_download import (
 )
 
 import deeplabcut
-from deeplabcut import Engine
 from deeplabcut.core.config import read_config_as_dict, write_config
+from deeplabcut.core.engine import Engine
 from deeplabcut.generate_training_dataset.metadata import (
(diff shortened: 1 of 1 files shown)


MMathisLab (approved):
lgtm; did not directly test though; in general @maximpavliv when lauching the GUI with `python -m deeplabcut`, this message should not appear

n-poulsen (approved):
LGTM! Tested, fixes the issue.

YannickJadoul:
Quick thing: I just tracked down the exact same issue with a colleague who had freshly installed DLC according to the docs' installation instructions: https://deeplabcut.github.io/DeepLabCut/docs/installation.html

That fresh installation didn't have this fix of the circular imports yet and could only load DLC in "light mode". I'm assuming it's because anaconda environment provided by the docs is installing a branch, and the current PR is not yet included in there?
https://github.com/DeepLabCut/DeepLabCut/blob/de5ea59adb89b35964099e6ded5a17e000f53e1d/conda-environments/DEEPLABCUT.yaml#L38

PR #2918, opened 2025-03-07
Add a COLAB for human pose estimation using RTMPose

n-poulsen:
This PR adds an example notebook to run human pose estimation on images and videos using RTMPose.


maximpavliv (approved):
Tested, everything works, thanks @n-poulsen !

PR #2919, opened 2025-03-07
Dynamic cropping for top-down models

n-poulsen:
Adds a `TopDownDynamicCropper` for top-down models. The `TopDownDynamicCropper` can be used instead of an object detector to analyze videos **containing a single animal** with top-down models. At frame 0, the full frame is split into (n, m) image patches, with a given overlap between the patches. Patches are then

- Resized to the input size required by the model with a top-down crop.
- Stacked into a batch and given to the pose estimation model
- The output poses for each patch are post-processed: the patch containing the highest average score prediction is selected as the patch containing the individual, and the pose from that patch is selected as the predicted pose.

At frame n, one of two things can happen:
- If the individual was successfully detected at frame n - 1, a bounding box is generated from the predicted pose and used as the bounding box for the next frame.
- If the individual was not detected at frame n - 1, patches are cropped as in frame 0 and the pose selected as in frame 0

An individual is considered to be successfully detected if at least `min_hq_keypoints` keypoint have scores above the `threshold`.

The bounding box is generated from the keypoints (either from all keypoints or only the ones above the threshold) with a margin around the keypoints. If the bounding box is smaller than a set minimum size, it is expanded to that size.


deeplabcut/pose_estimation_pytorch/apis/utils.py
```diff
@@ -678,7 +682,7 @@ def get_pose_inference_runner(
     if transform is None:
         transform = build_transforms(model_config["data"]["inference"])
 
-    if pose_task == Task.BOTTOM_UP:
+    if pose_task == Task.BOTTOM_UP or isinstance(dynamic, TopDownDynamicCropper):
```
maximpavliv, line 698: The docstring description of the `dynamic` argument needs to be updated (`Only for bottom-up pose estimation models` is not true anymore)

deeplabcut/pose_estimation_pytorch/runners/dynamic_cropping.py
```diff
@@ -95,19 +96,19 @@ def crop(self, image: torch.Tensor) -> torch.Tensor:
...
 
-    def update(self, pose: torch.Tensor) -> None:
+    def update(self, pose: torch.Tensor) -> torch.Tensor:
         """Updates the dynamic crop according to the pose model output.
 
         Uses the pose predicted by the model to update the dynamic crop parameters for
         the next frame. Scales the pose predicted in the cropped image back to the
         original image space and returns it.
 
-        This method modifies the pose tensor in-place; so pass a copy of the tensor if
```
maximpavliv: This method still modifies the pose tensor in-place, so maybe we should keep this comment? (Or is it already obvious with the method's name?)

tests/pose_estimation_pytorch/runners/test_dynamic_cropper.py
```diff
@@ -129,6 +132,51 @@ def test_dynamic_cropper_basic_crop(
...
+
+    # check that the difference between overlaps is at most 1
+    # FIXME(niels) - auto-correct the overlap to spread it out more evenly
+    # if n > 1:
+    #     overlaps = [
+    #         end_1 - start_2
+    #         for (start_1, end_1), (start_2, end_2) in zip(sections[:-1], sections[1:])
+    #     ]
+    #
+    #     assert max(overlaps) - min(overlaps) <= 1
```
maximpavliv, line 177: What about this?

maximpavliv (changes requested):
Adding this to the docs would be amazing!

Also, in the PR description, you write:

> Calling through the analyze_videos method still needs to be tested before this PR is merged.

Is this still needed to be done?

n-poulsen:
Yes `analyze_videos` was tested - I fixed the description

tests/pose_estimation_pytorch/runners/test_dynamic_cropper.py
n-poulsen: This test currently fails, as the generation of the patches can be improved (so that "extra" overlap is spread more evenly). I'd be in favor of keeping the to-do, as this can be a future improvement but doesn't really hurt performance.

deeplabcut/pose_estimation_pytorch/apis/utils.py
n-poulsen: Fixed in the docs!

maximpavliv (approved):
LGTM, thanks @n-poulsen !

PR #2927, opened 2025-03-19
DeepLabCut: Fix pandas FutureWarning in evaluate_multianimal.py (inplace=True in .replace())

Denny-Hwang:
Fix pandas FutureWarning in `evaluate_multianimal.py` related to `inplace=True` in `.replace()`

This Pull Request addresses a `FutureWarning` originating from the pandas library that appears during the evaluation of multi-animal pose estimation models in DeepLabCut.

**Problem:**

The warning is triggered by the use of the `.replace()` method with the `inplace=True` argument in the `evaluate_multianimal.py` file (likely around line [**Insert the line number you found, e.g., 346**]). Pandas is discouraging the use of `inplace=True` in combination with chained assignment, as it can lead to unexpected behavior and will be removed in a future version.

**Solution:**

This PR modifies the line to remove the `inplace=True` argument and instead assigns the result of the `.replace()` operation back to the DataFrame column. For example, the line was changed from:

```python
temp["bodyparts"].replace(..., inplace=True)

### CODE REVIEW

**Comment by n-poulsen** [comment]
Hi @Denny-Hwang! Thanks so much for taking the time to look into this! 

I've tested your code but am now getting a new warning (with `pandas==2.2.3` installed):

```bash
FutureWarning: Downcasting behavior in `replace` is deprecated and will be removed in a future 
version. To retain the old behavior, explicitly call `result.infer_objects(copy=False)`. To opt-in to 
the future behavior, set `pd.set_option('future.no_silent_downcasting', True)`
```

Based on [this thread](https://github.com/pandas-dev/pandas/issues/57734), it seems the issue comes from converting the bodyparts (which are `str`) into indices (which are `int`). I think one way to solve this next warning would be to silence the warning (with `pd.option_context`) and then convert to `int`s using `infer_objects(copy=False)`.

```python
with pd.option_context('future.no_silent_downcasting', True):
    temp["bodyparts"] = temp["bodyparts"].replace(
        dict(zip(joints, range(len(joints)))),
    ).infer_objects(copy=False)
```

Could you try fixing this warning at the same time?

MMathisLab:
@n-poulsen shall we push our suggestion then merge?

n-poulsen (approved):
LGTM!

MMathisLab (approved):

PR #2933, opened 2025-03-24
Add `bbox_margin` as a parameter to the pytorch config

n-poulsen:
This pull request:

- adds a `bbox_margin` parameter to the `pytorch_config.yaml` file, configuring how much margin should be added around keypoints to compute bounding boxes
- the default value is set to `20` (if the parameter is not set), so the current behavior of the code isn't changed
- adds documentation about this parameter in `docs/pytorch/pytorch_config.md`


maximpavliv (approved):
LGTM, thanks @n-poulsen

PR #2938, opened 2025-03-28
Fix missing detector in Create Training Dataset GUI with animaltokenpose and rtmpose architectures

maximpavliv:
In the Create Training Dataset GUI tab, the detector-related widgets were not shown with _animaltokenpose_ and _rtmpose_ architectures, despite they are Top-Down architectures. This Pull Request fixes this.


n-poulsen (approved):
LGTM! Thanks @maximpavliv!

PR #2945, opened 2025-04-04
added a method for curating frame extraction within a start and end t…

gsaaad:
Curating a specific number of frames using user's chosen start and end time in HH:MM:SS format. 
Minimum gap of frames between selected frames either in int # or percentage of the total frames of the video 2% gap. 
Introduced a separate number of frames selection using DLCs widget. 
Output pushes frames into a similar fashion with a meta-data like file [can be removed if redundant].

curatedFrames for a specific timepoint within a single selected video, using start-end HH:MM:SS format, allowing user to select [for example] behavior specific frames, or only when the mice are visible within a time window.

@AlexEMG


gsaaad:
Let me know what you require to proceed

gsaaad:
Does this need any more work or improvements? LMK

deruyter92:
@gsaaad, apologies for the delay, and thanks again for your contribution! 

 The development team is currently working on improvements to internal validation and expanding the test coverage for DeepLabCut. This is unfortunately prioritized over assisting in the co-design of additional features, such as the one proposed in this PR.

This PR would still require some major changes before it could be added to the main repository. To name a few issues:
- Mismatching mode name `AER-Curated` causes a bug
- Undefined variables (e.g. `videos` outside conditional)
- Inconsistent conditionals for state `manual`
- Vulnerabilities in path traversals
- Potential resource leaks for the VideoCapture
- No input validation
- No unit tests

Therefore I decided to close the current PR for now. Feel free to re-open a PR in case you want to apply these changes and once you think the code is in a solid state that is safe to merge.

Thanks again!

gsaaad:
@deruyter92 Thank you for getting back, no worries.
I will actively work on this to make it safe to merge.

Issue #2922, opened 2025-03-13
` pytorch` learning rate too aggressive on my dataset (defaults with TF are fine)


tsievert:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Bug description

I have a model that I trained for 5 iteration with great succes. I labelled extra outliers and upgraded to 3.0.0rc7 and wanted to train the new iteration with pytorch.
The pytorch training shows basically now learning, but if I switch back to TF, the model seems to learn just fine.
I have about 2300 labelled frames, in case larger datasets are probelematic.

### Operating System

operating systVersion: Linux version 4.18.0-553.34.1.el8_10.x86_64 ([mockbuild@x86-64-03.build.eng.rdu2.redhat.com](mailto:mockbuild@x86-64-03.build.eng.rdu2.redhat.com)) (gcc version 8.5.0 20210514 (Red Hat 8.5.0-22) (GCC)) 

### DeepLabCut version

3.0.0rc7

### DeepLabCut mode

single animal

### Device type

gpu

### Steps To Reproduce

1. Have existing trained model (pre DLC v3.0.0)
2. Extract outliers
3. Upgrade to latest 3.0.0rc
4. Label frames
5. Switch `config.yaml` to pytorch
6. Train model for 500 epochs
7. Observe no learning
```bash
Evaluation results for DLC_Resnet50_Callibration_analysisSep14shuffle1_snapshot_010-results.csv (pcutoff: 0.8):
train rmse            933.00
train rmse_pcutoff       NaN
train mAP               0.00
train mAR               0.00
test rmse             934.06
test rmse_pcutoff        NaN
test mAP                0.00
test mAR                0.00
Name: (0.95, 1, 10, -1, 0.8), dtype: float64
```
8. Add the engine switch to TF, train on same data
9. Normal learning

[learning_stats.csv](https://github.com/user-attachments/files/19234053/learning_stats.csv)

### Relevant log output

```shell

```

### Anything else?

I can provide all or a subset of my labelled frames if that helps.

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/DeepLabCut/DeepLabCut/blob/master/CODE_OF_CONDUCT.md)

n-poulsen:
Hi @tsievert! Thanks for reporting this - could you share the content of your `pytorch_config.yaml` with me?

tsievert:
Sure, here you go:
```yaml
data:
  colormode: RGB
[...]

PR #2960, opened 2025-04-14
lower default LR to 5e-4 for ResNets and CSPNeXt

n-poulsen:
Addresses the issue found in #2922. For some datasets, setting `1e-3` as an initial learning rate is too aggressive and prevents the model from learning. This PR changes the default to `5e-4` for ResNets and CSPNeXt backbones (other models already have lower learning rates as default), which resolved the issue on datasets where it occurred and led to more stable learning.


MMathisLab (approved):
lgtm

PR #2971, opened 2025-04-26
Use R-strings in docstrings containing backslashes

hmaarrfk:
Use R-strings in docstrings containing backslashes.


maximpavliv (approved):
Thanks @hmaarrfk!

PR #2977, opened 2025-05-05
Sleap to Deeplabcut conversion method 

agosztolai:
This PR implements a method to **convert a Sleap project to DeepLabCut project**, as suggested in https://github.com/DeepLabCut/DeepLabCut/issues/2734.

### DIFF
--- a/deeplabcut/utils/conversioncode.py
+++ b/deeplabcut/utils/conversioncode.py
@@ -397,25 +397,32 @@ def robust_split_path(s):
     return tuple(s.split(sep))
 
 
-def extract_frames_from_pkg_sleap(file_path, base_output_dir, scorer="me"):
+def convert_sleap_to_deeplabcut(
+    sleap_file: str | Path,
+    deeplabcut_dir: str,
+    scorer: str = "me",
+):
     """
     Convert a SLEAP project into a DeepLabCut project
     WARNING: Conversion might corrupt the data.
-    file_path : string
-        Path to the .slp file.
-    base_output_dir: string
-        Output directory.
-    scorer: string, optional
-        Name of scorer.
+
     Once conversion is complete, you will need to manually create a DeepLabCut config.yaml file based
     on the keypoint names in the generated .csv files.
     Finally, run deeplabcut.convertcsv2h5('config.yaml')
+
+    Args:
+        sleap_file : string | Path
+            Path to the .slp file.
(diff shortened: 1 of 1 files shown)


maximpavliv:
Hey @agosztolai 

I refactored a bit your method, hope it's okay for you 😀.

I am testing it now with a simple Sleap project I created, but it doesn't work for now on my side: it basically never enters inside [this condition](https://github.com/agosztolai/DeepLabCut/blob/afcddfcc2581a25127c5552eef0c5721961198ec/deeplabcut/utils/conversioncode.py#L467-L471), and, having manually checked, I don't see anything like this in my Sleap file. My Sleap file contains the following keys:
`<KeysViewHDF5 ['frames', 'instances', 'metadata', 'points', 'pred_points', 'suggestions_json', 'tracks_json', 'videos_json']>`.
The dataset under `'videos_json'` does store the path to the video I used when creating my Sleap project.
Therefore `video_names` is empty and nothing is done inside [this whole `for` loop](https://github.com/agosztolai/DeepLabCut/blob/afcddfcc2581a25127c5552eef0c5721961198ec/deeplabcut/utils/conversioncode.py#L474-L596) neither.

Is there an issue with your code or did I do something wrong when creating my Sleap project? I am new to Sleap so maybe the mistake is on my side. 
Or maybe we have different Sleap versions, which store data differently? I currently have _SLEAP v1.4.1a2_

Bzver:
Hello @maximpavliv , 

I can confirm that the code works on _SLEAP v1.4.1a2_, but it needs the .pkg.slp file from 'Predict - Export Labels Package... - Labelled frames' instead of the .slp project file.
scr

By the way, I have a couple of suggestions:
1. The line 527 and 528 should be `instance_dataset[i]` instead of `instance_dataset[i+1]`, the latter caused an out of range error on my end. 
2. I recommend adding the line
`"Task": f"{scorer}-{sleap_file.split('labels.')[1].rsplit('.pkg.slp', 1)[0]}", `
(Assuming the sleap_file follows the format of /path/to/labels.xxx.pkg.slp) 
below `config = {` at line 421, without which will throw a KeyError: ‘Task' when attempting to load from the generated config.yaml.

Bzver:
A further issue: during the 'Analyze Videos' step in DLC, the project froze due to the generated config.yaml missing the `multianimalproject:`

PR #2990, opened 2025-05-21
Fix `deeplabcut.analyze_images()` with CTD model

maximpavliv:
The [integration of CTD models](https://github.com/DeepLabCut/DeepLabCut/pull/2952) adapted the `deeplabcut.analyze_videos()` to support CTD models, but missed adapting the `deeplabcut.analyze_images()` API method accordingly, which resulted in a crash when trying to use `deeplabcut.analyze_images()` with a CTD model.

This Pull Request fixes this by:
- Creating and using a **conditions provider** (`cond_provider`) [when creating a pose inference runner](https://github.com/DeepLabCut/DeepLabCut/blob/1f102b447c1970b469a865ee2d53644b555c6d83/deeplabcut/pose_estimation_pytorch/apis/analyze_images.py#L471-L477) in `analyze_image_folder()`.

- Adding a `**torch_kwargs` argument to the `analyze_images()` in _compat.py_ (to allow passing `ctd_conditions` at this level). If left unspecified - the conditions are loaded accordingly to the specifications in the _pytorch_config.yaml_ file.

- The `create_labeled_images()` method in _deeplabcut/pose_estimation_pytorch/apis/visualization.py_ has been adapted as well. In fact, previously, the method assumed that, if the `image_predictions` contained a `"bboxes"` key - it also contained a `"bbox_scores"` key, both being generated by detectors in Top-Down architectures. Now, with the addition of CTD models, the `image_predictions` contain a `"bboxes"` key, but the bounding boxes have been built using the condition poses, therefore `image_predictions` doesn't contain a `"bbox_scores"` key. In summary, with this change, the method doesn't try to draw bounding boxes when using a CTD model.

**/!\ Spotted limitation**: the `deeplabcut.analyze_images()` API method is never tested in the integration tests!


deeplabcut/pose_estimation_pytorch/apis/analyze_images.py
```diff
@@ -101,7 +103,7 @@ def superanimal_analyze_images(
             The device to use to run image analysis.
 
         pose_threshold: float, default=0.4
-            The cutoff score when plotting pose predictions. To note, this is called 
+            The cutoff score when plotting pose predictions. To note, this is called
```
MMathisLab: why not use pcutoff here?

deeplabcut/pose_estimation_pytorch/apis/analyze_images.py
```diff
@@ -189,6 +191,7 @@ def superanimal_analyze_images(
         max_individuals=max_individuals,
         device=device,
         progress_bar=progress_bar,
+        # TODO: when COND_TOP_DOWN SuperAnimal models will be released - create & pass a conditions provider
```
MMathisLab, line 212: maybe already add this; we can release the mouse model now cc @AlexEMG @n-poulsen ?

MMathisLab (approved):
lgtm, but I did not test code

deeplabcut/pose_estimation_pytorch/apis/analyze_images.py
maximpavliv: This is `pcutoff` (see following line).

maximpavliv:
@n-poulsen I'd have a quick question for you:

This Pull Request addresses the issue of  `deeplabcut.analyze_images()` crashing when called with a CTD model.

To fix this, I [create and pass](https://github.com/DeepLabCut/DeepLabCut/blob/1f102b447c1970b469a865ee2d53644b555c6d83/deeplabcut/pose_estimation_pytorch/apis/analyze_images.py#L309-L322) a `cond_provider: CondFromModel` to `get_pose_inference_runner()` in `analyze_image_folder()`, similar to the [logic in `deeplabcut.analyze_videos()`](https://github.com/DeepLabCut/DeepLabCut/blob/a2680e8455a4122b3371c33b73cdaae7ba614cdd/deeplabcut/pose_estimation_pytorch/apis/videos.py#L457-L469). This fixes the issue.

However, I noticed in the `get_pose_inference_runner()`'s doctring that it should [build a `CondProvider` from the model config if `None` is passed by argument](https://github.com/DeepLabCut/DeepLabCut/blob/a2680e8455a4122b3371c33b73cdaae7ba614cdd/deeplabcut/pose_estimation_pytorch/apis/utils.py#L697-L698), which doesn't seem to be the case in [the method's implementation](https://github.com/DeepLabCut/DeepLabCut/blob/a2680e8455a4122b3371c33b73cdaae7ba614cdd/deeplabcut/pose_estimation_pytorch/apis/utils.py#L741). If `get_pose_inference_runner()` was building this conditions provider - `deeplabcut.analyze_images()` would work fine with a CTD model from the begining.

Do you rather think that:

1. The docstring is incorrect, and `get_pose_inference_runner()` not building the conditions provider itself **is the correct behavior**

or

2. I should add the construction of a `CondProvider`  if `None` is passed, using [`get_condition_provider()`](https://github.com/DeepLabCut/DeepLabCut/blob/a2680e8455a4122b3371c33b73cdaae7ba614cdd/deeplabcut/pose_estimation_pytorch/apis/ctd.py#L24) ? If it's this one - how should I get the project config (not directly accessible in [`get_pose_inference_runner()`](https://github.com/DeepLabCut/DeepLabCut/blob/a2680e8455a4122b3371c33b73cdaae7ba614cdd/deeplabcut/pose_estimation_pytorch/apis/utils.py#L672))? Construct it like `config=Path(model_config["metadata"]["project_path"])/"config.yaml"`? 😐 
Also in that case, does it make sense to also keep the current fix I added in this Pull Request in your opinion?

Thanks in advance for sharing your opinion on this 😃🙏

deeplabcut/pose_estimation_pytorch/apis/analyze_images.py
MMathisLab: cc again @AlexEMG and @n-poulsen ...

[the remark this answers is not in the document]
MMathisLab: cc @maximpavliv @yeshaokai

MMathisLab:
@maximpavliv see review above, thanks!

MMathisLab:
Status on thus PR?

maximpavliv:
@MMathisLab Ready for review and merge!

PR #3010, opened 2025-06-19
Fix CTD configs and docs inconsistencies

LucZot:
This PR:

- makes the BUCTD architecture docs consistent with the respective config files
- updates and cleans up the CTD config files.


MMathisLab (approved):
doc edits look good to me -- yamls I trust @LucZot has tested :D

AlexEMG (commented):
Great!

PR #3012, opened 2025-06-20
Shaokai/faster inference

yeshaokai:
1. Added pre-fetching to inference runner.
2. Replaced torch.no_grad() with torch.inference_mode() as it's better
3. Added automatic mixed precision inference

Async mode by default True and num_prefetch_batches by default 4.

Empirically, batch size 16 for detector and batch size 32 work well. Therefore, I ran speed testing with changes introduced in this PR, using superanimal_video_inference and a 800x600 video :

resnet50_fasterrcnn + hrnet32
12.7 FPS  -> 18 FPS

mobilenet_fasterrcnn + resnet50
25.4 FPS -> 31 FPS

ssdlite + rtmpose
25.7 FPS -> 33FPS

gpu memory usage is reduced with amp inference.


MMathisLab (approved):

deeplabcut/pose_estimation_pytorch/runners/inference.py
```diff
@@ -424,7 +579,10 @@ def predict(
                 }
             ]
         """
-        outputs = self.model(inputs.to(self.device), **kwargs)
+
+
+        
+
```
maximpavliv, line 585: This introduces a bug, `outputs` becomes undefined in the next line

MMathisLab: okay, lets write a` test` then @maximpavliv , as tests should not pass otherwise

yeshaokai: I believe the variables are still alive after leaving the with scope. Is that what you have in mind?

PR #3016, opened 2025-06-24
BUCTD GUI integration

maximpavliv:
This Pull Request adds a **widget to select conditions to the _Create training dataset_ tab in the DeepLabCut GUI**.

Screenshot from 2025-06-25 08-58-12

The widget is visible only when a CTD model is selected.
The conditions can be one of the following:
- A pytorch model snapshot (a check is applied to verify if the selected model is Bottom-Up , an warning message is displayed if that is not the case)
- Predictions stored in a _.h5_ file
- Predictions stored in a _.json_ file

The user can **choose the type of conditions** they want to use in the `QtWidgets.QFileDialog.getOpenFileName` dialog **using the filter** (in the bottom-right corner of the window).

Screenshot from 2025-06-25 09-03-42

If a CTD model is selected but conditions are missing - the shuffle creation fails (a warning message is displayed).

### DIFF
--- a/deeplabcut/gui/components.py
+++ b/deeplabcut/gui/components.py
@@ -341,12 +341,13 @@ class ConditionsSelectionWidget(QtWidgets.QWidget):
         def _shorten_path(path: str, max_length: int = 30) -> str:
             if len(path) <= max_length:
                 return path
-            return '...' + path[-(max_length - 3):]
+            return "..." + path[-(max_length - 3) :]
 
         self.selected_conditions_text.setText(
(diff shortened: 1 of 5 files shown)


MMathisLab:
hi @maximpavliv can you add a screen shot or short video of this?

maximpavliv:
> hi @maximpavliv can you add a screen shot or short video of this?

Yes, done!

AlexEMG:
For training -- it should also link to the docs & recommendations. I'd also suggest to say that it's typically preferable to use generative sampling (so no predictions during training).

MMathisLab (approved):

PR #3019, opened 2025-06-25
Shaokai/fix paf amp

yeshaokai:
The paf predictor was not part of the autocast scope, causing errors with amp.


maximpavliv (approved):
Fixes the issue, thanks @yeshaokai !

PR #3026, opened 2025-06-30
Update README.md

yeshaokai:
The top_down_hrnet_w32 topview experiment is still running.  Most experiments for top-down are done.


MMathisLab (approved):
🚀 awesome! I will merge, and we can also add links to model weights in next PR

PR #3028, opened 2025-06-30
reID revert to pre-DLC3 to stay with maDLC paper 

MMathisLab:
- tested on colab


AlexEMG (approved):

maximpavliv:
Should we also use DLC-2 in the [COLAB_3miceDemo.ipynb](https://github.com/DeepLabCut/DeepLabCut/blob/main/examples/COLAB/COLAB_3miceDemo.ipynb) ? It also uses the same demo project from Zenodo

MMathisLab:
Let's review this code ;) I am making a separate PR for that.

PR #3035, opened 2025-06-30
Set default to pm 180

AlexEMG:
We showed in the Primer that using rotations +- 180 degrees gives nice performance on the open field data even with very few annotated frames. That should be used also for PyTorch. Default is +- 30 in PyTorch, which is fine but for this dataset not ideal.


MMathisLab (approved):
lgtm!

examples/COLAB/COLAB_DEMO_mouse_openfield.ipynb
```diff
@@ -122,7 +122,28 @@
...
+    ")\n",
+    "\n",
+    "# Get the pytorch config path \n",
+    "pytorch_config_path = loader.model_folder / \"pytorch_config.yaml\"\n",
+    "\n",
+    "model_cfg = read_config_as_dict(pytorch_config_path)\n",
+    "model_cfg['data'][\"train\"][\"affine\"][\"rotation\"]=180\n",
+    "\n",
+    "# Save the modified config\n",
+    "dlc_torch.config.write_config(pytorch_config_path,model_cfg)"
```
maximpavliv, line 146: I think a slightly cleaner way would be to update the pytorch config on the run when calling `deeplabcut.train_network()` in the next cell, using the `pytorch_cfg_updates` argument:

```
deeplabcut.train_network(
    path_config_file,
    shuffle=1,
    save_epochs=5,
    epochs=200,
    batch_size=8,
    pytorch_cfg_updates={"data.train.affine.rotation": 180}
)
```

maximpavliv (changes requested):

examples/COLAB/COLAB_DEMO_mouse_openfield.ipynb
MMathisLab: I think it's better to be expressive, which is educational; i.e. this should be a real file edit vs. a passed variable (since when do we allow this anyhow? this is not in the TF API)

PR #2984, opened 2025-05-15
Improve CTD preprocessors pipeline

maximpavliv:
This Pull request

1. Adds a processing stage to the CTD preprocessors that filters out empty bounding boxes and corresponding poses. This prevents the CTD inference to crash when the BU provides a pose with 0 or 1 found bodypart, which was leading to trying to top-down crop using an empty bbox.
2. Adds a processing stage to filter out bounding boxes that are below a certain confidence level.
3. Adds unit test to test these stages.
4. Silences Runtime warning occuring during the computation of a bbox from an empty pose.

Issue #3001, opened 2025-06-07
TypeError: CondPreNet.forward() missing 1 required positional argument: 'cond_kpts'


HLH2000:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Operating System

Win11

### DeepLabCut version

3.0.0rc8

### What engine are you using?

pytorch

### DeepLabCut mode

multi animal

### Device type

RTX 3080

### Bug description 🐛

After running the video analysis process once without specifying an output destination, I attempted to run it again with destfolder=dest_folder. However, an error occurred: TypeError: CondPreNet.forward() missing 1 required positional argument: 'cond_kpts'.

### Steps To Reproduce

Initially, I ran this code, and no error messages appeared.

```python
videofile_path=str(r"C:\Users\HLH\Desktop\video")
deeplabcut.analyze_videos(
    config,
    [videofile_path],
    shuffle=CTD_SHUFFLE,
    ctd_tracking=True,
    save_as_csv=True 
)
deeplabcut.create_labeled_video(
    config,
    [videofile_path],
    shuffle=CTD_SHUFFLE,
    track_method="ctd",
    color_by="individual",
)

```
Then, I tried adjusting the parameters, and an error occurred.

```python
videofile_path=str(r"C:\Users\HLH\Desktop\video")
dest_folder = str(Path(videofile_path).parent / "custom-ctd-tracking")
deeplabcut.analyze_videos(
    config,
    [videofile_path],
    shuffle=CTD_SHUFFLE,
    destfolder=dest_folder,
    ctd_tracking=dict(
        bu_on_lost_idv=True,
        bu_max_frequency=50,
        bu_min_frequency=1,
        threshold_bu_add=0.5,
        threshold_ctd=0.01,
        threshold_nms=0.8,
    ),
)
deeplabcut.create_labeled_video(
    config,
    [videofile_path],
    shuffle=CTD_SHUFFLE,
    destfolder=dest_folder,
    track_method="ctd",
    color_by="individual",
)

```

### Relevant log output

```shell
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[7], line 4
      1 videofile_path=str(r"C:\Users\HLH\Desktop\video")
      2 dest_folder = str(Path(videofile_path).parent / "custom-ctd-tracking02")
----> 4 deeplabcut.analyze_videos(
      5     config,
      6     [videofile_path],
      7     shuffle=CTD_SHUFFLE,
[...]

PR #3040, opened 2025-07-01
Add ctd integration test

maximpavliv:
This pull request adds a CTD model to the integration-tested models in _examples/testscript_pytorch_multi_animal.py_ .

To pass the test successfully, a few fixes have been done to the CTD pipeline:

- When loading conditions from an existing _.pickle_ file, the _*_assemblies.pickle_ file is read (instead of the _*_full.pickle_)
- The CondFromFile.load_conditions_pickle() method is fixed (was not picking the conditions for the last frame)
- A fix is added to the `FilterLowConfidencePoses` `Preprocessor` stage (introduced in https://github.com/DeepLabCut/DeepLabCut/pull/2984), which was failing when the BU was making no keypoints predictions for a frame.
- The `create_video_with_all_detections()` method is fixed (previously was assuming that if predictions contained a `"bboxes"` key they also contained a `"bbox_scores"` key - this is not the case with CTD models. Similar case fixed in https://github.com/DeepLabCut/DeepLabCut/pull/2990).
- The `PadOutputs` `Postprocessor` stage is made more robust:
  - All outputs are now normalized to `np.ndarray` (previously, inputs that didn't need padding were not converted to np.ndarray)
  - The `PadOutputs` now takes an additional `expected_shapes` argument at `__init__()`, which it uses to pad correctly empty input arrays.


AlexEMG:
Let's also address #3001 here.

AlexEMG:
Can you make the tests pass?

maximpavliv:
> Can you make the tests pass?

@AlexEMG  The tests failure was caused by the current _superanimal_humanbody_ integration in main.
I now merged https://github.com/DeepLabCut/DeepLabCut/pull/3066 into this PR and the tests pass

maximpavliv:
> Let's also address #3001 here.

@AlexEMG , my latest commit should hopefully solve issue https://github.com/DeepLabCut/DeepLabCut/issues/3001 (it solves a very similar case on my machine, we will need to ask the user to confirm it also solves theirs).

_[Edit:]_ the bug raised in https://github.com/DeepLabCut/DeepLabCut/issues/3001 still happens when the [`FilterLowConfidencePoses` processing stage](https://github.com/DeepLabCut/DeepLabCut/blob/b400ad9bccbc0a992116f2f7b8a9c0844d7d141d/deeplabcut/pose_estimation_pytorch/data/preprocessor.py#L364) filters out all conditions. I opened a [separate Pull Request](https://github.com/DeepLabCut/DeepLabCut/pull/3076), which should fix this in all cases.

maximpavliv:
This PR no longer has any diffs against main because its commits were included in https://github.com/DeepLabCut/DeepLabCut/pull/3076. Closing as redundant.

PR #3062, opened 2025-07-18
Save all pytorch config updates in dlc_torch.train_network()

maximpavliv:
This Pull Request saves all updates done to the model configuration to the disk (_pytorch_config.yaml_) before starting training in `dlc_torch.train_network()`.

Previously, the pytorch config was [only saved when `pytorch_cfg_updates` were passed](https://github.com/DeepLabCut/DeepLabCut/blob/5374737dddbf201aba1ac9905626cc6e4d069be6/deeplabcut/pose_estimation_pytorch/apis/training.py#L326-L327). This prevented saving [other updated fields](https://github.com/DeepLabCut/DeepLabCut/blob/5374737dddbf201aba1ac9905626cc6e4d069be6/deeplabcut/pose_estimation_pytorch/apis/training.py#L306-L324).

### DIFF
--- a/deeplabcut/pose_estimation_pytorch/apis/training.py
+++ b/deeplabcut/pose_estimation_pytorch/apis/training.py
@@ -323,8 +323,8 @@ def train_network(
         if display_iters is not None:
             detector_cfg["train_settings"]["display_iters"] = display_iters
 
-    if pytorch_cfg_updates is not None:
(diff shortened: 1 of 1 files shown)


AlexEMG:
Thank you! Please update so it only saves if a change was made.

maximpavliv:
> Please update so it only saves if a change was made.

Done!

AlexEMG (approved):

Issue #42, opened 2018-08-14
Index Error on Step 1 EvaluateModelonDataset.py (tuple index out of range)


p-khosla:
**Describe the problem**
When running step1.evaluatemodelondataset.py the following error occurs.

Starting evaluation
Running  DeepCut_resnet_50_95shuffle1_400000forTask_movement  with # of trainingiterations: 400000
WARNING:tensorflow:From /home/dave/DeepLabCut/pose-tensorflow/nnet/pose_net.py:52: calling resnet_arg_scope (from tensorflow.contrib.slim.python.slim.nets.resnet_utils) with is_training is deprecated and will be removed after 2017-08-01.
Instructions for updating:
Pass is_training directly to the network instead of the arg_scope.
WARNING:tensorflow:From /home/dave/DeepLabCut/pose-tensorflow/nnet/pose_net.py:52: calling resnet_arg_scope (from tensorflow.contrib.slim.python.slim.nets.resnet_utils) with is_training is deprecated and will be removed after 2017-08-01.
Instructions for updating:
Pass is_training directly to the network instead of the arg_scope.
2018-08-14 17:27:04.116584: E tensorflow/stream_executor/cuda/cuda_driver.cc:406] failed call to cuInit: CUDA_ERROR_UNKNOWN
2018-08-14 17:27:04.116624: I tensorflow/stream_executor/cuda/cuda_diagnostics.cc:158] retrieving CUDA diagnostic information for host: 00e5be448947
2018-08-14 17:27:04.116636: I tensorflow/stream_executor/cuda/cuda_diagnostics.cc:165] hostname: 00e5be448947
2018-08-14 17:27:04.116688: I tensorflow/stream_executor/cuda/cuda_diagnostics.cc:189] libcuda reported version is: 384.130.0
2018-08-14 17:27:04.116722: I tensorflow/stream_executor/cuda/cuda_diagnostics.cc:369] driver version file contents: """NVRM version: NVIDIA UNIX x86_64 Kernel Module  384.130  Wed Mar 21 03:37:26 PDT 2018
GCC version:  gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.10) 
"""
2018-08-14 17:27:04.116749: I tensorflow/stream_executor/cuda/cuda_diagnostics.cc:193] kernel reported version is: 384.130.0
[...]

PR #3066, opened 2025-07-25
Superanimal humanbody: Use filtered detector

maximpavliv:
**Summary**

This Pull Request introduces a new `FilteredDetector` class, which wraps a generic object detector (typically trained on COCO) and filters its outputs to retain only detections for a specified `class_id`.

It also adds a utility function, `get_filtered_coco_detector_inference_runner()`, which returns an inference runner using a `FilteredDetector` wrapping one of the following torchvision models:
- `fasterrcnn_resnet50_fpn`
- `fasterrcnn_resnet50_fpn_v2`
- `fasterrcnn_mobilenet_v3_large_fpn`

**Main changes**

- Replaces the existing `superanimal_humanbody` inference pipeline with a `FilteredDetector`-based `DetectorInferenceRunner`.

- Solves multiple issues in the current `superanimal_humanbody` integration:

  - Fixes temporal shift in predictions
    Screenshot from 2025-07-22 12-11-52

  - Prevents CTD inference pipeline crashes (which solves [this comment](https://github.com/DeepLabCut/DeepLabCut/issues/3001#issuecomment-3065126868))
  - Resolves detector training bug where loss remained at 0.0
  - Removes unnecessary debug prints

**Structural improvements**

- Integrates `superanimal_humanbody` more cleanly into the inference flow: it is now handled via `deeplabcut.pose_estimation_pytorch.modelzoo.inference._video_inference_superanimal()`, like other PyTorch-based superanimal models.

- Exposes `FilteredDetector` as a reusable utility for users who want to focus on specific `class_ids`.

- Improves separation of concerns:
  - Preprocessor handles input data preparation
  - Postprocessor handles filtering and final outputs

**Related PR**

This PR is paired with [DLClibrary PR #42](https://github.com/DeepLabCut/DLClibrary/pull/42), which extends `dlclibrary.download_huggingface_model()` to rename the humanbody pose estimation snapshot upon download.


test_superanimal_humanbody.py
AlexEMG, line 1: Can you also integrate these additional options in the unit test that you added? tests/pose_estimation_pytorch/runners/test_filtered_detector_inference_runner.py

AlexEMG (approved):

test_superanimal_humanbody.py
maximpavliv: The checks from here that are missing in _tests/pose_estimation_pytorch/runners/test_filtered_detector_inference_runner.py_ are testing the `detector:` block in the _pytorch_config.yaml_, which I [currently don't add in my implementation](https://github.com/DeepLabCut/DeepLabCut/blob/5aeafda0458034dc578ff723bfacf043dc75fe6f/deeplabcut/pose_estimation_pytorch/modelzoo/utils.py#L119-L124) when building the _pytorch_config.yaml_. 

On the longer term, it would make sense to add the `FilteredDetector` configuration to the _pytorch_config.yaml_, but probably in an other form. Right now, _pytorch_config.yaml_ defines only DLC-native detectors. So we should carefully decide how and where to define non-DLC-native detectors in a way that is coherent and extendable. This would also leave us flexibility to support additional detector types in the future.
Once we add non-DLC-native detector configurations to the _pytorch_config.yaml_, we could implement training the pose model with _superanimal_humanbody_ as well. 

I propose to leave this for a future PR.

PR #3092, opened 2025-09-12
Fix snapshot index handling in PyTorch `analyze_videos` and extend overrides to other methods

maximpavliv:
### Summary

This PR fixes a bug in the PyTorch `analyze_videos` pipeline and extends the API for consistency across high-level methods.

### Bug Fix

- **Issue:** When calling `analyze_videos()` with non-default `snapshot_index` or `detector_snapshot_index` arguments, predictions used the correct indexes, but the tracking stages (`convert_detections2tracklets()` and `stitch_tracklets()`) always defaulted to the values from `config.yaml`. This caused crashes.

- **Fix:** Routed the argument-specified indexes through to the tracking methods, ensuring consistent behavior.

### API Extension

- Added `**torch_kwargs` to the following methods:
  - create_labeled_videos
  - create_video_with_all_detections
  - plot_trajectories
  - filterpredictions
  - extract_outlier_frames
  - analyzeskeleton

- Similarly to `analyze_videos`, these methods now support overriding `snapshot_index` and `detector_snapshot_index` via arguments, in addition to the existing `config.yaml` options.

This keeps the PyTorch backend consistent with TensorFlow while providing more flexibility to users.


arashsm79:
Tested the different methods with some videos with this PR across a bunch of projects and different snapshots for both the pose and detector models.
Everything worked fine. Thanks.
LGTM.

AlexEMG (approved):

PR #3098, opened 2025-09-17
Add optional `torch.compile` support to `InferenceRunner` via `InferenceConfig`

maximpavliv:
### Description

This PR adds optional support for `torch.compile` in the `InferenceRunner` class via the new `InferenceConfig` class (built on top of [PR #3104](https://github.com/DeepLabCut/DeepLabCut/pull/3104)).

If compilation is enabled, the model is compiled at `InferenceRunner` initialization using the specified backend. In our speed benchmarking suite, enabling `torch.compile` shows faster inference speed:
concat

However, there are some important caveats:
- Compilation may fail depending on the user’s environment or setup.
- It introduces a one-time compilation overhead during inference initialization, which may be confusing.
- Our current CI/CD tests fail when compilation is enabled (further investigation needed).

For these reasons, `torch.compile` is disabled by default, but advanced users can opt in at their own risk.

### Key changes
- Added `torch.compile` support via the `InferenceConfig` class.
- Compilation parameters can now be passed either through the `InferenceConfig` object or via the `inference:` block in `pytorch_config.yaml`, consistent with the multithreading configuration.

- When enabled, the model is compiled at initialization using the specified backend.

### Notes
- Default behavior remains unchanged for all users.
- Advanced users can enable compilation for potential inference speedups.
- This PR builds on the multithreading and InferenceConfig infrastructure introduced in [PR #3104](https://github.com/DeepLabCut/DeepLabCut/pull/3104).

### DIFF
--- a/deeplabcut/pose_estimation_pytorch/apis/utils.py
+++ b/deeplabcut/pose_estimation_pytorch/apis/utils.py
@@ -59,6 +59,7 @@ from deeplabcut.pose_estimation_pytorch.runners import (
     PoseInferenceRunner,
     TopDownDynamicCropper,
 )
+from deeplabcut.pose_estimation_pytorch.runners.inference import InferenceConfig
 from deeplabcut.pose_estimation_pytorch.runners.snapshots import (
     Snapshot,
     TorchSnapshotManager,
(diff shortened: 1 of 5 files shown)


arashsm79:
Tested this on various projects. After the initial compilation delay, the speed up is great!
Some notes for future:

- Would be nice to use the dot notation: `{"compile.enabled": False}`
- Would be better to error out if the user has input the wrong key. For example:
  ` inference_cfg={"compile": {"enble": False}}` should fail.
- Torch compile kept failing for CTD models. Maybe we should add a note for this.

maximpavliv:
Thanks for your review @arashsm79 !

As this PR no longer has any diffs against main because its commits were included in https://github.com/DeepLabCut/DeepLabCut/pull/3105, I'm closing as redundant.

I'll address your feedback in https://github.com/DeepLabCut/DeepLabCut/pull/3106 @arashsm79 .

PR #3110, opened 2025-10-02
PyTorch: Speed up single animal models

arashsm79:
This PR improves the prediction performance of various single animal PyTorch models.
The `for` loops for extracting values out of the `heatmap` tensor were replaced with advanced indexing.

 ## Details

Profiling results showed [`HeatmapPredictor.get_pose_prediction`](https://github.com/DeepLabCut/DeepLabCut/blob/33164a461c30c4144b27c5000e775472fb3a95a6/deeplabcut/pose_estimation_pytorch/models/predictors/single_predictor.py#L122) as a major hot spot in single animal inference procedure.

After this change, we see a massive speed-up in the major single animal models as shown below (thanks to @maximpavliv for running the benchmark)

  fps_vs_batchsize_128x128
  fps_vs_batchsize_256x256
  fps_vs_batchsize_512x512
</div>

### DIFF
--- a/deeplabcut/pose_estimation_pytorch/models/predictors/single_predictor.py
+++ b/deeplabcut/pose_estimation_pytorch/models/predictors/single_predictor.py
@@ -134,8 +134,8 @@ class HeatmapPredictor(BasePredictor):
 
         Example:
             >>> predictor = HeatmapPredictor(location_refinement=True, locref_std=7.2801)
-            >>> heatmap = torch.rand(32, 17, 64, 64)
-            >>> locref = torch.rand(32, 17, 64, 64, 2)
+            >>> heatmap = torch.rand(32, 64, 64, 17)
+            >>> locref = torch.rand(32, 64, 64, 17, 2)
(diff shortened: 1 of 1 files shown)


maximpavliv (approved):
Thanks a lot for this improvement — great work! 🙌
The vectorized implementation looks clean and efficient, and it passes all integration tests on my side.

Just one small suggestion: the **Example** section in the docstring should be updated for consistency with the tensor layout assumed by the code. Specifically, the example
```
>>> heatmap = torch.rand(32, 17, 64, 64)
>>> locref = torch.rand(32, 17, 64, 64, 2)
```
should be changed to
```
>>> heatmap = torch.rand(32, 64, 64, 17)
>>> locref = torch.rand(32, 64, 64, 17, 2)
```
so it matches the `(batch_size, height, width, num_joints)` format used in the implementation.

AlexEMG (approved):

AlexEMG (approved):

Issue #3108, opened 2025-09-28
PyTorch maDLC analysis crashes post-inference with `ValueError: need at least one array to stack`


tsievert:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Operating System

AlmaLinux 9.6 (Sage Margay) (HPC)

### DeepLabCut version

3.0rc10/rc12

### What engine are you using?

pytorch

### DeepLabCut mode

multi animal

### Device type

Nvidia L40S 48GB

### Bug description 🐛

After training up a PyTorch model, I ran the analysis step in order to extract outlier frames.
The analysis step run to 100% and then threw an error.
```python
100%|██████████| 1296914/1296914 [15:18:47<00:00, 23.53it/s]

Traceback (most recent call last):
  File "/home/ts/scripts/DLC_analyze_and_extract.py", line 71, in <module>
    sys.exit(main())
  File "/home/ts/scripts/DLC_analyze_and_extract.py", line 37, in main
    deeplabcut.analyze_videos(
  File "/home/ts/miniforge3/envs/dlc/lib/python3.12/site-packages/deeplabcut/compat.py", line 953, in analyze_videos
    return analyze_videos(
  File "/home/ts/miniforge3/envs/dlc/lib/python3.12/site-packages/deeplabcut/pose_estimation_pytorch/apis/videos.py", line 689, in analyze_videos
    _generate_assemblies_file(
  File "/home/ts/miniforge3/envs/dlc/lib/python3.12/site-packages/deeplabcut/pose_estimation_pytorch/apis/videos.py", line 883, in _generate_assemblies_file
    unique_bpts = np.stack(keypoint_preds[num_bodyparts:])
  File "/home/ts/miniforge3/envs/dlc/lib/python3.12/site-packages/numpy/core/shape_base.py", line 445, in stack
    raise ValueError('need at least one array to stack')

ValueError: need at least one array to stack
```
I tried with both `autotrack = true` and `false`, tried stitching the tracklets after the analysis. 
All resulted in the same error.
Even a short 2 minute clip caused the same error.

With some trial and error I figured out that the `_full.pickle` file is not written, but the `_full.pickle.dat`, `_full.pickle.dir`, `_full.pickle.bak`, and `_meta.pickle` exist.

### Steps To Reproduce

1. Train pytorch model
2. run analysis
3. see failure once 100% is reached
4. No `_full.pickle`

### Relevant log output

```shell

```

### Anything else?

[...]

PR #2861, opened 2025-02-05
ReID Tracking dataset creation for PyTorch models

n-poulsen:
Implements the creation of ReID transformer dataset creation for PyTorch pose estimation models.

- implements `deeplabcut.create_tracking_dataset` for PyTorch models
  - Closes #2794
  - Closes #2809
  - Creates a `FeatureShelfWriter` to save ReID features

Other changes

- removes the `dynamic` parameter from `deeplabcut.create_tracking_dataset` as it cannot be used there
- adds a docstring for `deeplabcut.create_tracking_dataset`
- correctly pulls the `pcutoff` from the inference config when creating assemblies in `convert_detections2tracklets`
- adds an `image_size` key to the context when running inference, and a `top_down_crop_size` for top-down models

PR #3112, opened 2025-10-06
Fix: properly use `unique_bodyparts` and `identity_scores` arguments in `ShelfWriter.add_prediction`

maximpavliv:
This PR fixes a bug introduced in **PR https://github.com/DeepLabCut/DeepLabCut/pull/2861** and reported in **Issue https://github.com/DeepLabCut/DeepLabCut/issues/3108**, where the `unique_bodyparts` and `identity_scores` arguments in `ShelfWriter.add_prediction()` were being overwritten with `None` due to redundant `kwargs.get()` calls.

These reassignments caused the method to ignore valid input arguments when passed directly. The fix removes the two lines that overwrote these parameters, restoring the correct behavior while keeping `**kwargs` in the signature for subclass compatibility.


arashsm79:
LGTM. Thanks!

deruyter92 (approved):

C-Achard:
Maybe add a unit test for this specifically @deruyter92 ?

deruyter92:
Added unit tests, which revealed another bug in writing metadata. Now addressed that as well. 

`ShelfWriter.close()` did `self._db["metadata"]["nframes"] = self._frame_index`, which modifies a deserialized copy of the dict. This is now fixed.

PR #3117, opened 2025-10-09
PyTorch: Speed up PAF cost computation

arashsm79:
### Summary
Improve PAF performance by performing affinity computation on the GPU with advanced indexing.

- Affinities are now calculated using torch operations.
- The cost per batch dictionary is created more efficiently.

### Details
This implementation tries to delegate the parts that can be parallelized to the GPU by using torch operations instead of numpy ones. (thanks to @maximpavliv for running the benchmark)

  fps_vs_batchsize_128x128
  fps_vs_batchsize_256x256
  fps_vs_batchsize_512x512
</div>

The figure below shows the parts of the execution that can be optimized.
The part outlined by the red rectangle concerning `compute_peaks_and_costs` is now as optimized as I could make it.
The blue rectangle is concerned with the assembly procedure which did not get into in this PR. There is a lot of room for optimization in there as well, which may require a lot of refactoring/changes.

image

### DIFF
--- a/deeplabcut/pose_estimation_pytorch/models/predictors/paf_predictor.py
+++ b/deeplabcut/pose_estimation_pytorch/models/predictors/paf_predictor.py
@@ -130,9 +130,9 @@ class PartAffinityFieldPredictor(BasePredictor):
             >>> stride = 8
             >>> poses = predictor.forward(stride, output)
         """
-        heatmaps = outputs["heatmap"] # (batch_size, num_joints, height, width)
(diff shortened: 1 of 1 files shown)


maximpavliv (approved):
Good job! ✅

- Code is clearer, docstrings are much more detailed, and variable naming is improved (`batch_size`, `paf_limb_inds`).
- GPU usage is now more efficient, avoiding unnecessary early CPU transfers → speed is improved.
- The inference results slighly differ from expected results.

AlexEMG (approved):

PR #3121, opened 2025-10-10
PyTorch: Speed up DEKR predictor

arashsm79:
### Summary
This PR improves the performance of DEKR predictor. 

- Use advanced indexing instead of `for` loops.
- Remove optimization TODOs.
- Optimize both with and without the use of heatmap

### Details
Profiling the DEKR predictor's forward call, shows a significant increase in speed. But the overall inference time of the architecture remains mostly unchanged (very minor improvements) since it is largely dominated by HRNet and torch native modules.  (thanks to @maximpavliv for running the benchmark)

  fps_vs_batchsize_128x128
  fps_vs_batchsize_256x256
  fps_vs_batchsize_512x512
</div>

Below shows (the rectangle in blue) that most of the inference time is spent in the HRNet and PyTorch native inference procedures.
I have optimized the rest of it (the rectangle in red) as much as I could.

image

### DIFF
--- a/deeplabcut/pose_estimation_pytorch/models/predictors/dekr_predictor.py
+++ b/deeplabcut/pose_estimation_pytorch/models/predictors/dekr_predictor.py
@@ -139,7 +139,7 @@ class DEKRPredictor(BasePredictor):
         pose_ind, ctr_scores = self.get_top_values(center_heatmaps)
 
         posemap = posemap.permute(0, 2, 3, 1).view(batch_size, h * w, -1, 2)
-        
+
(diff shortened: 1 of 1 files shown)


maximpavliv (approved):
The refactoring looks great — the vectorized version is much cleaner.
I’ve run the full integration testing suite on my side, and everything passes without issues. Nice work on this improvement!

deruyter92 (commented):
I ran the code step-by step and confirmed that evaluation results are close to previous implementation. The vectorization looks great, also happy that you added the tensor shapes in the comments. Looks all good to me!

PR #3129, opened 2025-10-18
CI: Increase timeout in multiprocessing test

arashsm79:
We sometimes get spurious failures on Windows CI because of too short a timeout in one of the tests (rerunning the check usually fixes it).

This PR increases the timeout to avoid these sporadic failures.


deruyter92:
Closing this PR since the issue was addressed in PR #3142, skipping the test.

PR #3147, opened 2025-11-15
Add method to export PyTorch DeepLabCut model to ONNX

Yu-changqing:
This script provides a straightforward and reliable method to export a DeepLabCut PyTorch pose estimation model to ONNX.


AlexEMG:
Excellent! Do you also have code for running inference?

Yu-changqing:
> Do you also have code for running inference?

Yeah, I’ve been writing some inference code for Dlc，but it’s still a bit rough at the moment.

PR #3159, opened 2025-12-06
Avoid redundant superanimal colormap calls during pytorch inference

juan-cobos:
Simply placing colormap calls outside the for loop of video paths and create output_video when create_labeled_video is True.


deruyter92 (approved):
@juan-cobos, thanks for changing this. Looks good.

Issue #3130, opened 2025-10-20
Frames not correctly labeled in video generation


Jiachen24:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Operating System

Windows 11 Pro

### DeepLabCut version

3.0.0 rc6

### What engine are you using?

pytorch

### DeepLabCut mode

single animal

### Device type

RTX 4060

### Bug description 🐛

Hi! I used jupyter notebook to generate high quality video for an analysed data. The generated frames don't look correctly labeled. However, when I check the frames using labeling GUI, the labels look great. I attached an example of same frame, one is saved by video generation and the other is what it looks in the labeling GUI. I used the same model and I didn't crop the video when I analysed it. 

### Steps To Reproduce

import deeplabcut

config_path = "F:/Deeplabcut/projects/CSA_tongue-Jiachen-2025-09-26/config.yaml"

videos = [
   "E:/video/135/135_tongue_day2.mp4"]

deeplabcut.create_labeled_video(
    config_path, 
    videos, 
    shuffle = 3, 
    filtered = True, 
    pcutoff = 0.75,
    save_frames = True)

### Relevant log output

```shell
Loading DLC 3.0.0rc6...
C:\Users\Limlab\anaconda3\envs\deeplabcut\lib\site-packages\tqdm\auto.py:21: TqdmWarning: IProgress not found. Please update jupyter and ipywidgets. See https://ipywidgets.readthedocs.io/en/stable/user_install.html
  from .autonotebook import tqdm as notebook_tqdm
Starting to process video: E:/video/135/135_tongue_day2.mp4
Loading E:/video/135/135_tongue_day2.mp4 and data.
Duration of video [s]: 3959.55, recorded with 30.01 fps!
Overall # of frames: 118843 with cropped frame dimensions: 1200 1200
Generating frames and creating video.
100%|██████████| 118843/118843 [14:53:23<00:00,  2.22it/s]  
Labeled video E:\video\135\135_tongue_day2DLC_HrnetW32_CSA_tongueSep26shuffle3_snapshot_280_filtered_p75_labeled.mp4 successfully created.
```

### Anything else?

<img width="1200" height="1200" alt="Image" src="https://github.com/user-attachments/assets/84b4b35e-bce7-439a-9820-cf2e3df5c70a" />

<img width="1919" height="947" alt="Image" src="https://github.com/user-attachments/assets/e9cc762e-645f-4622-934c-ff97b2a46e61" />

[...]

Issue #3060, opened 2025-07-16
DLC v3.0.0rc10 extracted outlier frames do not maintain marked absolute positions due to cropping setting


QiongBan:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Operating System

Windows 11 Pro

### DeepLabCut version

3.0.0rc10

### What engine are you using?

pytorch

### DeepLabCut mode

single animal

### Device type

NVIDIA RTX A6000

### Bug description 🐛

Hello, I was using GUI defined crop params for first extraction of frames at the beginning of the project. After the video analysis and creation were finished, I set the cropping to True for analysis and outlier frame detection in config file and used the coords I got from those params through GUI. However, the returned manual extracted and automatically extracted frames all have deviations although the relative spatial relationships are retained. I am afraid that editing these would result in contamination of training results.
When disabled cropping in config file, I was able to get frames with marks that stay at the correct places.

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/DeepLabCut/DeepLabCut/blob/master/CODE_OF_CONDUCT.md)

deruyter92:
Hi @QiongBan, sorry for the late response. Are you still encountering this issue? Could you maybe share a minimal example with us to reproduces your error? If you could share a step-by-step instruction of what you did, that would be very helpful for us to locate the cause of the problem. Thanks in advance!

Issue #1979, opened 2022-09-10
Labels on labeled video placed from cropped coordinates


jonpedros:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Bug description

Hello everyone,

I'm having an issue with the create_labeled_video function. Currently, I've been analyzing novel videos by setting cropping coordinates for each file separately. I do this because I labeled frames cropped around the eye and now pupil border detection performance is perfect this way (but quite finicky when using the whole frame or even with dynamic cropping).

However, when creating the labeled videos, the labels are placed according to the cropped frames label coordinates, as you can see here. 

![vlcsnap-2022-09-10-09h38m51s478](https://user-images.githubusercontent.com/85462635/189463319-f14d4c76-2f11-4daf-8ce3-5da3a275c47d.png)

Here is a frame from a labeled video created with `displaycropped=True`

![vlcsnap-2022-09-10-09h38m55s898](https://user-images.githubusercontent.com/85462635/189463340-63bc031a-895a-4327-ab13-9afd7a389e9d.png)

The issue doesn't happen with no cropping or dynamic cropping.

I found [this](https://forum.image.sc/t/madlc-video-analysis-with-cropping-wrong-labeling/43671) discussion on image.sc, in which the user solved the exact same problem by using `cropping=False` on the config.yaml file. I checked this boolean is false for my file, so that shouldn't be an issue. 

Lastly, I noticed that, when using `save_frames=True`, the created frame_temp folder has all the files with the correct label placing, although I haven't been able to check the finished video for this mode because the file is not written (but with no error output) and only a few frames remain in the temporary folder.

### Operating System

Windows 10 (Native)

### DeepLabCut version

DLC 2.2.2

### DeepLabCut mode

single animal

### Device type

NVIDIA GeForce RTX 2600 SUPER

### Steps To Reproduce

1 - Use network trained from cropped frames
2 - Analyze novel video with cropping parameters
3 - Create labeled video

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/DeepLabCut/DeepLabCut/blob/master/CODE_OF_CONDUCT.md)

jeylau:
Since individual frames are stored temporarily, I assume you've passed `save_frames=True`; does the problem remain if it's left to `False` (so that the fast video creation mode is used)?

jonpedros:
[...]

PR #3161, opened 2025-12-18
cropping-related fixes in `analyze_videos` and `create_labeled_video`

deruyter92:
This PR addresses some functional vulnerabilities in the code for creating labeled videos. Potentially related to #3130, #3060 and #1979.

The following changes are implemented:
1. **Fix disfunctional evaluation of config in gui.tabs.analyze_videos**. In the previous code, the configuration value for `cropping` is misinterpreted because the conditional compares to a string value instead of a boolean. This is now changed to boolean evaluation.
2. **Ambivalous interpretation of argument `cropping=None`**. 
In most places, `None` is passed to reflect that no cropping should be applied, but in pytorch version of `analyze_videos`, `None` resulted in cropping coordinates loaded from config.yaml. This PR changes the behaviour as follows: The default argument is changed to `cropping='from_config'` which follows prior behavior; `cropping=None` is now interpreted as no cropping to align with its interpretation elsewhere, `cropping=[x1, x2, y1, y2]` is still interpreted as cropping with the specified parameters.

3. **Crop margins are always added to the bounding box coordinates when plotting in CreateVideo**, even when the output is a cropped video and the coordinates therefore already reflected the correct values. This is a bug that is now fixed in this PR.  Margins should only be added when using coordinates from a cropped video to generate the full video. 

4. **When metadata specifies cropping=False, this is still interpreted as cropping=True** when the crop parameters don't match the video dimensions. Although this should normally cause no problems, it would be better to always follow the the value specified in metadata. This is now corrected in this PR.

### DIFF
--- a/deeplabcut/compat.py
+++ b/deeplabcut/compat.py
@@ -12,7 +12,7 @@
 from __future__ import annotations
 
 from pathlib import Path
-from typing import Iterable
+from typing import Iterable, Literal
 
 import numpy as np
 from ruamel.yaml import YAML
@@ -666,7 +666,7 @@ def analyze_videos(
     in_random_order: bool = True,
     destfolder: str | None = None,
     batchsize: int = None,
-    cropping: list[int] | None = None,
+    cropping: list[int] | None | Literal['from_config'] = 'from_config',
     TFGPUinference: bool = True,
     dynamic: tuple[bool, float, int] = (False, 0.5, 10),
     modelprefix: str = "",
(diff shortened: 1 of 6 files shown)


deruyter92:
This PR will be replaced by a broader-scoped refactoring task to restructure the keypoint handling. The PR is kept open as a placeholder until those are started. Changed the label to 'not planned' for now as this specific PR is not planned to be merged.

Issue #3164, opened 2026-01-04
Disabling excessive pop-ups in Analyze videos


gianmarco-terrones:
**Is your feature request related to a problem? Please describe.**
In the Analyze videos tab of the GUI, when one selects the checkboxes **Plot trajectories** and **Show trajectory plots** and chooses a set of body parts, four pop-up plots appear once the **Analyze Videos** button is clicked. These plots have the titles Figure 1, Figure 2, Figure 3, and Figure 4. 

<img width="929" height="670" alt="Image" src="https://github.com/user-attachments/assets/00364e35-2b47-40f8-9ba5-93f13d382889" />

<img width="579" height="499" alt="Image" src="https://github.com/user-attachments/assets/99a72acf-d95b-41a8-bf00-5f9c20f4ea37" />

However, if the **Analyze Videos** button is pressed again and **Show trajectory plots** is still checked, 8 figures pop up. If the button is pressed again, 12 figures pop up, and so on.

**Describe the solution you'd like**
It would be great if whenever **Analyze Videos** button is pressed and **Show trajectory plots** is selected, only four plots pop up, corresponding to the current selection of body parts.

**Describe alternatives you've considered**
This isn't an alternative per se, more a proposed solution. Add the following in line 307 of deeplabcut/utils/plotting.py, in the function `plot_trajectories`:
```
    # Close existing figures if showing new ones to prevent accumulation
    # when the Analyze Videos button is pressed multiple times
    if showfigures:
        plt.close("all")
```

**Additional context**
The sets of four plots that appear correspond to hist.png, plot.png, plot-likelihood.png, and trajectory.pn in videos/plot-poses of the opened DLC project.

deruyter92:
Hi @gianmarco-terrones, this sounds reasonable to me. I have created PR #3168 for your proposed feature, please try it if you want.

deruyter92:
This issue was resolved in PR #3168

PR #3168, opened 2026-01-06
Close figures before plotting

deruyter92:
Closing all plots before plotting new ones. This avoids overflowing the screen with old plots.

This commit changes behavior from only closing figures when `showfigures=False` to always closing figures before plotting. Addresses feature request #3164.


deruyter92:
@MMathisLab, this PR is a matter of design choice / preferred behavior for the `plot_trajectories` function. 
- Current behavior: all figures stay open, you have to manually close previous plots if you don't want to see them. 
- Merging this PR -> all figures close before plotting. Only the fresh plots will be visible when plotting trajectories. 

Both are valid options.

MMathisLab (approved):

PR #3175, opened 2026-01-12
Refactor list_videos_in_folder for clarity and correctness

juan-cobos:
Refactors `list_videos_in_folder` to improve code clarity, fix bugs, and streamline logic:

**Key changes:**
- **Moved suffix normalization out of loop**: Video suffixes are now computed once before iteration instead of being recalculated for each directory
- **Replaced list with set for O(1) lookups**: Changed `video_suffixes` from list to set for more efficient suffix matching
- **Simplified suffix normalization**: Consolidated duplicate normalization logic using `lstrip('.')` instead of conditional checks
- **Added file validation for non-directory paths**: Individual files are now validated against `video_suffixes` before being added to results
- **Improved error handling**: Moved `path.exists()` check before path type determination to catch errors earlier
- **Fixed shuffle behavior**: Shuffle now applies to the complete video list instead of per-directory, ensuring consistent behavior
- **Enhanced type hints**: Added `Path` to `data_path` type annotation for better type safety

These changes reduce code duplication, improve performance with set-based lookups, and ensure only valid video files are returned.

### DIFF
--- a/deeplabcut/pose_estimation_pytorch/apis/utils.py
+++ b/deeplabcut/pose_estimation_pytorch/apis/utils.py
@@ -344,7 +344,15 @@ def list_videos_in_folder(
             videos.append(path)
 
     # Remove duplicates while preserving order (using resolved paths to handle symlinks)
-    videos = list({v.resolve(): v for v in videos}.values())
+    seen = set()
+    unique_videos: list[Path] = []
+    for v in videos:
+        resolved = v.resolve()
+        if resolved in seen:
+            continue
+        seen.add(resolved)
+        unique_videos.append(v)
+    videos = unique_videos
 
     if shuffle:
         random.shuffle(videos)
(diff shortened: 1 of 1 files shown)


deruyter92 (approved):
Thanks for addressing this! Looks good.

Since we are adjusting this function now anyway, I pushed some additional refactors.
- added case normalization (match lowercases)
- added deduplication of paths
- changed assertion to FileNotFoundError.

deruyter92:
@juan-cobos thanks! 3c3d0c4 is exactly the opposite from da1ccd7 right? Is that on purpose? (see copilot suggestion) 
Both solutions are fine with me. 

The rest of the code looks good to me.

juan-cobos:
@deruyter92 I understand removing duplicates, although I’m not sure how likely that case is (it wasn’t checked before). What I don’t quite follow is keeping unresolved paths, since it's is effectively arbitrary (you'd just keep the first one).

deruyter92:
agreed, not worried about the order and path.resolve raising potential errors is actually preferred behavior. 

thanks for changing it back. Looks all good to me!

Issue #3176, opened 2026-01-12
Append, not overwrite `learning_stats.csv` when continuing training from checkpoint [minor request]


guptadivyansh:
**Is your feature request related to a problem? Please describe.**
By looking at the (very useful) test metrics in `learning_stats.csv`, I decided to continue training my network for some more epochs. I wanted to plot the validation RMSE curves, but as soon as I restarted training, the csv only had the new epochs and the old metrics were gone.  

**Describe the solution you'd like**
The CSV logger should just append more epochs as rows when training is continued, rather than creating (and thus overwriting) a 'new' file. 

**Describe alternatives you've considered**
Parsing the lost metrics from the raw logs in 'train.txt'. Fortunately, 'train.txt' already opens the file in append mode, so the logging continues with retraining. The learning_stats.csv is obviously much nicer to parse though.

**Additional context**
dlc version 3.0.0rc13 using pytorch models.

deruyter92:
@guptadivyansh, thanks for your request. We have been working on a fix for this issue (see PRs listed above this comment). Please try it out and let us know if it solves your problem.

deruyter92:
This issue was resolved in https://github.com/DeepLabCut/DeepLabCut/pull/3177. Thanks again for raising the concern.

PR #3177, opened 2026-01-13
Fix CSVLogger: read existing learning_stats.csv file when resuming previous training

deruyter92:
Previously, when resuming training the CSVLogger created a new learning_stats.csv file for the second training run, overwriting previously saved training metrics.

This commit addresses feature request #3176 and changes the behaviour to reading any previously saved log metrics and appending new epochs to the learning_stats.csv file.

- added `_load_existing_data` method to CSVLogger
- added unit test for `_load_existing_data`

### DIFF
--- a/deeplabcut/pose_estimation_pytorch/runners/logger.py
+++ b/deeplabcut/pose_estimation_pytorch/runners/logger.py
@@ -459,21 +459,21 @@ class CSVLogger(BaseLogger):
         try:
             with open(self.log_file, "r", newline="") as f:
                 reader = csv.DictReader(f)
-                
+
                 # Update logged metrics from header
                 if "step" not in reader.fieldnames:
                     return  # Invalid format
-                    
+
                 metric_names = [m for m in reader.fieldnames if m != "step"]
                 self._logged_metrics.update(metric_names)
-                
+
                 # Load data rows
                 for row in reader:
                     try:
                         step = int(row["step"])
                     except (ValueError, KeyError):
                         continue
-                    
+
                     # Convert metric values: empty strings -> None, numeric strings -> float
                     step_metrics = {}
                     for metric in metric_names:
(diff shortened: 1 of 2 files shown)


MMathisLab:
closed in favor of #3179

C-Achard (approved):

C-Achard:
Looks good ! Should we consider adding some more validation in case the previous output is somehow broken, interrupted or to try and account for potential formatting changes in the future ? Or do you feel that would be redundant for now/the current solution is enough ?

deruyter92:
> Should we consider adding some more validation in case the previous output is somehow broken, interrupted or to try and account for potential formatting changes in the future ?
Good thought!
- The new _load_existing_data method is agnostic to the specific columns in the existent CSV. If they are different from the new training run, this doesn't break the CSV file (missing fields are just kept empty)
- The only required field is `step`,  containing an integer value. There are no further constraints on what the existing data should look like. 
- I have adjusted the behavior now, based on your suggestion: In case any exception occurred during reading (e.g. corrupted CSV, or CSV that misses 'step' column), the reader now starts with an empty CSV instead of keeping the succeeded rows from the existent CSV. 

I don't think it would be a good idea to put specific constraints on the values or data types that should be in the existent CSV, except for the 'step' index. The CSVLogger is currently quite flexible and even logs different metrics for different epoch types (e.g. validation epochs). I think the benefits are limited when we drop this flexible/column-agnostic behavior of the CSVLogger.

C-Achard:
@deruyter92 Sounds good, thanks !

C-Achard (approved):

deruyter92:
Thanks @C-Achard! Ok, @MMathisLab, please have a look if you agree. In that case these changes can be merged!

deruyter92:
@MMathisLab let me know what you think! I think it could be merged.

PR #3185, opened 2026-01-22
Keyword-only `key` argument for pd.DataFrame.to_hdf()

deruyter92:
**Motivation:**
Pandas officially released version 3.0 (January 2026), but our code is not compatible with this version.
- Positional usage of the `key` argument in pd.DataFrame.to_hdf() is no longer be supported. Pandas 2.2 and higher raise a warning for keyword-only usage of the `key` parameter, and raise an error for pandas version 3.0. This was already implemented in most of the code, but this PR addresses the remaining places where that was not the case. 

**Changes:**
This PR replaces `df.to_hdf(dataname, "df_with_missing", ...)` with `df.to_hdf(dataname, key="df_with_missing", ...)` (and similar) where that was not yet implemented.


deruyter92:
Minimal changes. Should help to prevent issues when users have a wrong pandas installation (also prevented by our restricted requirements, but good to address anyway).

C-Achard (approved):
Should be fine, very low risk

PR #3190, opened 2026-01-29
[dev] C1 - centralize project config I/O and add testing

deruyter92:
This PR is part of the WIP for migrating from dictionary configs to typed & validated configurations (see #3193 for an overview). 

**Goal of this PR:**
Centralize the logic for reading, writing and adjusting configurations. And improve test coverage of the current behavior (before changing to a different configuration setup).
 
**Motivation:**
Currently reading and writing yaml files, and the logic for adjusting configuration files is spread across different places in the repository. 
- Keeping the input/output and config adjustment logic in a centralized place facilitates easier maintenance / migration, better testing and makes it clear for users which methods are available. 
- Before making larger adjustments, such as typed configurations, or different field names, it is important to capture the current behavior using tests. 

**Changes**
 1. Move all reading and writing functions for yaml files to `core.config`
 2. exporting to `auxiliaryfunctions` to keep existing references in place for backwards compatibility
 3. add pytests for core.config and the modelzoo `generalized_data_converter`
 
 This PR only pertains to project configs and core yaml reading / writing. The pose configs in pytorch pose estimation are kept unchanged.

### DIFF
--- a/deeplabcut/core/config.py
+++ b/deeplabcut/core/config.py
@@ -11,7 +11,6 @@
 """Centralized helpers for reading, writing, and creating configuration files (YAML)."""
 from __future__ import annotations
 
-import os
 import warnings
 from pathlib import Path
 from typing import Callable
@@ -34,7 +33,7 @@ def read_config_as_dict(config_path: str | Path) -> dict:
     Raises:
         FileNotFoundError: if the config file does not exist
     """
-    if not os.path.exists(config_path):
+    if not Path(config_path).exists():
         raise FileNotFoundError(
             f"Config {config_path} is not found. Please make sure that the file exists."
         )
@@ -264,7 +263,7 @@ def read_config(configname: str | Path) -> dict:
     """
     ruamelFile = YAML()
     path = Path(configname)
-    if os.path.exists(path):
+    if path.exists():
         try:
             with open(path, "r") as f:
                 cfg = ruamelFile.load(f)
--- a/deeplabcut/core/config.py
+++ b/deeplabcut/core/config.py
@@ -8,13 +8,19 @@
 #
 # Licensed under GNU Lesser General Public License v3.0
 #
-"""Simple helper methods related to configuration files stored in yaml files"""
+"""Centralized helpers for reading, writing, and creating configuration files (YAML)."""
 from __future__ import annotations
 
+import os
+import warnings
 from pathlib import Path
 from typing import Callable
 
+import yaml
+import ruamel.yaml.representer
 from ruamel.yaml import YAML
+
+from deeplabcut.core.engine import Engine
 
 
 def read_config_as_dict(config_path: str | Path) -> dict:
@@ -24,7 +30,14 @@ def read_config_as_dict(config_path: str | Path) -> dict:
 
     Returns:
         The configuration file with pure Python classes
+
+    Raises:
+        FileNotFoundError: if the config file does not exist
     """
+    if not os.path.exists(config_path):
+        raise FileNotFoundError(
+            f"Config {config_path} is not found. Please make sure that the file exists."
+        )
     with open(config_path, "r") as f:
         cfg = YAML(typ="safe", pure=True).load(f)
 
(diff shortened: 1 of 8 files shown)


deeplabcut/core/config.py
```diff
@@ -24,7 +30,14 @@ def read_config_as_dict(config_path: str | Path) -> dict:
 
     Returns:
         The configuration file with pure Python classes
+
+    Raises:
+        FileNotFoundError: if the config file does not exist
     """
+    if not os.path.exists(config_path):
```
C-Achard: If we're importing  pathlib I'd be in favor of not using os for filepath handling

deeplabcut/modelzoo/generalized_data_converter/datasets/materialize.py
```diff
@@ -117,8 +117,7 @@ def __init__(self):
 
     def create_cfg(self, proj_root, kwargs):
         self.cfg.update(kwargs)
-        with open(os.path.join(proj_root, "config.yaml"), "w") as f:
-            yaml.dump(self.cfg, f)
+        write_project_config(os.path.join(proj_root, "config.yaml"), self.cfg)
```
C-Achard: Suggested : use pathlib instead rather than a mix

deeplabcut/utils/auxiliaryfunctions.py
```diff
@@ -28,277 +28,28 @@
...
-import ruamel.yaml.representer
-import yaml
-from ruamel.yaml import YAML
 
+from deeplabcut.core import config as _config
 from deeplabcut.core.engine import Engine
 from deeplabcut.core.trackingutils import TRACK_METHODS
 from deeplabcut.utils import auxfun_videos, auxfun_multianimal
 
+# NOTE JR 2026-01-29: Configuration I/O is now centralized in deeplabcut.core.config
```
C-Achard: Small comment but I realized with the GitHub VSCode extension you can @ yourself and others
```suggestion
# NOTE @deruyter92 2026-01-29: Configuration I/O is now centralized in deeplabcut.core.config
```

deeplabcut/utils/auxiliaryfunctions.py
```diff
@@ -28,277 +28,28 @@
 
 import numpy as np
 import pandas as pd
-import ruamel.yaml.representer
-import yaml
-from ruamel.yaml import YAML
 
+from deeplabcut.core import config as _config
```
C-Achard: Just for readability, there's lots of "config" in this file
```suggestion
from deeplabcut.core import config as DLC_config
```

deeplabcut/core/config.py
```diff
@@ -72,3 +84,294 @@ def pretty_print(
...
+                    write_project_config(configname, cfg)
+        except Exception as err:
+            if len(err.args) > 2:
+                if (
+                    err.args[2]
+                    == "could not determine a constructor for the tag '!!python/tuple'"
+                ):
+                    with open(path, "r") as ymlfile:
+                        cfg = yaml.load(ymlfile, Loader=yaml.SafeLoader)
+                        write_project_config(configname, cfg)
```
C-Achard, line 293: @deruyter92 When is this expected to occur ?

deeplabcut/modelzoo/generalized_data_converter/datasets/materialize.py
```diff
@@ -108,7 +108,7 @@ def __init__(self):
         x2 = 640
         y1 = 277
         y2 = 624
-        corer2move2 = [50, 50]
+        corner2move2 = [50, 50]
```
C-Achard, line 111: @deruyter92 Does this require any sort of backwards compatibility checks ? This might mean old configs were created with the wrong field name, is that correct ?

deeplabcut/modelzoo/generalized_data_converter/datasets/materialize.py
```diff
@@ -159,7 +158,7 @@ def __init__(self):
...
```
C-Achard, line 161: Same comment as before, is this fine for back. comp.  ?

deeplabcut/utils/auxiliaryfunctions.py
```diff
@@ -28,277 +28,28 @@
...
-        for key in edits:
-            cfg.pop(key)
-        write_plainconfig(output_name, cfg)
-    return cfg
+def write_plainconfig(configname: str | Path, cfg: dict, overwrite: bool = True) -> None:
+    """Write a config dict to YAML (alias for write_config). See deeplabcut.core.config."""
+    core_config.write_config(config_path=configname, config=cfg, overwrite=overwrite)
+edit_config = core_config.edit_config
+write_config_3d = core_config.write_config_3d
+write_config_3d_template = core_config.write_config_3d_template
```
C-Achard, line 59:
```suggestion
create_config_template = core_config.create_config_template
create_config_template_3d = core_config.create_config_template_3d
read_config = core_config.read_config
write_config = core_config.write_project_config
edit_config = core_config.edit_config
write_config_3d = core_config.write_config_3d
write_config_3d_template = core_config.write_config_3d_template


def read_plainconfig(configname: str | Path) -> dict:
    """Load a YAML config (alias for read_config_as_dict). See deeplabcut.core.config."""
    return core_config.read_config_as_dict(config_path=configname)


def write_plainconfig(configname: str | Path, cfg: dict, overwrite: bool = True) -> None:
    """Write a config dict to YAML (alias for write_config). See deeplabcut.core.config."""
    core_config.write_config(config_path=configname, config=cfg, overwrite=overwrite)

```

tests/core/config/test_core_config.py
```diff
@@ -0,0 +1,314 @@
...
+def test_read_config_as_dict_loads_yaml(tmp_path):
+    config_path = tmp_path / "config.yaml"
+    config_path.write_text("a: 1\nb: [2, 3]\nc:\n  d: 4\n")
+    cfg = read_config_as_dict(str(config_path))
+    assert cfg == {"a": 1, "b": [2, 3], "c": {"d": 4}}
+
+def test_read_config_as_dict_raises_when_file_missing():
+    with pytest.raises(FileNotFoundError):
+        read_config_as_dict(Path("/nonexistent/config.yaml"))
+
```
C-Achard, line 43: This is good; should there also be testing for missing/misnamed/broken fields ? Or only once we expand config validation efforts

tests/pose_estimation_pytorch/modelzoo/test_generalized_data_converter_config.py
```diff
@@ -0,0 +1,175 @@
...
+    "multianimalproject": False,
+    "uniquebodyparts": [],
+    "x1": 0,
+    "x2": 640,
+    "y1": 277,
+    "y2": 624,
+    "corner2move2": [50, 50],
+    "move2corner": True,
+    "identity": False,
+}
```
C-Achard, line 66: Does this need to be re-defined or could we import it from elsewhere ? Maybe there is a good reason to have this here but if not I'd try to define this in the code and grab it from there

C-Achard (commented):
Very good, I hope my above comments make sense; if they do need further action I'll wait a bit before approving

deeplabcut/core/config.py
deruyter92: I just moved it along from the auxiliaryfunctions and keeping the original implementation for now. Aiming to deprecate the old `read_config` functionality in favor of constructing it using the ProjectConfig class in PR #3194. 

I will add a test for PyYAML-specific tag !!python/tuple and see if it is handled correctly. Thanks for pointing it out

**Edit:** 
- The yaml loading is inconsistently implemented (sometimes using a safeloader, sometimes allowing constructors). When consistently using safeloaders, the reading should actually break. 
- This try/exception logic seems to be incorrectly implemented. In fact, the solution inside the exception block is expected to raise a parsing Error, while the part in the try block should work without a problem (it is unsafely loading yaml files with python tags).
- I have implemented a test, that is skipped for now, to capture the current behavior without changing it. The test can be unskipped once the behavior is improved (e.g. safeloading everywhere)

PR #3191, opened 2026-01-29
[dev] C2 - Add typed configs as pydantic dataclasses & omegaconf dictconfigs

deruyter92:
This PR is part of the WIP for migrating from dictionary configs to typed & validated configurations (see #3193 for an overview). (Credits to @arashsm79 for originally introducing these changes).

**Motivation**
Introduces typed, validated config classes (Pydantic + OmegaConf) and a shared `ConfigMixin` for project and pose configs. Replaces ad-hoc dicts with a consistent load/save/validation API and better type support.

**Changes:**
- Core:  `ConfigMixin`, `ProjectConfig`, and utils (moved from `core/config.py`). New `core/types.py` with `PydanticNDArray`.
- Weight init: `WeightInitialization` is a Pydantic dataclass using `ConfigMixin`; `snapshot_path` optional; `conversion_array` typed as `PydanticNDArray`.
- Pose PyTorch: New typed configs in `pose_estimation_pytorch/config/` (data, inference, logger, model, pose, project, runner, training). `GenSamplingConfig` moved from `data/generative_sampling.py` to `config/data.py`. Inference config classes moved from `runners/inference.py` to `config/inference.py`; inference runner uses `InferenceConfig.from_any()`.
- Deps: `omegaconf>=2.3.0`, `pydantic>=2.12.3`.
- Tests: `tests/core/config/test_config_mixin.py` for `ConfigMixin`.

**Notes:**
- The typed configurations are not yet used in the pipeline, but implemented in parallel as part of the work-in-progress.
- The typed configurations in this PR mimic the structure of the older dictionary configs (same fields) to facilitate easy migration. These will be updated later.


deeplabcut/pose_estimation_pytorch/config/project.py
```diff
@@ -0,0 +1,133 @@
...
+    # Project path
+    project_path: str = ""
+    # Engine
+    engine: str = "pytorch"
+
+    # Annotation data set configuration
+    video_sets: dict[str, Any] = field(default_factory=dict)
+    # VV TODO @deruyter92 2026-01-30: following the old original config.yaml template for now. VV
+    # VV We should change this to a list[str] in the future. VV
+    bodyparts: list[str] | str = 'MULTI!'
```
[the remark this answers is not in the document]
deruyter92: agreed, this will be addressed in the future. In this stage I want the new typed configurations to identically match the old dictionary configurations, including inconsistencies.

deeplabcut/pose_estimation_pytorch/config/data.py
```diff
@@ -0,0 +1,187 @@
...
+    loader: DLCLoaderConfig | COCOLoaderConfig | None = Field(
+        default=None, discriminator="type"
+    )
+
+    @field_validator("train", "inference", mode="before")
+    @classmethod
+    def validate_transforms(cls, v):
+        from deeplabcut.pose_estimation_pytorch.data import build_transforms
+
+        build_transforms(v)
```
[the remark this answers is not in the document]
deruyter92: actually, the return value should be `v` (a dictionary), not `build_transforms(v)` (the composed transforms). This is now resolved.

deeplabcut/core/config/project_config.py
```diff
@@ -0,0 +1,162 @@
...
+
+from deeplabcut.core.config.config_mixin import ConfigMixin
+
+
+@dataclass
+class ProjectConfig(ConfigMixin):
+    """Complete project configuration.
+
+    Mirrors the structure of the project config.yaml (and metadata in pose config).
+    Field names match the old dictionary keys for round-trip compatibility.
```
C-Achard, line 26: [Note to self] Is the aliasing part of this PR or defined elsewhere ? Likely we could have some config abstraction that implements basic (abstract) helpers for migration, aliasing, etc to centralize as much as possible. Not sure if this would go in the mixin, likely best to take advantage of the mixin but have those implemented elsewhere

deeplabcut/core/config/config_mixin.py
```diff
@@ -0,0 +1,100 @@
...
+
+    @classmethod
+    def from_yaml(cls, yaml_path: str | Path) -> Self:
+        return cls.from_dict(read_config_as_dict(yaml_path))
+
+    def to_yaml(self, yaml_path: str | Path, overwrite: bool = True) -> None:
+        data = CommentedMap(self.to_dict())
+        for f in fields(self):
+            if (comment := f.metadata.get("comment")):
+                data.yaml_set_comment_before_after_key(f.name, before=comment)
```
C-Achard, line 86: This is very nice ! Definitely helpful for users imo

C-Achard (approved):

deeplabcut/core/config/project_config.py
deruyter92: Good point, still working on it in a follow-up PR. 

- Note to us: in case we decide to move to fully typed configurations instead of DictConfigs, it is easy to use the pydantic aliasing system (which just provides a reference to the new field). 

```python
@dataclass
class ExampleConfig():
    learning_rate: float = Field(alias="lr")
```    
    
- In OmegaConf this can also be achieved by adding an 'interpolation' from the old field to the new field. By registering a new resolver, we can even automatically provide warnings. 

```python
def alias_resolver(cfg, old_key, new_key):
    warnings.warn(
        f"Config field '{old_key}' is deprecated, use '{new_key}' instead",
        DeprecationWarning,
        stacklevel=3,
    )
    return OmegaConf.select(cfg, new_key)

OmegaConf.register_new_resolver("alias", alias_resolver)
```


I just want to avoid things getting cluncky with the interfacing between both Pydantic and OmegaConf aliases, so will think about it a bit more. Let's discuss this.

deruyter92: I have added it on the todo-list

Issue #3193, opened 2026-01-30
[WIP] Migration to structured configurations


deruyter92:
## Migrating to typed and validated configurations
This Issue will act as a placeholder to summarize the work in progress for migrating from plain dictionaries to typed and validated configuration classes. Originally this idea was formulated in #3172, but to facilitate easier review and a smooth migration the progress will be spilt into some intermediate steps with separate PRs. The old configuration system will stay in place during the intermediate steps and a versioning system will be put in place to facilitate long-term backwards compatibility across config versions.

#### Procedure
The migration will consist of small PRs subsequent changes, which can be reviewed separately, before merging into a [feature branch](https://github.com/DeepLabCut/DeepLabCut/tree/feat/structured_configs).  

### Work in progress / roadmap
(ticking after review and merging into the [feature branch](https://github.com/DeepLabCut/DeepLabCut/tree/feat/structured_configs))
- [x] Centralize project configuration logic 
- [x] Improve testing to capture existing behavior 
- [x] New config_mixin for easy moving between types (yaml file <> dict <> dataclass <> DictConfig)
- [x] Add typed configs for project config, pytorch config, etc. with pydantic validation 
- [x] Replace dictionary configs with identical DictConfig version (smooth transition)  
- [x] Add versioning system for migration between old and new configs
- [x] Improve configurations -> reduce duplicated fields, correct casing
- [x] Address in-place configuration edits throughout the pipeline
- [x] Add aliasing system for accessing new fields using old fieldnames (e.g. corer2move2 -> corner2move2)
- [x] Replace loaders in core config (e.g. deprectate `read_config`  in favor of typed ProjectConfig)
- [x] Address None-type / missing config issues in downstream code
- [x] Add tracking system for storing / logging changes in configuration (clean/dirty status w.r.t. disk) 
- [x] Refactor patterns getting items from None-type subconfigurations (configurations that are allowed to be missing) 
- [ ] Add flags for affordance in downstream use cases
- [ ] Add typed 3D Project Configs (and tensorflow ?)
- [ ] storing parallel versions (i.e. copy instead of overwrite)
- [x] Fix circular imports in core/config
- [x] Verify test coverage and improve if necessary
- [x] Improve loaders in e.g. `deeplabcut/pose_estimation_pytorch/data/base.py` (mismatching init, mixed project/pose logic)
- [x] Remove intermediate OmegaConf DictConfig representation and migrate to fully typed
[...]

PR #3194, opened 2026-01-30
[dev] C3 - Replace configuration dictionaries with DictConfigs.

deruyter92:
This PR is part of the WIP for migrating from dictionary configs to typed & validated configurations, and follows PR #3191. (see issue https://github.com/DeepLabCut/DeepLabCut/issues/3193 for an overview).

**Summary:** 
Refactors the PyTorch pose pipeline to use typed config classes (PoseConfig, ProjectConfig, ModelConfig) and OmegaConf DictConfig instead of raw dicts.

**Main changes:**
- Isolated the logic for loading default model configuration for a given model
- `make_pose_config` now creates an OmegaConf DictConfig validated against PoseConfig dataclass
- internally, the PyTorch Loaders now use PoseConfig DictConfig configurations instead of dictionaries

### DIFF
--- a/deeplabcut/core/config/utils.py
+++ b/deeplabcut/core/config/utils.py
@@ -13,11 +13,16 @@ from __future__ import annotations
 
 import warnings
 from pathlib import Path
-from typing import Callable
+from typing import TYPE_CHECKING, Callable
+
+if TYPE_CHECKING:
+    from deeplabcut.core.config.project_config import ProjectConfig, ProjectConfig3D
 
 import yaml
 import ruamel.yaml.representer
 from ruamel.yaml import YAML
+from omegaconf import DictConfig
+from pydantic import ValidationError
 
 from deeplabcut.core.engine import Engine
 
@@ -97,126 +102,10 @@ def create_config_template(multianimal: bool = False) -> tuple:
     Returns:
         (cfg_file, ruamelFile) for further editing and dumping.
     """
-    if multianimal:
-        yaml_str = """\
-# Project definitions (do not edit)
-Task:
-scorer:
-date:
-multianimalproject:
-identity:
-\n
-# Project path (change when moving around)
-project_path:
-\n
-# Default DeepLabCut engine to use for shuffle creation (either pytorch or tensorflow)
-engine: pytorch
-\n
-# Annotation data set configuration (and individual video cropping parameters)
-video_sets:
-individuals:
-uniquebodyparts:
-multianimalbodyparts:
-bodyparts:
-\n
-# Fraction of video to start/stop when extracting frames for labeling/refinement
-start:
-stop:
-numframes2pick:
-\n
-# Plotting configuration
-skeleton:
-skeleton_color:
-pcutoff:
-dotsize:
-alphavalue:
-colormap:
-\n
-# Training,Evaluation and Analysis configuration
-TrainingFraction:
-iteration:
-default_net_type:
-default_augmenter:
-default_track_method:
-snapshotindex:
-detector_snapshotindex:
-batch_size:
-\n
-# Cropping Parameters (for analysis and outlier frame detection)
-cropping:
-#if cropping is true for analysis, then set the values here:
-x1:
-x2:
-y1:
-y2:
-\n
-# Refinement configuration (parameters from annotation dataset configuration also relevant in this stage)
-corner2move2:
-move2corner:
-\n
-# Conversion tables to fine-tune SuperAnimal weights
-SuperAnimalConversionTables:
-        """
-    else:
-        yaml_str = """\
-# Project definitions (do not edit)
-Task:
-scorer:
-date:
-multianimalproject:
-identity:
-\n
-# Project path (change when moving around)
-project_path:
-\n
-# Default DeepLabCut engine to use for shuffle creation (either pytorch or tensorflow)
-engine: pytorch
-\n
-# Annotation data set configuration (and individual video cropping parameters)
-video_sets:
-bodyparts:
-\n
-# Fraction of video to start/stop when extracting frames for labeling/refinement
-start:
-stop:
-numframes2pick:
-\n
-# Plotting configuration
-skeleton:
-skeleton_color:
-pcutoff:
-dotsize:
-alphavalue:
-colormap:
-\n
-# Training,Evaluation and Analysis configuration
-TrainingFraction:
-iteration:
-default_net_type:
-default_augmenter:
-snapshotindex:
-detector_snapshotindex:
-batch_size:
-detector_batch_size:
-\n
-# Cropping Parameters (for analysis and outlier frame detection)
-cropping:
-#if cropping is true for analysis, then set the values here:
-x1:
-x2:
-y1:
-y2:
-\n
-# Refinement configuration (parameters from annotation dataset configuration also relevant in this stage)
-corner2move2:
-move2corner:
-\n
-# Conversion tables to fine-tune SuperAnimal weights
-SuperAnimalConversionTables:
-        """
-
+    warnings.warn("This function is deprecated. Use ProjectConfig instead.")
+    from deeplabcut.core.config.project_config import ProjectConfig
     ruamelFile = YAML()
-    cfg_file = ruamelFile.load(yaml_str)
+    cfg_file = ProjectConfig(multianimalproject=multianimal).to_dict()
     return cfg_file, ruamelFile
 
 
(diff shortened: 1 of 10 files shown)


tests/core/config/test_core_config.py
```diff
@@ -168,19 +166,19 @@ def test_create_config_template_3d_has_expected_keys():
...
     config_path = tmp_path / "config.yaml"
     config_path.write_text("project_path: /other/path\n")
     cfg = read_config(config_path)
-    assert cfg["engine"] == "tensorflow"
+    assert cfg["engine"] == "pytorch"
     assert cfg["project_path"] == str(tmp_path)
     # File should have been updated
     cfg_again = read_config_as_dict(config_path)
-    assert cfg_again["engine"] == "tensorflow"
+    assert cfg_again["engine"] == "pytorch"
```
C-Achard, line 223: Could we check the tf path as well ?

deeplabcut/pose_estimation_pytorch/data/dlcloader.py
```diff
@@ -164,7 +167,9 @@ def get_dataset_parameters(self) -> PoseDatasetParameters:
         Returns:
             An instance of the PoseDatasetParameters with the parameters set.
         """
-        crop_cfg = self.model_cfg["data"]["train"].get("top_down_crop", {})
+        crop_cfg = OmegaConf.select(
+            self.model_cfg, "data.train.top_down_crop", {}
+        )
```
C-Achard, line 174: I see what you meant, `select` is indeed nice

deeplabcut/core/config/utils.py
```diff
@@ -97,126 +102,10 @@ def create_config_template(multianimal: bool = False) -> tuple:
...
-\n
-# Refinement configuration (parameters from annotation dataset configuration also relevant in this stage)
-corner2move2:
-move2corner:
-\n
-# Conversion tables to fine-tune SuperAnimal weights
-SuperAnimalConversionTables:
-        """
-
+    warnings.warn("This function is deprecated. Use ProjectConfig instead.")
```
C-Achard:
```suggestion
    warnings.warn("This function is deprecated. Use deeplabcut.core.config.ProjectConfig instead.")
```

deeplabcut/pose_estimation_pytorch/config/make_pose_config.py
```diff
@@ -177,20 +168,83 @@ def make_pytorch_pose_config(
...
+
+    # Add inference config and weight init configs
+    pose_config.inference = InferenceConfig().to_dictconfig()
+    if weight_init is not None:
+        weight_init = WeightInitialization.from_any(weight_init).to_dictconfig()
+        pose_config.train_settings.weight_init = weight_init
+
+    # Update default values for the specific model architecture and project config.
+    # TODO JR 2026-01-28: using legacy v0 config (dict) for the defaults,
+    # we should move to typed model defaults and use omegaconf merge to update
```
C-Achard, line 236: Another reason to use omegaconf, so again we need to review whether we truly need it

deeplabcut/pose_estimation_pytorch/data/base.py
```diff
@@ -54,12 +57,54 @@ def __init__(
...
+                    raise ValueError(
+                        "`model_config_path` and `model_config` arguments cannot be provided together! "
+                        "Please provide only `model_config`."
+                    )
+                if Path(model_config_path).is_dir():
+                    model_config_path = Path(model_config_path) / "pytorch_config.yaml"
+                model_config = model_config_path
+            elif model_config is None:
+                raise ValueError(
+                    "`model_config` (or legacy argument `model_config_path`) must be provided."
```
C-Achard: Here we just need to decide whether we document/encourage the use of legacy args. I'd be in favor of not even mentioning it. This also reduces maintenance overhead in terms of docstring edits imo
```suggestion
                    "`model_config` argument must be provided."
```

deeplabcut/pose_estimation_pytorch/config/make_pose_config.py
```diff
@@ -177,20 +168,83 @@ def make_pytorch_pose_config(
...
+    pose_config.net_type = NetType(net_type) if net_type is not None else None
+
+    # Add inference config and weight init configs
+    pose_config.inference = InferenceConfig().to_dictconfig()
+    if weight_init is not None:
+        weight_init = WeightInitialization.from_any(weight_init).to_dictconfig()
+        pose_config.train_settings.weight_init = weight_init
+
+    # Update default values for the specific model architecture and project config.
+    # TODO JR 2026-01-28: using legacy v0 config (dict) for the defaults,
```
C-Achard:
```suggestion
    # TODO @deruyter92 2026-01-28: using legacy v0 config (dict) for the defaults,
```

deeplabcut/pose_estimation_pytorch/data/cocoloader.py
```diff
@@ -46,12 +49,23 @@ class COCOLoader(Loader):
     def __init__(
         self,
         project_root: str | Path,
-        model_config_path: str | Path,
+        model_config: PoseConfig | DictConfig | Path | str | None = None,
         train_json_filename: str = "train.json",
         test_json_filename: str = "test.json",
+        model_config_path: str | Path | None = None, # <--- deprecated
```
C-Achard: Since None is a valid input (based on type annot of this and `model_config`), we may want to distinguish None as an actual passed value versus the arg being empty.

To do so we could implement something like:
```
_DEPRECATED = object()
```
and then check as you already do in super() :
```
    if model_config_path is not _DEPRECATED:
        warnings.warn(
            "`model_config_path` is deprecated; use `model_config` instead.",
            DeprecationWarning,
            stacklevel=2,
        )
```

So then this would be: 
```suggestion
        model_config_path: str | Path | None = _DEPRECATED_, # <--- deprecated
```
which flags it nicely and gives us more handles to detect _DEPRECATED args.

Decorators for deprecated args are another option but too heavy for our use case I think, lmk.

tests/core/config/test_core_config.py
C-Achard: Or not relevant here ?

tests/core/config/test_core_config.py
```diff
@@ -10,6 +10,7 @@
 #
```
C-Achard, line 10: Do you feel confident this is enough in terms of coverage/branching cov ? Maybe check again just in case, perhaps in a C5 PR we could expand coverage. Would be nice to know how well the configs code is covered, and having that in a separate PR downstream of everything else could help locate blind spots. (But this is a lot already)

C-Achard (changes requested):
Overall very good, but I would change/check a few things in this one.

Also to have this written down somewhere, I'd strongly favor checking whether we truly need OmegaConf and its features; I feel we do, even if we do not take full advantage of them yet, but keeping dependencies lean is also a very important consideration imo.

You probably have a better outlook than I do,  but another path I see is just making pydantic dataclasses have dict-like access and use that in downstream code.

deeplabcut/pose_estimation_pytorch/data/cocoloader.py
deruyter92: good suggestion!

tests/core/config/test_core_config.py
deruyter92: good one. Added it to `test_read_config_preserves_existing_engine_and_project_path`. (the default Engine is pytorch, so missing should always resolve to Pytorch, but existing can be both.)

deruyter92:
@C-Achard, some relevant additional changes: 


**Pathlib and Enum serialization (see https://github.com/DeepLabCut/DeepLabCut/pull/3194/commits/538ca3a00ac0c1fe977bebec0d5498050c9d428c and https://github.com/DeepLabCut/DeepLabCut/pull/3194/commits/347ff18f8cb3b8ecdf3525193c378e0f3394a510):**
For the `project_path` and `pose_config_path` field, the default type is now changed to Path. For Enums and Paths, the values are stored as plain str when written to yaml file and converted back.  

**Empty yaml field handling (see https://github.com/DeepLabCut/DeepLabCut/pull/3194/commits/b890d6c908f79daacc948c9f66ceded81129329a):**
When yaml fields are empty, the field is ignored when reading. Without this, the code would break in several places, but we might want to consider stricter validation in the future, where empty fields are interpreted as None and a ValidationError is raised when None-type values are not allowed. 



e.g. `ProjectConfig.from_yaml('myconfig.yaml')` with the following yaml
```yaml
# myconfig.yaml
Task: openfield
scorer: Pranav
multianimalproject:
identity:
```
results in
```
ProjectConfig(Task='openfield', scorer='Pranav', multianimalproject=False, identity=None, ...)
```

PR #3197, opened 2026-02-03
[dev] C4 - add config migration system

deruyter92:
This PR is part of the WIP for migrating from dictionary configs to typed & validated configurations (see https://github.com/DeepLabCut/DeepLabCut/issues/3193 for an overview).

Introduces a config versioning and migration system for configurations. This facilitates easier changing of configuration schemas while keeping backward compatibility and reducing risks of missing-field-errors. 

**Changes**
- core/config/versioning: a registry for keeping track of migrations and adding changes over subsequent configuration versions. No changes are implemented yet, only placeholders, preparing for migration from v0 to v1 
- core/config/migration_mixin: the `MigrationMixin` class can be used as mixin for dataclasses to facilitate easy migration. It enables automatic loading and validation of earlier configurations into the current versions format. 
- tests/core/config/test_versioning.py for versioning and MigrationMixin.

**Usage**
After applying changes to the configuration dataclass (e.g. renaming of fields, or changing value types), the `CURRENT_CONFIG_VERSION` should be incremented accordingly (e.g. from 0 to 1). The corresponding transformations from the old field to the new field can be added to a migration function. e.g. `migrate_v0_to_v1`. 

For instance when deciding that value 'MULTI!' for field `bodyparts` should be converted to an empty list, one can add the following transformation:

```python
@register_migration(0, 1)
def migrate_v0_to_v1(config: dict) -> dict:
    """Migrate from unversioned/legacy config (v0) to v1.
    """
    normalized = config.copy()
    if normalized.get('bodyparts')  == 'MULTI!':
         normalized['bodyparts'] = []
    return normalized
```


deeplabcut/core/config/migration_mixin.py
```diff
@@ -0,0 +1,36 @@
...
+        # Convert to dictionary if ArgsKwargs is passed
+        if isinstance(data, ArgsKwargs):
+            names = [f.name for f in fields(cls)]
+            data = dict(
+                zip(names, data.args or []),
+                **data.kwargs or {},
+            )
+        if isinstance(data, dict):
+            data = migrate_config(data, target_version=CURRENT_CONFIG_VERSION)
+        return handler(data)
```
C-Achard: Do you prefer having this here or as part of versioning.py ?

deeplabcut/core/config/versioning.py
```diff
@@ -0,0 +1,151 @@
...
+
+@register_migration(1, 0)
+def migrate_v1_to_v0(config: dict) -> dict:
+    """Migrate from v1 to v0 (legacy format).
+    """
+    normalized = config.copy()
+
+    # TODO @deruyter92 2026-01-30: Migration logic goes here.
+
+    return normalized
```
C-Achard: This is really cool :D

deeplabcut/core/config/versioning.py
```diff
@@ -0,0 +1,151 @@
...
+        raise ValueError(
+            f"Missing migration from version {current_version} to {target_version}. "
+            f"Available migrations: {list(_MIGRATIONS.keys())}"
+        )
+    
+    migration_func = _MIGRATIONS[migration_key]
+    migrated = migration_func(migrated)
+    
+    return migrated
+
```
C-Achard: I think it'd be worth it to record which migration are applied and have the possibility for much more verbose logging, this cold be useful both for debug and as user info that their configs are being modified.

C-Achard: Also we need to decide if we overwrite config files or create a new copy, and then check folders for the most up-to-date config (or is that too complicated ? We may want to keep it simple)

deeplabcut/core/config/versioning.py
```diff
@@ -0,0 +1,151 @@
...
+    if migration_key not in _MIGRATIONS:
+        raise ValueError(
+            f"Missing migration from version {current_version} to {target_version}. "
+            f"Available migrations: {list(_MIGRATIONS.keys())}"
+        )
+    
+    migration_func = _MIGRATIONS[migration_key]
+    migrated = migration_func(migrated)
+    
+    return migrated
```
C-Achard, line 220: See comments on test file for potentially more robust handling

tests/core/config/test_versioning.py
```diff
@@ -0,0 +1,215 @@
...
+# DeepLabCut Toolbox (deeplabcut.org)
+# © A. & M.W. Mathis Labs
+# https://github.com/DeepLabCut/DeepLabCut
+#
+# Please see AUTHORS for contributors.
+# https://github.com/DeepLabCut/DeepLabCut/blob/main/AUTHORS
+#
+# Licensed under GNU Lesser General Public License v3.0
+#
+"""Tests for the configuration versioning and migration system."""
```
C-Achard, line 11: We may want to increase coverage of :

- Migration with missing fields
- Migrating e.g. 0 to 3 directly is rejected properly (not sequential)
- Downgrades
- Multiple sequence of migrations

I think this will be increasingly important, especially as we add validation logic. If something goes wrong in the middle of the migration we need clear debug ouput and some form of graceful handling to avoid leaving a broken state I think

C-Achard (commented):
Love this ! The tricky part will be in the actual conversion but otherwise this is nice, just consider adding better error handling/edge case testing/debug outputs/graceful error handling where relevant

deeplabcut/core/config/migration_mixin.py
deruyter92: Good one. Moved to versioning, thanks.

tests/core/config/test_versioning.py
deruyter92: Thanks, good suggestions. I have now improved the test coverage

deeplabcut/core/config/versioning.py
deruyter92: First point is addressed. Second point I added to the todo's

PR #3212, opened 2026-02-18
[dev] C3-II Additional refactoring of configurations 

deruyter92:
This PR is part of the WIP for migrating from dictionary configs to typed & validated configurations, and follows PR https://github.com/DeepLabCut/DeepLabCut/pull/3191. (see issue https://github.com/DeepLabCut/DeepLabCut/issues/3193 for an overview). It is a follow-up of #3194, with additional refactors that are required to make the change to DictConfig working. 

**Summary**
1. **Improve YAML I/O** :  A pair of central factory functions `get_yaml_loader()` and `get_yaml_dumper()` were added to deeplabcut/core/config/utils.py, instead of having downstream functions creating YAML handlers themselves. The dumper registers custom representers for Path, Enum, DictConfig, and ListConfig so these types are transparently serialized as plain YAML scalars/mappings. All call-sites across the codebase (metadata.py, export.py, modelzoo/config.py, test_trainset_metadata.py, etc.) now use these centralized factories instead of constructing YAML() / YAML(typ="safe", pure=True) ad hoc.

2. **Updates to the core configuration types**
   - PoseConfig and ProjectConfig now use ConfigDict(extra="forbid"), making them reject unknown fields - this was the intended behaviour in earlier PRs. 
   - The `make_pytorch_pose_config` flow is restructured (and corrected): 
Constructs a fully-typed PoseConfig object (with TrainSettingsConfig, WeightInitialization, etc.) before merging architecture defaults. Similarly, make_pytorch_test_config now constructs and validates a TestConfig instance. The Loader.update_model_cfg method now re-validates the config through PoseConfig after applying dot-path updates. **This will be again further simplified at the end of the PR series (e.g. after moving to fully-typed configs**
   - A new TestConfig dataclass was added to formalize the previously untyped test/inference config dict.
   - Many (sometimes ambiguous) fields that were used in the old configuration system are added when missing:  
      - NetType enum was extended with CSPNeXt variants (these were missing).
      - TrainSettingsConfig.weight_init now allows None (coding for default weight initialization, as before).
      - RunnerConfig gained a device field for backward compatibility with test scripts.
      - ProjectConfig gained legacy/alias fields (resnet, croppedtraining, with_identity, unique_bodyparts) 
      - **a from_yaml override is added** to ProjecConfig that adjusts project_path when the YAML file lives in a different directory. This matches prior behaviour. 
      -   GenSamplingConfig removed its bbox_margin field; Nearly the entire codebase uses the data.bbox_margin field in pose_config directly (for the only exception, PoseDatasetParameters it is now sourced from the data config as well).
 

3. **Improved handling/conversion of config Types**:  While in the central pipeline, we aim to move to typed configs, some places (e.g. before initial construction of the PoseConfig, or when serializing state dicts) may benefit from using plain dictionaries with native python types. 
    - This is made more explicit by using a `@ensure_plain_config` decorator that converts the config back to a plain dictionary. It is applied to key functions that must operate on plain dicts: `update_config`, `update_config_by_dotpath`, `replace_default_values`, `_load_pose_config_defaults`, `_add_ctd_conditions`, `add_metadata`, and the central registry's `build_from_cfg`.
   - In some places a conversion from non-native to native python types is put in place (e.g. `paf_predictor.py` now uses int instead of numpy.int64, which was equivalent but non-serializable)
   - Throughout the codebase (~25 locations), isinstance checks were broadened from dict to (dict, DictConfig) and from list to (list, ListConfig) to handle cases where OmegaConf containers leak into code expecting plain Python types. Each of these is marked with a TODO to eventually standardize on one representation. 
   - Numerous TODO comments flagging areas where config updates happen after validation, unvalidated super-animal configs, and typed-vs-plain dict decisions that still need resolution. 

4. **None-safe .get() pattern fix**
A widespread pattern cfg.get("key", {}) was replaced with cfg.get("key") or {}. When a key exists but is explicitly set to None in the config, the old pattern returns None (ignoring the default), leading to AttributeError on the chained .get(). The new pattern handles both missing keys and None values. These are marked with TODOs for future refactoring.

5. **Fix invalid test configuration:** in one of the test scripts, detector conditions are added to all of the training, even when bottom-up or ctd models are used. When these models are missing other required detector specifications, the new PoseConfig system throws an (accurate) validation error. The script is adjusted such that no partially detector-configs are added.

### DIFF
--- a/deeplabcut/core/config/project_config.py
+++ b/deeplabcut/core/config/project_config.py
@@ -168,6 +168,9 @@ class ProjectConfig(ConfigMixin):
         metadata={"comment": "\nConversion tables to fine-tune SuperAnimal weights"},
     )
 
+    # @TODO @deruyter92 2026-02-13: These aliases are used in parallel. One of them should be removed.
+    with_identity: bool | None = field(default=False, metadata={"comment": "Alias for 'identity'. Kept for backwards compatibility."})
+
     # TODO @deruyter92 2026-02-06: These parameters are no longer used in the new pipeline.
     resnet: int | None = field(
         default=None,
--- a/deeplabcut/core/config/project_config.py
+++ b/deeplabcut/core/config/project_config.py
@@ -170,6 +170,7 @@ class ProjectConfig(ConfigMixin):
 
     # @TODO @deruyter92 2026-02-13: These aliases are used in parallel. One of them should be removed.
     with_identity: bool | None = field(default=False, metadata={"comment": "Alias for 'identity'. Kept for backwards compatibility."})
+    unique_bodyparts: list[str] = field(default_factory=list, metadata={"comment": "Alias for 'uniquebodyparts'. Kept for backwards compatibility."})
 
     # TODO @deruyter92 2026-02-06: These parameters are no longer used in the new pipeline.
     resnet: int | None = field(
--- a/deeplabcut/core/config/project_config.py
+++ b/deeplabcut/core/config/project_config.py
@@ -97,7 +97,10 @@ class ProjectConfig(ConfigMixin):
     # VV TODO @deruyter92 2026-01-30: following the old original config.yaml template for now. VV
     # VV We should change this to a list[str] in the future. VV
     bodyparts: list[str] | str = 'MULTI!' 
-    individuals: list[str] = field(default_factory=list)
+
+    # TODO @deruyter92 2026-02-06: The current pipeline requires at least one individual defined in the 
+    # default configuration. This will be removed in the future.
+    individuals: list[str] = field(default_factory=lambda: ['individual_1']) 
     uniquebodyparts: list[str] = field(default_factory=list)  # multi-animal project key
     multianimalbodyparts: list[str] = field(
         default_factory=list
@@ -164,6 +167,16 @@ class ProjectConfig(ConfigMixin):
         default=None,
         metadata={"comment": "\nConversion tables to fine-tune SuperAnimal weights"},
     )
+
+    # TODO @deruyter92 2026-02-06: These parameters are no longer used in the new pipeline.
+    resnet: int | None = field(
+        default=None,
+        metadata={
+        "comment": "\nThese are very old parameters that are no longer used "
+        "in most cases. They are kept for backwards compatibility."
+        },
+    )
+    croppedtraining: bool | None = None
 
     @classmethod
     def from_yaml(cls, yaml_path: str | Path, *args, **kwargs) -> Self:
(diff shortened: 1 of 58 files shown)


deeplabcut/core/config/project_config.py
```diff
@@ -161,3 +167,37 @@ class ProjectConfig(ConfigMixin):
...
+        metadata={
+        "comment": "\nThese are very old parameters that are no longer used "
+        "in most cases. They are kept for backwards compatibility."
+        },
+    )
+    croppedtraining: bool | None = None
+
+    @classmethod
+    def from_yaml(cls, yaml_path: str | Path, *args, **kwargs) -> Self:
+        """
```
C-Achard, line 186: To make the contract more explicit :
- Define to_disk or to_yaml write here
- Set a read  only flag
- Make a "autoupdateconfig path" method that  corrects the path is needed

This makes the  config file self-documenting and should minimize redefinitions/patching downstream. Helper glue can be added but this schema should still be the authority

deeplabcut/pose_estimation_pytorch/config/utils.py
```diff
@@ -151,6 +154,11 @@ def update_config(config: dict, updates: dict, copy_original: bool = True) -> di
     return config
 
 
+# TODO @deruyter92 2026-02-17: This function is currently still used to update 
+# the config in a late stage (after the config is initially created and validated).
+# We should move away from this strategy and update all override arguments during 
+# config creation.   
+@ensure_plain_config
```
C-Achard, line 161: Here I think there's not much choice  if we do not want to touch the underlying pipeline, but imo the primary concern there is not separating concerns :
- Config is  user-set, validated first, never touched again
- Runtime data is modified in place and read/written whenever needed, with whatever validation it needs.
But again his I think is best done in a follow up PR, in the  meantime fake it and validate where possible

deeplabcut/pose_estimation_pytorch/config/pose.py
```diff
@@ -135,3 +152,32 @@ class PoseConfig(ConfigMixin):
     runner: RunnerConfig | None = None
     train_settings: TrainSettingsConfig | None = None
     detector: DetectorConfig | None = None
+
+
+@dataclass(config=ConfigDict(extra="forbid"))
```
C-Achard, line 157: Make sure this includes proper self-set flags (no public setters) that clearly specify e.g. if it can be used for training, or if there is a missing  field despite the stated model type for strong, self-documenting validation

C-Achard (approved):
See future notes ! Great job

PR #3208, opened 2026-02-12
Add fmpose3d functionality

deruyter92:
## Add FMPose3D integration for monocular 3D pose estimation

Add functionality for using https://github.com/AdaptiveMotorControlLab/FMPose3D to lift 2D predictions to 3D keypoints. (see [paper](https://arxiv.org/abs/2602.05755))

### Summary

- Integrate [FMPose3D](https://github.com/AdaptiveMotorControlLab/FMPose3D) into the DeepLabCut model zoo, enabling monocular 3D pose estimation via flow matching for both **human** (17 H36M joints) and **animal** (26 Animal3D joints) pipelines
- Add a convenience wrapper `get_fmpose3d_inference_api()` that creates a ready-to-use inference object with automatic weight downloading from HuggingFace Hub
- Add integration tests covering API initialization (lightweight) and full end-to-end 2D-to-3D inference (network-dependent)
- Add a Jupyter notebook recipe to the docs cookbook showing end-to-end usage, two-stage workflows, and how to lift existing DLC 2D predictions to 3D

### Details

**Core integration** (`fmpose3d.py`): A thin wrapper that constructs an `FMPose3DConfig` and returns an `FMPose3DInference` object. Supports both `fmpose3d_humans` and `fmpose3d_animals` model types, optional custom snapshot paths, device selection, and extra config kwargs.

**Tests** (`test_fmpose_integration.py`):
- `test_api_init` — parametrized over both model types; verifies the API object has the expected interface (`prepare_2d`, `pose_3d`, `predict`). Runs on CPU with no network required.
- `test_prepare_2d_and_pose_3d` / `test_predict_end_to_end` — full integration tests that download weights and run inference on the included example image. Skipped automatically when no network is available.

**Recipe notebook** (`fmpose3d.ipynb`): Covers three workflows:
1. End-to-end human pose estimation (single `predict` call)
2. End-to-end animal pose estimation
3. Two-stage pipeline (separate 2D detection + 3D lifting)
4. Lifting existing DLC 2D predictions (from `analyze_images` or `analyze_videos`) to 3D

PR #3221, opened 2026-02-25
Update FMPose3D modelzoo integration

deruyter92:
Note: this PR depends on the bleeding-edge code from the FMPose3D repository. After release, the latest FMPose3D dependency needs to be specified in DeepLabCut before merging. 

**Summary:**
This PR adds a dedicated FMPose3D inference path in the PyTorch modelzoo flow and wires early routing for fmpose3d_* models. It relocates FMPose-specific wrappers/inference helpers to pose_estimation_pytorch/modelzoo/fmpose_3d, standardizes outputs to include explicit 3D artifacts (*_3d.h5 / *_3d.json).

**Motivation:**
Where the first implementation of FMPose3D in #3208 was rather separated from the rest of the modelzoo pipeline, this PR aims to integrate it better in the existing modelzoo video inference pipeline.

**Changes:**
1. Revert preliminary implementation in #3208: 
     - move api wrapper to pose_estimation_pytorch
     - delete recipe notebook on how to use the API

2. Route video inference through video_inference_superanimal:
     - early branching to own video inference pipeline
     - just a thin wrap around FMPose3D, using the superanimal inference format
     
3. Add brief example usage in the existing superanimal docs 

**Example usage:**
```python
import deeplabcut

video_path = "demo-video.mp4"
result = deeplabcut.video_inference_superanimal(
    videos=[video_path],
    superanimal_name="",  # ignored for fmpose3d models
    model_name="fmpose3d_animals",
    batch_size=8,
    fmpose_return_3d=True,  # (optional: include 3D dataframe in returned payload)
)
```

**Additional considerations / limitations:**
- Single-individual 3D limitation: upstream FMPose3D lifting still effectively uses one individual, so multi-animal settings are clamped/warned rather than fully supported.
- Mixed return contract: enabling fmpose_return_3d changes payload structure, which adds API complexity for callers.
- Inconsistent parameter intuition: the entire 2D modelzoo convention currently follows the intuition that `superanimal_name` -> means 'dataset family' and `model_name` -> means 'model architecture'. For FMPose3D the 'dataset family' is different and therefore ignored. 
- Inseparability: the core principle of FMPose3D is lifting 2D keypoints to 3D, however to simplify the integration, FMPose3D is treated as a single model that directly computes video -> 2D -> 3D and the 2D step is now made inseparable. Therefore, users that want to use the 3D lifting on their own 2D keypoints should use the FMPose3D api, rather than this modelzoo inference function.
- No Video adaptation / Transfer learning: while the other 2D modelzoo models can be finetuned using video adaptation or further trained using pretrained weight initialization this is not the case for the FMPose3D model. If people want to finetune the backbone 2D estimator, they should work with a standard 2D superanimal model and use the FMPose3D api for lifting their keypoints, which is fairly straightforward. 
- Redundant outputs: consistent with the 2D superanimal pattern, both JSON and HDF 3D artifacts are written, which increases storage and maintenance overhead. This might need to be reconsidered in the future.
- Format bridge complexity: FMPose outputs are converted into DLC-style 3D DataFrames, consistent with the 3D pipelines in DeepLabCut. While this works fine if formatting conventions are kept the same, a more robust (validated) data format would be useful.


docs/ModelZoo.md
```diff
@@ -92,6 +92,25 @@ deeplabcut.video_inference_superanimal([video_path],
...
+import deeplabcut
+
+video_path = "demo-video.mp4"
+deeplabcut.video_inference_superanimal(
+    videos=[video_path],
+    superanimal_name="superanimal_quadruped",  # ignored for fmpose3d models
+    model_name="fmpose3d_animals",
+    batch_size=8,
+    fmpose_return_3d=True,  # include 3D dataframe in returned payload
+)
```
C-Achard, line 119: I think being able to call it like this is really nice in the end, great job !

tests/pose_estimation_pytorch/modelzoo/test_fmpose_integration.py
```diff
@@ -10,13 +10,18 @@
 #
 import pathlib
 import socket
+from types import SimpleNamespace
 
 import numpy as np
+import pandas as pd
 import pytest
```
C-Achard, line 17: Maybe add pytest marks for the future throughout the file !

deeplabcut/pose_estimation_pytorch/modelzoo/fmpose_3d/inference.py
```diff
@@ -0,0 +1,261 @@
...
+    for frame_idx in range(num_frames):
+        n_det = min(num_persons, max_individuals)
+        bodyparts_array = np.zeros((max_individuals, num_bodyparts, 3))
+        bodyparts_array[:n_det, :, :2] = all_kpts[:n_det, frame_idx, :num_bodyparts, :2]
+        bodyparts_array[:n_det, :, 2] = all_scores[:n_det, frame_idx, :num_bodyparts]
+        per_frame.append({"bodyparts": bodyparts_array})
+    return per_frame
+
+
+def _poses3d_to_dataframe(poses_3d: list[np.ndarray], df_2d, scorer_3d: str):
```
C-Achard, line 68: Side note, perhaps flag these for when we move away from pandas. Can be a simple comment for now, but we can try to be consistent within folders and define each flag in the __init.py__ of the folder

```suggestion
# NOTE Pandas I/O - currently no intermediate validation or flexibility re: data format
def _poses3d_to_dataframe(poses_3d: list[np.ndarray], df_2d, scorer_3d: str):
```

deeplabcut/pose_estimation_pytorch/modelzoo/fmpose_3d/inference.py
```diff
@@ -0,0 +1,261 @@
...
+    """Create and fill a 3D dataframe using the shared auxiliary helper."""
+    df_3d, scorer_3d, bodyparts = auxiliaryfunctions_3d.create_empty_df(
+        df_2d, scorer_3d, "3d"
+    )
```
deruyter92, line 72: Add a note that this may need attention when refactoring the keypoints.


```suggestion

# NOTE:  i_o; pandas; kpt_refactor;  this function may need to change in the future, to improve dataframe
#  i/o migration to validated keypoint schemas (parquet)
def _poses3d_to_dataframe(poses_3d: list[np.ndarray], df_2d, scorer_3d: str):
    """Create and fill a 3D dataframe using the shared auxiliary helper."""
    df_3d, scorer_3d, bodyparts = auxiliaryfunctions_3d.create_empty_df(
        df_2d, scorer_3d, "3d"
    )
```

C-Achard (commented):
Thanks, looks good overall. I generally agree with your point on separation of concerns, but probably the main concern for now is that it conforms to the modelzoo API

tests/pose_estimation_pytorch/modelzoo/test_fmpose_integration.py
deruyter92: Good point, will do that

MMathisLab:
< Note: this PR depends on the bleeding-edge code from the FMPose3D repository >

Shall I make a new release of FMPose3D on pypi?

MMathisLab:
Here I disagree:

> Inconsistent parameter intuition: the entire 2D modelzoo convention currently follows the intuition that superanimal_name -> means 'dataset family' and model_name -> means 'model architecture'. For FMPose3D the 'dataset family' is different and therefore ignored. 

Because it uses the 2D SA_quadruped data, or human data, for example, then the architecture is the 3D lifting pipeline... 

```
    superanimal_name="",  # ignored for fmpose3d models
    model_name="fmpose3d_animals",
```

why not use this, which better adheres to the SA style? 
```
    superanimal_name="quadruped",  # quadruped, human
    model_name="fmpose3d", 
```
and keep naming consistent: 

`fmpose_return_3d=True` --> `fmpose3d_return_3d=True`

deeplabcut/modelzoo/video_inference.py
```diff
@@ -352,6 +364,42 @@ def video_inference_superanimal(
...
+            provided_superanimal_name = superanimal_name or "<not provided>"
+            if superanimal_name != recommended_superanimal_name:
+                warnings.warn(
+                    "For FMPose3D models, model selection is driven by 'model_name'. But for API "
+                    "consistency, it is recommended to set 'superanimal_name' to the corresponding value."
+                    f"Provided superanimal_name={provided_superanimal_name!r} differs from the "
+                    f"recommended value for {model_name!r}: "
+                    f"{recommended_superanimal_name!r}.",
+                    stacklevel=2,
+                )
```
deruyter92, line 386: It will just give a warning now if a non-recommended superanimal name was provided. 

The recommended names are now (in line with the existing names):
```python
recommended_superanimal_name = {
   "fmpose3d_animals": "superanimal_quadruped",
   "fmpose3d_humans": "superanimal_humanbody",
}
```

@MMathisLab do you agree? Or should they be as you literally suggested: 
```python
recommended_superanimal_name = {
   "fmpose3d_animals": "quadruped",
   "fmpose3d_humans": "human",
}
```

deruyter92:
> < Note: this PR depends on the bleeding-edge code from the FMPose3D repository >
> 
> Shall I make a new release of FMPose3D on pypi?

yes that would be great!

deeplabcut/modelzoo/video_inference.py
MMathisLab: I thnk literally as I suggest, to conform to other models

deeplabcut/modelzoo/video_inference.py
```diff
@@ -352,6 +364,42 @@ def video_inference_superanimal(
...
         )
     elif framework == "pytorch":
+        if model_name.startswith("fmpose3d"):
+            logger.info("Running video inference on %s using %s", videos, model_name)
+
+            recommended_superanimal_name = {
+                "fmpose3d_animals": "superanimal_quadruped",
+                "fmpose3d_humans": "superanimal_humanbody",
+            }.get(model_name)
+
```
deruyter92, line 376:
```suggestion

            recommended_superanimal_name = {
                "fmpose3d_animals": "quadruped",
                "fmpose3d_humans": "human",
            }.get(model_name)

```

deeplabcut/modelzoo/video_inference.py
deruyter92: Ok as you wish. changed now:
 `"superanimal_quadruped"` -> `"quadruped"` 
 `"superanimal_humanbody"` -> `"human"`

deruyter92:
~Note to self, @C-Achard, and @MMathisLab ~
~Changed the status to draft to make sure that it is only merged after~
~1. Releasing the latest FMPose3D package~
~2. Updating the lower-bound required version of the fmpose3d extra in this PRs requirement.txt and setup.py.~

**Edit**: a new version of FMPose3d is now published. The requirements are updated! ✅

pyproject.toml
deruyter92:
```suggestion
fmpose3d = [ "fmpose3d>=0.0.8" ]
```

MMathisLab (approved):

Issue #3218, opened 2026-02-24
Task / Feature request: native in-memory batched inference API for Model Zoo


deruyter92:
Task / Feature request: native in-memory batched inference API for Model Zoo

**Summary**
Add a first-class inference pipeline that accepts in-memory image/frame arrays (np.ndarray / optionally torch.Tensor) and runs batched prediction without requiring files on disk.

**Motivation**
Current high-level Model Zoo flows are mainly file-path based (video/image paths, output files). This is awkward for services pipelines and notebooks where frames already exist in memory. 

**Problem today**
No clear public API for batched in-memory inputs. Existing convenience APIs emphasize disk I/O and artifact writing. For array-based predictions users end up stitching together lower-level runners or relying on the web-app inference class (which needs extra glue code, is not documented and can be error-prone).

**Requested API shape (example)**
```python
preds = deeplabcut.modelzoo.video_analysis(
   frames=frames, # list/iterable of HxWx3 arrays or [N,H,W,3]
   superanimal_name="superanimal_topviewmouse", 
   model_name="hrnet_w32", 
   detector_name="fasterrcnn_resnet50_fpn_v2", 
   batch_size=16,  
   detector_batch_size=16,  
   max_individuals=3,   
   device="cuda:0",   
)
``` 

**Acceptance criteria**
Supports top-down and bottom-up SuperAnimal models.
Fully in-memory path (no required writes to image/video/h5/json).
True batching for detector and pose stages.
Returns structured predictions per frame.
Clear validation/errors for shape/dtype/device mismatches.
Documented example usage.

**Impact**
This would make DLC Model Zoo much easier to use in production / integrated projects, reduce boilerplate, and improve reliability/performance for non-file-based workflows.

deruyter92:
solved in #3222

PR #3222, opened 2026-02-26
Add internal helper for batched modelzoo inference from in-memory arrays (inference runner)

deruyter92:
## Summary

- Adds a new in-memory inference helper:
  - `create_superanimal_inference_runners(...)`
  - located in `deeplabcut.pose_estimation_pytorch.modelzoo.inference`
- Exposes this helper from `deeplabcut.pose_estimation_pytorch.modelzoo.__init__` for convenient imports.

## Motivation

Currently it is fairly easy to run batched inference on our PyTorch models, but the modelzoo api is an exception that requires us to save images to disk first. This PR implements a modelzoo inference runner factory to easily run batched inference from in-memory image arrays. This can be used internally for faster inference. (Inspired by `get_inference_runners`). 

This solves #3218 

## Example usage

```python
from pathlib import Path
import numpy as np
from PIL import Image
from deeplabcut.pose_estimation_pytorch.modelzoo.inference import (
    create_superanimal_inference_runners,
)

img_paths = [
    "/path/to/images/frame_0000.png",
    "/path/to/images/frame_0001.png",
    "/path/to/images/frame_0002.png",
]
images = [np.asarray(Image.open(Path(p)).convert("RGB")) for p in img_paths]

pose_runner, det_runner, model_cfg = create_superanimal_inference_runners(
    superanimal_name="superanimal_quadruped",
    model_name="hrnet_w32",
    detector_name="fasterrcnn_resnet50_fpn_v2",
    max_individuals=10,
    batch_size=1,
    detector_batch_size=1,
)

det_preds = det_runner.inference(images)
pose_inputs = list(zip(images, det_preds))
pose_preds = pose_runner.inference(pose_inputs)
```


deeplabcut/pose_estimation_pytorch/modelzoo/inference_helpers.py
```diff
@@ -0,0 +1,228 @@
...
+
+    pose_snapshot_path = customized_pose_checkpoint
+    if pose_snapshot_path is None:
+        pose_snapshot_path = get_super_animal_snapshot_path(
+            dataset="superanimal_humanbody",
+            model_name=model_name,
+            download=True,
+        )
+
+    COCO_PERSON = 1
```
C-Achard: Just a small curiosity, why not directly pass this as a default arg or keep it at the top ?
Perhaps this could be moved to a more centralized config/constants file at a later point too

C-Achard (commented):
Looks good so far, thanks !

deeplabcut/pose_estimation_pytorch/modelzoo/inference_helpers.py
deruyter92: Good comment. The answer is: I copied the pattern from elsewhere, which is actually not a good idea. 
I have now centralized the constant in modelzoo.utils instead, to reduce code duplication and (which should also make it clear that it is not intended to be a tunable parameter for each of the callers).

deeplabcut/pose_estimation_pytorch/modelzoo/inference_helpers.py
```diff
@@ -0,0 +1,228 @@
...
+        >>> pose_preds = pose_runner.inference(pose_inputs)
+        >>> print(len(pose_preds))
+    """
+    if model_name.lower().startswith("fmpose3d"):
+        raise NotImplementedError(
+            "FMPose3D is not supported in this helper. Use the FMPose3D inference API."
+        )
+
+    if device in (None, "auto"):
+        device = "cuda:0" if torch.cuda.is_available() else "cpu"
```
C-Achard: This could be left for users to decide ideally, choosing 0 by default can sometimes not be ideal behavior

C-Achard: I assume this had to be done due to downstream call sites not taking "device". If it's not too difficult, perhaps using this PR to add them could be worth it ? We can keep the current behavior if device is not set as a fallback without risk I believe

deeplabcut/pose_estimation_pytorch/modelzoo/inference_helpers.py
```diff
@@ -0,0 +1,228 @@
...
+        and detector_name is None
+        and customized_detector_checkpoint is None
+    ):
+        logging.warning(
+            "Top-down model configured without a detector. "
+            "Returning detector_runner=None; pass bboxes in pose input context."
+        )
+
+    model_cfg = update_config(model_cfg, max_individuals=max_individuals, device=device)
+    # Local import avoids import-time circular dependency with modelzoo package init.
```
C-Achard: Is this because of the __init__ export?

deeplabcut/pose_estimation_pytorch/modelzoo/inference_helpers.py
```diff
@@ -0,0 +1,228 @@
...
+        customized_pose_checkpoint=customized_pose_checkpoint,
+        customized_detector_checkpoint=customized_detector_checkpoint,
+    )
+
+    pose_runner, detector_runner = get_inference_runners(
+        model_config=model_cfg,
+        snapshot_path=weight_init.snapshot_path,
+        max_individuals=max_individuals,
+        num_bodyparts=len(model_cfg["metadata"]["bodyparts"]),
+        num_unique_bodyparts=len(model_cfg["metadata"]["unique_bodyparts"]),
```
C-Achard: Note for later: if we improve the config schema this should be a getter not an implicit len() check (e.g. `model_cfg.num_bodyparts()`)

C-Achard (commented):
Really nice, I like the minimally invasive design.
If we're alright with making small tweaks under the hood, see comments, otherwise all good.

(Note : did not have the opportunity to test it yet)

deeplabcut/pose_estimation_pytorch/modelzoo/inference_helpers.py
deruyter92: Good point. Actually, the num_bodyparts are automatically resolved when the arguments are omitted, so I will just omit them here.

deeplabcut/pose_estimation_pytorch/modelzoo/inference_helpers.py
deruyter92: yes it was. Instead of going for the fast solution, I now implemented a more proper lazy exporting in __init__.py, which also resolves the problem.

deeplabcut/pose_estimation_pytorch/modelzoo/inference_helpers.py
deruyter92: 'auto' is well-handled downstream, only None would cause issues. I now changed the lines so that None is mapped to 'auto'. 

Downstream the device is resolved in pose_estimation_pytorch.utils:

```python
def resolve_device(model_config: dict) -> str:
    """Determines which device should be used from the model config

    When the device is set to 'auto':
        If an Nvidia GPU is available, selects the device as cuda:0.
        Selects 'mps' if available (on macOS) and the net type is compatible.
        Otherwise, returns 'cpu'.
    Otherwise, simply returns the selected device

    Args:
        model_config: the configuration for the pose model

    Returns:
        the device on which training should be run
    """
    device = model_config["device"]
    supports_mps = "resnet" in model_config.get("net_type", "resnet")

    if device == "auto":
        if torch.cuda.is_available():
            return "cuda"
        elif supports_mps and torch.backends.mps.is_available():
            return "mps"
        return "cpu"
    return device
```

BTW for people that want another GPU to be detected automatically, it is easiest to configure this via env variable, e.g.:
```bash
export CUDA_VISIBLE_DEVICES=1
```

MMathisLab (approved):
I did not run the PR, so maybe speed test this, but otherwise Lgtm!

deruyter92:
This PR basically exposes the inference runner that is used in superanimal image folder analysis, so you don't have the overhead of reading and writing images. As the backend inference runner is exactly the same (and you have to load the images at some point in your pipeline) the comparison is a bit unfair. But assuming that you already loaded the images in memory elsewhere in your script (the usecase of this PR), the benefits are as follows: 

image

deruyter92:
Thanks for the review! I'll merge it once the tests are complete!

PR #3229, opened 2026-03-05
Optimize CI runner time use on multiple pushes to same PR

C-Achard:
Introduce a concurrency block to the build job in .github/workflows/python-package.yml to cancel outdated matrix runs for the same PR and matrix combo. The concurrency group combines workflow name, PR number (or run id for non-PR runs), OS and python-version; cancel-in-progress is enabled for pull_request events to reduce wasted CI runs when new commits are pushed.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -7,7 +7,7 @@ on:
     branches: [ main ]
 
 jobs:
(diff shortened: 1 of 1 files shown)


C-Achard:
Superseded by #3046

PR #3238, opened 2026-03-11
DLC-live-GUI  docs : Install from PyPI pre-release

C-Achard:
Replace source/editable installs with PyPI pre-release installs (use --pre) across uv/pip and pip instructions, and remove the repository clone/install-from-source workflow. 

- [x] Wait for DLC-live-GUI 2.0.0 rc release

Add an "important" note explaining the package is distributed as a pre-release, and introduce a "Both backends" tab-item with a caution about increased environment size and example install commands for installing both pytorch and tf extras. 
Also update corresponding conda/mamba sections to use the pre-release package.


MMathisLab (approved):

C-Achard:
Thanks @MMathisLab !

PR #3253, opened 2026-03-23
update conda yaml: add pyside6 as conda dependency

### LINKED ISSUE #3252: DeepLabCut YAML install works on macOS, but on Windows the GUI fails with Qt/PySide6 import errors
Hi, I’m trying to set up DeepLabCut on a Windows workstation for behavior analysis, and I ran into a GUI issue when installing from the YAML environment file.

On macOS, the YAML-based installation works fine.
On Windows, the environment is created successfully, but the GUI does not launch.

I created the environment from the provided YAML file and activated it, then ran: python -m deeplabcut

DeepLabCut starts loading, but then it falls back to light mode and GUI functionality is unavailable:
Loading DLC 3.0.0rc13...
DLC loaded in light mode; you cannot use any GUI (labeling, relabeling and standalone GUI)

Then I get a Qt-related error: qtpy.QtBindingsNotFoundError: No Qt bindings could be found

I checked the environment, and qtpy, pyside6, pyside6-addons, pyside6-essentials, and qt-main were all present.
However, the following imports fail on Windows:
python -c "from PySide6 import QtCore; print(QtCore.__file__); print(QtCore.qVersion())"
python -c "from PySide6 import QtWidgets; print(QtWidgets.QApplication)"

ImportError: DLL load failed while importing QtCore: The specified procedure could not be found.
ImportError: DLL load failed while importing QtWidgets: The specified procedure could not be found.

So it seems that:

the package is installed, but the actual Qt DLLs / bindings are not loading correctly on Windows, which causes qtpy to think that no usable Qt binding is available.

-Environment notes
Windows: GUI fails after YAML install
macOS: same YAML-based workflow works fine
GPU is AMD Radeon RX 550, so I am using CPU only
Miniconda is being used

python and conda paths look correct in the environment

**Has anyone seen this specific issue on Windows with YAML-based installation?**

I’m mainly trying to understand:

1. whether this is a known PySide6 / Qt DLL issue on Windows,

2. whether the YAML file pulls a problematic Qt/PySide combination on Windows,

3. and whether there is a recommended fix for Windows besides manually rebuilding the environment.

Any suggestions would be appreciated.

deruyter92:
solves #3252

**Summary**
The installation using our conda yaml can cause a dll shadowing problem on some machines if ffmpeg is installed via conda and pyside6 via pip (the conda-installed ffmpeg dll's are resolved first, but can have a mismatching version for pyside6). This issue of mixing conda and pip installations (for ffmpeg and pyside6) can be addressed by either installing both via pip or both via conda. This PR includes pyside6 as a conda dependency for users of the conda yaml file.

**Notes:**
1. We might want to stop supporting mixed conda + pip installations and delete this yaml file altogether, or at least clearly communicate that installation via uv or pip is preferred over mixed installations via conda + pip (this is not required anymore and still a big cause of installation problems). 
2. An alternative to the current solution in this PR is to remove ffmpeg as a conda requirement, which also solves the problem. It seemed safer for me to add them both as conda requirement for now as I don't know the original motivation for putting ffmpeg as conda requirement in the yaml. If this is not necessary, we might prefer adding both as pip rather than conda requirements.

### DIFF
--- a/conda-environments/DEEPLABCUT.yaml
+++ b/conda-environments/DEEPLABCUT.yaml
@@ -29,6 +29,7 @@ dependencies:
   - ipython
   - jupyter
   - ffmpeg
+  - pyside6
   - pip:
     - torch
     - torchvision


C-Achard:
It seems that the PyPi version of 'ffmpeg' is a thin wrapper for python that assumes the binaries are already installed, which is why conda is in use there.

I've seen a few packages use a different method, which is to pause when ffmpeg is not installed and ask whether to run a script to install the binaries; not sure if we should consider doing the same.

C-Achard:
Regarding the installation and the python wrapper : [see here](https://github.com/kkroening/ffmpeg-python?tab=readme-ov-file#installing-ffmpeg)

C-Achard (approved):
Good fix for now!

deruyter92:
Cleaned it up
 (thanks for adding your CI changes in this PR @C-Achard, but now they are merged in main so we can keep those separate).

C-Achard:
Indeed, sorry for the extra work @deruyter92

MMathisLab (approved):

PR #3140, opened 2025-11-07
fix ruamel.yaml dumping newline for long filenames

deruyter92:
**Motivation**
This PR fixes an issue with external package `ruamel.yaml`,  causing corrupted yaml files when creating a project with long video paths containing spaces. (Fixes https://github.com/DeepLabCut/DeepLabCut/issues/3119)

**Changes**
- The issue is patched by increasing the width parameter before dumping, preventing long sequences to be moved to a new line, as suggested by the author [here](https://stackoverflow.com/questions/31197268/pyyaml-yaml-dump-produces-complex-key-for-string-key-122-chars/31199123#31199123).
- Some unit tests are added for writing config files with various (unorthodox) video paths, to make sure that video configurations are loaded as intended.
- **An additional issue is fixed in** `deeplabcut.create_project.new`: The create_project function unintendedly treated empty video directories as a video file, for which a symbolic link is created. This behavior is now fixed

PR #3275, opened 2026-04-08
Fix failing local Windows tests due to ruamel parsing

C-Achard:
Update multianimal project fixture to produce stable YAML: import ruamel.yaml.scalarstring.SingleQuotedScalarString and use it for the video_sets key (posix path), set yaml width to avoid unwanted wrapping, and use project_dir.as_posix() for project_path. 

Also write config.yaml with explicit UTF-8 encoding and ensure crop is a string. 
These changes make YAML dumping deterministic and platform-independent for the test.


C-Achard:
(Similar to #3140, but in tests)

deruyter92 (approved):
Good fix!

PR #3276, opened 2026-04-08
Use pinned ffmpeg version in CI

C-Achard:
### Problem description

Due to the way the BtBn ffmpeg build CI is written, there is a ~2-3 hours gap in the day, starting around 1 PM CEST, where the latest release URL points to the previous build rather than the latest, a drift that causes the download to fail in our CI, as the files in the "latest" release do not have the correct name.

### Proposed solution

To alleviate this, the PR installs a pinned version of ffmpeg from BtBn which should be valid for roughly 2 years, limiting the impact for build downloads. (Includes a note to update in two years.)

### Alternatives

We could revert to Chocolatey or Winget, but with some retry logic to account for sporadic download failures.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -124,16 +124,19 @@ jobs:
           elif [ "$RUNNER_OS" == "macOS" ]; then
             brew install ffmpeg || true
           fi
-      - name: Install ffmpeg (Windows, BtbN latest)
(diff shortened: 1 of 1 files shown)


deruyter92 (approved):

deruyter92:
@C-Achard agreed as discussed, thanks for implementing. 

We should merge this one as soon as possible, as it obstructs running CI tests a few hours per day.

AlexEMG (approved):

PR #3278, opened 2026-04-09
Docs versioning: Add glob support, better validation and reporting

C-Achard:
### Purpose

The docs & notebook versioning tool currently does not support small QoL features such as glob matching, early exit when no files were matched in --targets, and listing which files it ran on.
This PR aims to extend the file to address these points and make it more comfortable to use.

###  Changes

- Normalize CLI target specs (handle Windows/backslashes, ./, absolute paths) and classify them as file/dir/glob. 
- Implement matching logic and validation (report matched files and unmatched selectors, return rc=2 on unmatched), and apply target specs in scan/update. 
- Add helpers (normalize_target_spec, compile_target_specs, validate_requested_targets, print_target_match_summary, iter_scan_candidate_paths) to perform the above and adjust main argument help. 
- Update tests to cover directory, glob, Windows-style paths, and CLI reporting.

### DIFF
--- a/tests/tools/docs_and_notebooks_checks/test_check_contracts.py
+++ b/tests/tools/docs_and_notebooks_checks/test_check_contracts.py
@@ -4,6 +4,7 @@ import importlib.util
 import json
 import os
 import subprocess
+from collections.abc import Callable
 from datetime import date, datetime, timezone
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


deruyter92 (approved):
Looks good!

PR #3287, opened 2026-04-21
Introduce mdformat pre-commit hook

C-Achard:
## Summary

- Adds `mdformat` v1.0 as a pre commit hook with the myst plugin. This helps automate formatting of documentation and catch any mistakes in structure or syntax, making docs maintenance easier.
- Fixes a small issue with metadata checks being incorrectly run on docs audits files in tools/ .

## Scope

- As a hook with write permissions, it is meant to run locally, not in CI


C-Achard:
One annoyance to note with this : if one has e.g. a technically invalid bullet list which still renders fine(e.g. because of extra newlines), this might try to "fix" it by making every bullet a 1.
I assume there may be other places where the ruleset fixes what isn't broken, which puts a bit more pressure on careful review. Worth thinking about before merging.

deruyter92 (approved):

C-Achard:
Note: this has been extensively used in docs audits (e.g. #3299) and is very useful in enforcing standard md formatting.
Definitely recommend merging when we can.

PR #3290, opened 2026-04-21
[Docker] Simplify and modernize Dockerfile

### LINKED ISSUE #3282: Simplify and modernize DeepLabCut Docker images and build pipeline
**Summary**
Rework the docker/ images and build flow so public tags match real use cases, reduce overlap between base and core, align installs with released versions, and replace the build.sh matrix with a maintainable, CI-friendly approach.

**Problem**
- base vs core differ only by an unused env variable `DLClight` and a MOTD; end-user tooling (deeplabcut_docker.sh) only uses core and jupyter, so publishing and documenting base separately adds little value.
- Version skew: base installs DLC from Git, but not tagged DLC versions.
- Build script: build.sh needs updating. It builds via echoed args and xargs, copying examples/ into docker/ though Dockerfiles do not need it for the image build; helper greps may not match image names.
- CUDA × stage matrix misses clear policy for which variants must stay supported.

**Proposed direction**
- Images: prefer a single primary runtime image. Drop `base`, and optionally keep `jupyter` if it is still used.
- Dockerfile layout: One multi-stage Dockerfile explicit targets instead of four parallel Dockerfiles and bash script
- Install source: Install deeplabcut==${DEEPLABCUT_VERSION} from PyPI (or GH release) so tags match shipped code

**Acceptance criteria (suggested)**
- Public Hub tags and README describe at most runtime + jupyter (plus test if still published), with base removed from public docs.
- Image DLC version matches the tag (no silent “always main”).
- Developers can still build/push via one documented command, but replacing the bash script with modernized approach (e.g. single multi-stage Dockerfile).

### LINKED ISSUE #3285: deeplabcut-docker notebook throws permission error

### Operating System

Setup:
Ubuntu 24.04 with Nvidia GPU
deeplabcut-docker bash --gpus all works perfectly.
deeplabcut-docker notebook starts the notebook server, but it cannot load any files. Throws a 'permission denied' error for /app.

Image

### DeepLabCut version

deeplabcut-latest-gui-jupyter:latest 

pytorch

### DeepLabCut mode

single animal

### Device type

VIDIA RTX PRO 4000 Blackwell (not relevant)

### Bug description 🐛

Upon launching the dockerized image I should be able to open / create notebooks in the user directory (or the specified directory). Instead, the notebook interface opens but is unable to access the /app folder, and therefore does not sow any files, nor allow creation of files.

### Steps To Reproduce

_No response_

### Relevant log output

```shell

```

### Anything else?

_No response_

### Code of Conduct

deruyter92:
This PR is part 1 of updating the docker setup and only entails the Dockerfiles and how we build them. The `deeplabcut-docker` is also updated, but that is discussed separately in PR #3291  to avoid mixing topics.

adresses #3282, #3285

**Motivation**
- Currently we have many dockerfiles with overlapping (or functionally identical) purpose. This PR replaces that with a single Dockerfile, which is easier to maintain and more transparent for users.
- This PR simplifies the variety of image tags to a plain **`latest`**, **`latest-jupyter`** and **`3.0rc14-core-cuda12.4`** or similar. 
    - `cuda12.4` is kept as default which should run on basically most machines from the last 5 years (including newer ones thanks to backward compatibility).
    - Older DeepLabCut (pre-)releases remain on DockerHub with the appropriate tag.
    - `<cudnn_version>` is dropped from the name, since all were using the same. 
    - `base` is dropped as image version since it is equivalent to `core`. 
    - `cuda 12.1` is dropped as default, and `cuda11.8` will remain on DockerHub to support those with legacy hardware. 
- Images were not strictly matching the PyPI versions. This is changed in this PR for transparency.
- Building docker images was done via (somewhat fragile) shell scripts. This PR replaces that with CI-friendly builds using Docker bake / buildx
- The tools for Dockerfiles (targeting Docker users) and the `deeplabcut-docker` package (for deeplabcut-docker developers) was not clearly separated. The package is now moved to its own subdirectory.

**Changes**
- [x] **Single multi-stage docker/Dockerfile**: replaces separate Dockerfile.base / core / jupyter / test. 
- [x] **Versioned tags** like `{version}-core-cuda{CUDA}` (no cudnn suffix in bake tags).
- [x] **pip install "deeplabcut[modelzoo,wandb]==${DEEPLABCUT_VERSION}"** instead of git main
- [x] **Buildx bake defines core + jupyter**, shared build-args
- [x] **MARK_LATEST** parameter  to flag optional latest / latest-jupyter tag.
- [x]  **docker/docker-bake.hcl** instead of docker/build.sh
- [x] **move deeplabcut-docker package** to its own subdirectory, separating it from the dockerfile logic.

**Not included:**
- optional: add CI workflow for building docker images on release (should be easy to add now). Can be added in a separate PR.


docker/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+      libglib2.0-0 \
+      build-essential \
+      git \
+      ffmpeg \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN pip install --upgrade pip && \
+    pip install "deeplabcut[modelzoo,wandb]==${DEEPLABCUT_VERSION}"
+
+RUN chmod a+rwx -R \
```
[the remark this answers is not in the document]
deruyter92: adressed

docker/docker-bake.hcl
```diff
@@ -0,0 +1,47 @@
...
+variable "DLC_VERSION" {
+  default = "3.0.0rc14"
+}
+variable "CUDA_VERSION" {
+  default = "12.4"
+}
+variable "CUDNN_VERSION" {
+  default = "9"
+}
+variable "PYTORCH_VERSION" {
```
C-Achard, line 13: Should we add this to the image tag ? Currently two images with different torch versions would be released as the same. I guess it's okay if we shift torch compat in regards to DLC_VERSION, but still disentangling could be worth it for the future.

docker/Dockerfile
C-Achard, line 1: Should this run as root or do we want a defined USER ?

docker/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+RUN pip install --upgrade pip && \
+    pip install "deeplabcut[modelzoo,wandb]==${DEEPLABCUT_VERSION}"
+
+# Create pretrained weights directory and make it writable
+# (same default as HuggingFaceWeightsMixin in backbones/base.py: Path(__file__).parent / "pretrained_weights")
+RUN PRETRAINED=$(python -c "from pathlib import Path; import deeplabcut.pose_estimation_pytorch.models.backbones.base as b; print(Path(b.__file__).parent / 'pretrained_weights')") && \
+    mkdir -p "$PRETRAINED" && \
+    chmod a+rwx -R "$PRETRAINED"
+
+RUN mkdir -p /app && chmod a+rwx /app
```
C-Achard, line 32: Is this because of #3285 ? It's the exact same path that appears in the issue, and it really looked like perm issues. Did you broaden this as a result or was it already like this ?

docker/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+
+WORKDIR /app
+
+# ── jupyter ───────────────────────────────────────────────────────────────────
+FROM core AS jupyter
+
+RUN pip install "notebook<7"
+EXPOSE 8888
+ENTRYPOINT ["jupyter", "notebook", \
+    "--no-browser", "--NotebookApp.token=deeplabcut", "--ip", "0.0.0.0"]
```
C-Achard: Is it okay to hardcode the token ?

docker/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+ENV CUDA_VERSION=${CUDA_VERSION}
+ENV PYTORCH_VERSION=${PYTORCH_VERSION}
+ENV DEEPLABCUT_VERSION=${DEEPLABCUT_VERSION}
+
+WORKDIR /app
+
+# ── jupyter ───────────────────────────────────────────────────────────────────
+FROM core AS jupyter
+
+RUN pip install "notebook<7"
```
C-Achard, line 49: A small tangential comment : apparently one can use HEALTHCHECK to ensure the EXPOSE started correctly. Not mandatory, but perhaps worth adding as an extra validation, if easy.
```
HEALTHCHECK --interval=30s --timeout=5s --start-period=20s --retries=3 \
CMD python -c "import urllib.request; urllib.request.urlopen('http://127.0.0.1:8888/api/', timeout=3).read()" || exit 1
```

C-Achard (approved):
Looks great, very nice consolidation !

Most comments below are about changing existing behavior rather than improving the layout, and may have been handled elsewhere; please filter accordingly.

docker/Dockerfile
deruyter92: it was already in place. Only difference is that I use python now to check the pretrained_weights path, which seems slightly less fragile.

docker/Dockerfile
deruyter92: good suggestion, I've added it!

docker/Dockerfile
deruyter92: On the one hand: there is currently no pretence of secrecy (it is also in the public docs, DockerHub etc). 
On the other: I agree that we should at least mention this, and offer the opportunity for safer alternatives.

- I've added the option to define a custom `NOTEBOOK_TOKEN` env var in `docker run`.
- I have added a warning now in the README, which should help users refrain from exposing the port to the internet.

docker/Dockerfile
deruyter92: This is a valid concern, but I think for now we should keep the current configuration (root in the base image):
- Downstream images that need to create a user will encounter issues when this base image already switched to a non-root user. 
- The current deeplabcut-docker implementation is an example: it creates a thin wrapper image with a custom user. Implementig a custom user in this base image would break that workflow.
- Downstream images that just want to run as a fixed non-root user can simply add USER themselves as the last line of their own Dockerfile.

A middle ground worth considering is creating a generic named non-root user and document that downstream users need to override it if they need a different UID. I would prefer to keep the current configuration for now since the benefits don't outweigh breaking the current workflow IMO. But again, a valid concern and happy to change it if you guys think otherwise.

deruyter92 (commented):
Thanks @C-Achard, see my response below! Mostly accepted all suggestions, thanks for the helpful review!

docker/docker-bake.hcl
deruyter92: I am unsure about this: 

If we normally just build 1 image per DeepLabCut version, the torch version should be pinned for that specific version. Do you foresee that we will build DeepLabCut Docker images for multiple torch versions? 

The only variant that I foresee will be different is the cuda version: we can support different images for different machines.


I would be in favor of keeping the name as short as possible, and just publishing a small set of docker images for every DeepLabCut release. Pinning our own preferred dependencies for that version, instead of publishing images for different torch versions.

I see your argument, and will happily change it if you disagree. 

(Note to self: this change would need to be propagated to deeplabcut-docker as well).

docker/Dockerfile
C-Achard: Sounds good, thanks for clarifying!

docker/docker-bake.hcl
C-Achard: Okay, if we keep version pins I agree that is much easier. Let's keep it simple. Thanks!

C-Achard (approved):

MMathisLab (approved):
:)

PR #3291, opened 2026-04-21
[Docker 2] update `deeplabcut-docker` package

### LINKED ISSUE #3282: Simplify and modernize DeepLabCut Docker images and build pipeline

### LINKED ISSUE #3285: deeplabcut-docker notebook throws permission error

deruyter92:
This is Part 2 of updating the Docker setup, only concerning the `deeplabcut-docker` package. This builds on top of earlier refactors of the dockerfiles, discussed in PR #3290 

fixes #3282, #3285

**Motivation:**
The deeplabcut-docker package hasn't been actively maintained for some time and currently it is not working for the jupyter images on Docker Hub. This PR follows PR #3290 in simplifying the organization of Dockerfiles and naming, replaces vendored shell script with pure python, and modernizes the setup via pyproject.toml.


docker/package/pyproject.toml
```diff
@@ -1,3 +1,34 @@
...
+]
+requires-python = ">=3.10"
+classifiers = [
+  "Operating System :: MacOS",
+  "Operating System :: POSIX :: Linux",
+  "Programming Language :: Python :: 3 :: Only",
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
```
C-Achard: Should we support these ? (Might be fine since this is building/launching vs running DLC which would be in the container)

Edit : I meant 3.13-3.14

docker/package/deeplabcut_docker.py
```diff
@@ -1,75 +1,176 @@
...
-    argv = ["deeplabcut_docker.sh", launch_args.container, *docker_arguments]
-    print(_MOTD, file=sys.stderr)
-    pty.spawn(argv)
-    print("Container stopped.", file=sys.stderr)
+    run = _docker() + ["run", "-it", "--rm", "-v", f"{os.getcwd()}:/app", "-w", "/app"]
+    if mode == "notebook":
+        port = os.environ.get("DLC_NOTEBOOK_PORT", "8888")
+        _log("Starting the notebook server.")
+        _log(f"Open your browser at http://127.0.0.1:{port}")
+        _log("If prompted for a password, enter 'deeplabcut'.")
```
C-Achard: The image config uses a Jupyter token rather than a password. Would it be clearer to say “token” here?

```suggestion
        _log("If prompted for a token, enter 'deeplabcut'.")
```

docker/package/deeplabcut_docker.py
```diff
@@ -1,75 +1,176 @@
...
-/  / _ \ ___  ___  ___   / /  ___ _ / /  / ___/__ __ / /_ \
-\ / // // -_)/ -_)/ _ \ / /__/ _ `// _ \/ /__ / // // __/ /
-//____/ \__/ \__// .__//____/\_,_//_.__/\___/ \_,_/ \__/  \
-\_________________________________________________________/
-                       ___)( )(___ `-.___.
-                      (((__) (__)))      ~`
 
-Welcome to DeepLabCut docker!
-"""
+def _assert_jupyter_image(ref: str) -> None:
```
C-Achard: Given the docs say :
> To use a specific image instead of the default tags, pass `--image repo:tag`.
> Make sure that the image supports jupyter notebooks when passing `notebook`.

This check might not be entirely correct in that some images that do have entrypoints e.g. via jupyter-lab could be rejected. This could become more complicated than worth it in the end, I can look into more robust checks if you want

docker/package/deeplabcut_docker.py
```diff
@@ -1,75 +1,176 @@
...
+                f"RUN groupadd -g {gid} {group} || groupmod -o -g {gid} {group}",
+                f"RUN useradd -d /home -s /bin/bash -u {uid} -g {gid} {user}",
+                f"RUN chown -R {user}:{group} /home /app",
+                f"USER {user}",
+            )
+        )
+        + "\n"
+    )
+    subprocess.run(
+        _docker() + ["build", "-q", "-t", local, "-"],
```
C-Achard, line 112: Do we want to build every time or cache/reuse ? Is this idempotent ?

docker/package/deeplabcut_docker.py
```diff
@@ -1,75 +1,176 @@
...
+        uid, gid = os.getuid(), os.getgid()
+    except AttributeError:
+        sys.exit("deeplabcut-docker requires a POSIX system (Linux or macOS).")
+    user = pwd.getpwuid(uid).pw_name
+    group = grp.getgrgid(gid).gr_name
+    _log(f"Configuring a local image for user {user} ({uid}) in group {group} ({gid})")
+    dockerfile = (
+        "\n".join(
+            (
+                f"FROM {remote}",
```
C-Achard, line 101: Is there a reason the generated user’s home directory is /home rather than /home/<user>? Maybe a user-specific home dir would be safer for notebook config and caches

C-Achard (commented):
Looks good overall. Some checks could be strengthened (sudo docker check, jupyter validity), but not sure it is worth it given this is meant as a small helper.
I think it's fine to keep the scope tiny.

Also the helper rebuilds a fixed local tag every time. Is that intentional for simplicity, or do we want to avoid repeated rebuilds / potential clashes between concurrent runs?

Let me know what you think is worth addressing, I can approve as-is or we can discuss. Thanks !

docker/package/deeplabcut_docker.py
deruyter92: agreed, thanks for raising this point. I changed it to a way softer approach: just warning if 'jupyter' is missing from the entrypoint without loadly failing. This keeps the user in charge of specifying a valid image without the possiblity of falsely rejecting an image.

docker/package/deeplabcut_docker.py
deruyter92: This has no good reason, and should be changed indeed. Addressed now, thanks for spotting!

docker/package/deeplabcut_docker.py
deruyter92: Thanks to dockers layer caching, the only rebuild is actually the thin Dockerfile wrapper, which should build in less than a second. The slowest part is probably the pull command contacting Docker Hub. While this is slightly wasteful, I don't think it's worth adding caching complexity.

docker/package/deeplabcut_docker.py
deruyter92: Correct, good suggestion thanks! I included your other suggestion here as well: instructions to set your own custom token instead of a publicly known one.

docker/package/pyproject.toml
deruyter92: well, no let's indeed stick to the DLC-supported range. Thanks for correcting this.

deruyter92 (commented):
Thanks @C-Achard  for the elaborate review. Addressed all your comments now.

C-Achard (approved):

PR #3280, opened 2026-04-09
Update napari-DLC docs for refactor

C-Achard:
As https://github.com/DeepLabCut/napari-deeplabcut/pull/175 is approaching completion, this PR updates and expands the napari-DLC relevant docs to be more complete and up-to-date.
Also fixes minor typos in other docs where found.

### TODO

Requires :

- [x] Merge just before napari-DLC release based on https://github.com/DeepLabCut/napari-deeplabcut/pull/175 and any related fixes/PRs

---

### Automated summary

This pull request expands and improves the documentation for the napari-based DeepLabCut GUI, focusing on clearer navigation, more comprehensive guides, and fixing reference errors. 
Two new pages—basic usage and advanced features—have been added for the napari GUI, with links and references updated throughout the docs to point to these new resources. 
Additionally, several typos in internal references have been corrected, ensuring that cross-page links work as intended.

**Documentation Improvements for napari GUI:**

* Added two new documentation pages: `docs/gui/napari/basic_usage.md` (basic usage) and `docs/gui/napari/advanced_usage.md` (advanced features), providing detailed guides and workflows for the napari-based labeling GUI. [[1]](diffhunk://#diff-6feacd7b232a996cc31f817ae520662e2ae42e51de2d2b2008b2076fc8fa7c7dR1-R247) [[2]](diffhunk://#diff-59458ed55f0821d4e9cf3a33146ae84a6e5325ea99c99c0f563ff366c93b6679R1-R73)
* Updated the table of contents (`_toc.yml`) to include the new napari GUI sections, making them accessible from the main navigation.
* Updated links and references in existing guides (e.g., `PROJECT_GUI.md`, `labeling.md`) to point to the new napari GUI landing page and sections, replacing older or broken links. [[1]](diffhunk://#diff-ecf3ef79567c0e351986a03c9da4fbc298f6db1eacb3eb453ff697d3a6cf906eL32-R32) [[2]](diffhunk://#diff-2e621abbff2ac53b3d435bea55e7743b49e76fcb32f233596a4ee13e82553d3aL56-R56)

**Reference and Typo Fixes:**

[...]

PR #3292, opened 2026-04-22
Update and diversify TensorFlow optional installations.

### LINKED ISSUE #3271: Diversify optional installs for TensorFlow stack (LTS)
**Summary**
Currently, there is still no one-solution-for-all for installing DeepLabCut with the TensorFlow extra. While the current matrix takes into account different Python versions and operating systems, users with older CUDA drivers may still encounter installation issues. Having a diverser set of installation options can help to avoid users to have to go and search for a compatible stack of TF versions. This will be particularly relevant in the TF LTS branch. 

**Description**
Installing pip install deeplabcut[tf] on Linux + Python 3.11 currently allows tensorflow>=2.12 with no upper bound, so users can land on very new TF builds (e.g. 2.21) with CUDA runtimes that don’t match their NVIDIA driver. Alternatively, pinning TF 2.12 conflicts with the rest of the stack because TF 2.12 requires NumPy &lt; 1.24 while core DLC allows numpy<2, so typical envs (NumPy 1.26 + pandas 2.2+) break or need manual downgrades.

**Proposed behavior**
- For Linux + Python ≥ 3.11, default the tf extra to a bounded, supported TF line (e.g. 2.15.x) whose NumPy requirements overlap core DLC (numpy>=1.23.5,<2 is compatible with TF 2.15.x and existing DLC deps).
- Optionally add an explicit numpy>=1.23.5,<2 in the tf extra so resolution is predictable.
- Pin companion packages (e.g. tensorpack==0.11, tf-slim==1.1.0, compatible tf-keras).
- Repeat for extras `tf-latest` (matching current specs) , `tf-cu12`, `tf-cu11`. 
- Docs: state minimum NVIDIA driver / link to TensorFlow’s CUDA compatibility for the chosen default; consider a second optional extra (e.g. tf-cuda11 vs tf-cuda12) for machines that cannot use the default wheel.

**Acceptance criteria**
Fresh pip install deeplabcut[tf] on Linux + Python 3.11 pulls a known-good TF + NumPy + backend set without upgrading to unsupported TF or forcing NumPy &lt; 1.24 unless the user opts into `tf-latest`, `tf-cu12` or `tf-cu11`.
pyproject.toml markers are consistent with documented supported stacks. 
.

deruyter92:
### Summary
The current tensorflow installation via the optional extra `tf` is designed as a one-solution-for-all, instead of providing a tailored fit for multiple users with different machines. Also, there is divergence between the recommendations in the docs, CI and the current pyproject.toml. This PR implements a more diverse set of tensorflow installations, and aligns the CI so that these versions are actually tested. The documentation is also updated accordingly.

Addresses #3271 

### Optional dependencies
The new configuration is as follows:
- `tf`: tensorflow>=2.12,<2.18 with all related dependencies -> recommended range that fits the documented last-supported versions and was tested on cuda 11.8 and cuda 12.0. Takes into account the installed python version.
- `tf-cu11` tensorflow==2.14 with all related dependencies -> for cuda 11.x machines.   
- `tf-cu12` tensorflow==2.18 with all related dependences -> for cuda 12.x machines
- `tf-latest` tensorflow>=2.18): 'best effort' support for users that prefer the latest installation (not tested).

### CI
- The separate hardcoded installation of TensorFlow is removed from CI and replaced with installing the `tf` extra.
- The intelligent testing workflow now includes the option to specify custom extras for each job (turned out not to be necessary for tf, but still useful for the future).
- A basic smoke test is implemented for just testing the tensorflow installation with some basic operations: `tests/test_tf_install_smoke`
- The smoke test is hooked up to a job in `intelligent-testing.yaml` testing a matrix of os<>python<>extras. The job is purposely implemented in parallel to the existing lanes (e.g. full testing lane), as it doesn't need downloading ffmpeg etc.

### ~Documentation~
The documentation changes are now moved to a separate PR #3317.


C-Achard (commented):
Looks nice so far; thanks a lot for handling this! t

Two questions regarding the scope :

- How should we handle documentation?
- Can we have minimal CI testing that covers each of these paths ? What do you think would be the best method there ? Maybe a workflow only on main pushes that installs each and runs one of the lighter TF test script ?

deruyter92:
yess working at it! good suggestions

C-Achard:
For the docs happy to coordinate and add this to my ongoing updates!

deruyter92:
Note: adding some smoke tests in CI, which I will probably move to a separate PR later to keep the discussions separated. But first want to test for compatible versioning. 

I will also update the docs once finished and coordinate with you @C-Achard & help with the new docs to avoid parallel work.

deruyter92:
@C-Achard, it's ready for review now. Let me know what you think about the CI changes: for now I added a simple smoke test directly in the intelligent testing workflow, which means we can avoid (downloading and) installing ffmpeg for every run, and also doesn't affect the concurrency with the full-matrix tests. If you think it should make use of the python-package workflow, like the other test lanes, I can see arguments for that as well.

docs/installation.md
C-Achard, line 1: Can we cherry-pick/adapt changes to the latest docs audit updates ?

C-Achard (approved):

C-Achard:
Seems tensorflow-macos is starting to fail in CI, perhaps we should make this high priority

docs/gui/napari_GUI.md
C-Achard, line 1: Note: this has already been replaced by #3280 and can be safely discarded I think

MMathisLab (approved):

deruyter92:
Cleaned up this PR and fixed failing tests:
- cherry-pick #3316 to fix failing CircleCI test (connections issues with new HuggingFace Xet)
- install `dependency-groups` in CI to ensure pip uses `dev` dependency group
- move TF-docs update to separate PR #3317 to resolve merge conflicts with the new docs format.

PR #3295, opened 2026-04-28
Add conditional replacement of `@torch.inference_mode` for inference on AMD DirectML GPUs

### LINKED ISSUE #3289: [Bug] DeepLabCut 3.0 PyTorch with AMD GPU (DirectML) fails on ConvTranspose2d in inference_mode — Workaround: use no_grad

### Operating System

Window 11

### DeepLabCut version

3.0.0rc14 (PyTorch engine)

pytorch

### DeepLabCut mode

single animal

### Device type

AMD Radeon RX 5700 XT

### Bug description 🐛

When running `analyze_videos` using an AMD GPU via the DirectML backend, the process fails with the following error:
`RuntimeError: Cannot set version_counter for inference tensor`
This occurs during the `ConvTranspose2d` operation in the model's head.

### Steps To Reproduce

The issue is caused by the `@torch.inference_mode()` decorator in `pose_estimation_pytorch/runners/inference.py`. Tensors created in `inference_mode` are immutable and have restricted version counting, which is currently incompatible with the DirectML backend for specific convolutional kernels like `ConvTranspose2d`.

### Relevant log output

```shell
Final Verified Workaround
Replacing `@torch.inference_mode()` with `@torch.no_grad()` in the following file/lines fixes the issue and allows the AMD GPU to be utilized correctly via DirectML:
**File:** `deeplabcut/pose_estimation_pytorch/runners/inference.py`
**Lines:** 267 and 682
By switching to `no_grad`, the tensors are compatible with the DirectML execution path, enabling full hardware acceleration for AMD users.
```

### Anything else?

_No response_

### Code of Conduct

deruyter92:
**Motivation**
Currently, our inference runners use `@torch.inference_mode`, which is not supported for AMD GPUs with DirectML inference mode. Essentially `@torch.inference_mode` is a stricter version of `@torch.no_grad`, which is newer and faster and works for most users. However, since it does not work for AMD DirectML users. It would be worthwhile to conditionally replace it with `@torch.no_grad` when necessary.

solves #3289 

**Changes**
This PR replaces `@torch.inference_mode` with a conditional `@_no_grad_decorator`, controlled by the env variable `DLC_DIRECTML_NO_GRAD`. The decorator resolves to `@torch.no_grad` for DirectML users if they set the env var to "true", otherwise it defaults to `@torch.inference_mode` (keeping current behavior).

### DIFF
--- a/deeplabcut/pose_estimation_pytorch/runners/inference.py
+++ b/deeplabcut/pose_estimation_pytorch/runners/inference.py
@@ -40,7 +40,7 @@ from deeplabcut.pose_estimation_pytorch.runners.dynamic_cropping import (
 )
 from deeplabcut.pose_estimation_pytorch.task import Task
 
-# NOTE 2026, April: AMD GPUs with DirectML inference mode currently do not
+# NOTE @deruyter92 2026-04-28: AMD GPUs with DirectML inference mode currently do not
 # support torch.inference_mode, which is stricter than torch.no_grad. The ENV
 # variable is used to conditionally use torch.no_grad instead. See PR #3295.
 _directml_no_grad: bool = os.getenv("DLC_DIRECTML_NO_GRAD", "false").lower() in (
--- a/deeplabcut/pose_estimation_pytorch/runners/inference.py
+++ b/deeplabcut/pose_estimation_pytorch/runners/inference.py
@@ -10,6 +10,7 @@
 #
 from __future__ import annotations
 
+import os
 import threading
 import warnings
 from abc import ABCMeta, abstractmethod
@@ -38,6 +39,22 @@ from deeplabcut.pose_estimation_pytorch.runners.dynamic_cropping import (
     TopDownDynamicCropper,
 )
 from deeplabcut.pose_estimation_pytorch.task import Task
+
+# NOTE: JR 2026-04-28: AMD GPUs with DirectML inference mode currently do not
+# support torch.inference_mode, which is stricter than torch.no_grad. The ENV
+# variable is used to conditionally use torch.no_grad instead.
+_directml_no_grad: bool = os.getenv("DLC_DIRECTML_NO_GRAD", "false").lower() in (
+    "true",
+    "1",
+)
+
+
+def _no_grad_decorator(fn):
+    """
+    Conditional decorator for inference mode, controlled by the DLC_DIRECTML_NO_GRAD ENV variable.
+    Uses @torch.no_grad if set to "true", otherwise defaults to @torch.inference_mode.
+    """
+    return torch.no_grad()(fn) if _directml_no_grad else torch.inference_mode()(fn)
 
 
 def _merge_defaults(cls, data: dict[str, Any]):
@@ -268,7 +285,7 @@ class InferenceRunner(Runner, Generic[ModelType], metaclass=ABCMeta):
             the predictions for each of the 'batch_size' inputs
         """
 
-    @torch.inference_mode()
+    @_no_grad_decorator
     def inference(
         self,
         images: (Iterable[str | Path | np.ndarray] | Iterable[tuple[str | Path | np.ndarray, dict[str, Any]]]),
@@ -648,7 +665,7 @@ class CTDInferenceRunner(PoseInferenceRunner):
         self._idx_to_id = None
         self._ctd_track_ages = None  # the age of each CTD tracklet
 
-    @torch.inference_mode()
+    @_no_grad_decorator
     def inference(
         self,
         images: (Iterable[str | Path | np.ndarray] | Iterable[tuple[str | Path | np.ndarray, dict[str, Any]]]),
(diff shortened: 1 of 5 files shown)


deeplabcut/pose_estimation_pytorch/runners/inference.py
```diff
@@ -39,6 +40,22 @@
...
+    "1",
+)
+
+
+def _no_grad_decorator(fn):
+    """
+    Conditional decorator for inference mode, controlled by the DLC_DIRECTML_NO_GRAD ENV variable.
+    Uses @torch.no_grad if set to "true", otherwise defaults to @torch.inference_mode.
+    """
+    return torch.no_grad()(fn) if _directml_no_grad else torch.inference_mode()(fn)
```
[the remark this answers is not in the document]
deruyter92: This was a conscious choice:

- It **consistently** matches the existing `TORCH_LOAD_WEIGHTS_ONLY` pattern in `base.py`.
- The grad context is selected once, yielding **zero overhead** instead of checking every call to `inference()`.
- Users set `DLC_DIRECTML_NO_GRAD` in their shell before launching Python, not mid-session. The choice is based on the user's machine. There's no realistic scenario where you'd want it to flip during a run.

deeplabcut/pose_estimation_pytorch/runners/inference.py
```diff
@@ -39,6 +40,22 @@
...
+
```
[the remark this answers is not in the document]
deruyter92: changed now to `_inference_mode_decorator`, which resembles the default `@torch.inference_mode`, and IMO is less vague than 'conditional inference'.

I also like the original `_no_grad` terminology, as both modes essentially just disable gradient computations, and it is clearly recognizable for PyTorch users..

deeplabcut/pose_estimation_pytorch/runners/inference.py
```diff
@@ -39,6 +40,22 @@
 )
 from deeplabcut.pose_estimation_pytorch.task import Task
 
+# NOTE: JR 2026-04-28: AMD GPUs with DirectML inference mode currently do not
+# support torch.inference_mode, which is stricter than torch.no_grad. The ENV
+# variable is used to conditionally use torch.no_grad instead.
+_directml_no_grad: bool = os.getenv("DLC_DIRECTML_NO_GRAD", "false").lower() in (
+    "true",
+    "1",
+)
```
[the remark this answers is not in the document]
deruyter92: done

deeplabcut/pose_estimation_pytorch/runners/inference.py
```diff
@@ -39,6 +40,22 @@
 )
 from deeplabcut.pose_estimation_pytorch.task import Task
 
+# NOTE 2026, April: AMD GPUs with DirectML inference mode currently do not
```
C-Achard: I'm just being obsessive there, so feel free to ignore, but should we consistently format notes as:
`# NOTE @ author year-month-day ` ? 
(It's not crucial, and we can agree on a different format)

```suggestion
# NOTE @deruyter92 2026-04-28: AMD GPUs with DirectML inference mode currently do not
```

deeplabcut/pose_estimation_pytorch/runners/inference.py
C-Achard, line 1: Should we look into implementing a custom error message when `RuntimeError: Cannot set version_counter for inference tensor` occurs, which suggests toggling the env var as an potential workaround ?

pyproject.toml
```diff
@@ -113,6 +113,7 @@ Documentation = "https://deeplabcut.github.io/DeepLabCut/README.html"
 [dependency-groups]
 dev = [
   "coverage",
+  "importlib",
```
C-Achard: It's part of the standard library no ? Is there a particular reason ?

C-Achard (approved):
Good fix, thanks! 
Just minor comments, otherwise LGTM

deeplabcut/pose_estimation_pytorch/runners/inference.py
deruyter92: perfect. Agreed!

pyproject.toml
deruyter92: My mistake, thanks for the correction. Restored uv.lock and pyproject.toml now.

deeplabcut/pose_estimation_pytorch/runners/inference.py
deruyter92: Great suggestion! I have added an error hint now. It does not test the specific hardware, but just augments any RuntimeError now with the following message:

> If you are using an AMD GPU with DirectML, this error may be caused by @torch.inference_mode being incompatible with the DirectML execution path. Try setting the environment variable DLC_DIRECTML_NO_GRAD=true, which will switch the inference context to @torch.no_grad.

BTW, also updated the TechHardware docs with a small hint related to this error.

C-Achard (approved):
Thanks for the docs update!

AlexEMG (approved):
Lgtm

PR #3303, opened 2026-05-04
update `list_videos_in_folder`

### LINKED ISSUE #3300: [Bug] DLC3 cannot process suffix-less video files

### Operating System

Not really applicable, this is a problem in Python logic, but nevertheless, applies on each of:
Windows 11 25H2
Ubuntu 20.04, 22.04, 24.04

### DeepLabCut version

3.0.0rc14

pytorch

### DeepLabCut mode

single animal

### Device type

Not really applicable (error arises before any inference takes place, but nevertheless, applies on each of:
CPU (Xeon w5-2445, Ryzen 7 360, generic virtualised Haswell-era Intel)
GPU (RTX A4000, A100-40)

### Bug description 🐛

We implement a shared DLC service which pulls data from the common s3 storage array and runs the requested model. Videos are downloaded and cached to the GPU host. Since users often re-run the same video within a day or two, videos are cached on the GPU host for 72 hours. 

Caching uses the standard pattern of assigning a globally unique hash to the downloaded file (so, e.g., `s3://bucket/data/subject/session/video_0001.mp4` becomes `/cache/bucket/abcd1234`, with the real names stored in a hash map for later reference. 

Under DLC2.3.11, this works as expected: the output directory looks like so:
```
/results/job-id
    abcd1234_meta.pickle
    abcd1234.h5
```
These files are renamed appropriately when sent back to s3 storage, of course. 

Under DLC3.0.0rc14, however, this fails for both Tensorflow and Pytorch based models. The log looks like so:
```python
>>> deeplabcut.analyze_videos(
            config=dlc_cfg_filepath,
            videos=video_filepaths,
            shuffle=dlc_model['dlc_shuffle'],
            trainingsetindex=dlc_model['dlc_trainingsetindex'],
            destfolder=output_dir,
)
Analyzing videos with <path to model>
 -> Using detector <path to model>
Using scorer: DLC_Hrnet32_...
No .h5 files were created during video analysis. Please check your code and ensure that the video inference and output generation are correct.
```
Inference finishes very quickly (~5s instead of 20-40 minutes) with, as the log indicates, no output.
    

### Steps To Reproduce

I have traced this down to the two functions used to enumerate videos.

Suppose I have the following directory:
```bash
/home/user/videos
    video_0001.mp4
    abcd1234
```
Where `abcd1234` is a copy of `video_0001.mp4`, renamed to a hash value comparable to usage on our GPU host (except with 8 digits instead of the default 64)

**For Pytorch:**
`deeplabcut.pose_estimation_pytorch.apis.videos.analyze_videos` calls (currently line #543) `deeplabcut.,pose_estimation_pytorch.apis.utils.list_videos_in_folder`
```python
>>> from deeplabcut.pose_estimation_pytorch.apis import utils as dlc_t_utils
>>> videos = dlc_t_utils.list_videos_in_folder(
    data_path = ["/home/user/videos/video_0001.mp4", "/home/user/videos/abcd1234"],
    video_type=None,
    shuffle=False,
)
>>> print(videos)
["/home/user/videos/video_0001.mp4"]
# This is the error: it fails to pick up the hex-named file
```

As of c190b94, the guilty are lines 322 to 332
```python
# deeplabcut.pose_estimation_pytorch.apis.utils lines 319-332
    if not video_type:
        video_suffixes = {f".{ext.lower()}" for ext in auxfun_videos.SUPPORTED_VIDEOS}
    else:
        video_suffixes = {f".{video_type.lstrip('.').lower()}"}

    videos = []
    for path in map(Path, data_path):
        if not path.exists():
            raise FileNotFoundError(f"Could not find: {path}. Check access rights.")

        if path.is_dir():
            videos.extend(f for f in path.iterdir() if f.is_file() and f.suffix.lower() in video_suffixes)
        elif path.is_file() and path.suffix.lower() in video_suffixes:
            videos.append(path)
```

If a) I supply a list of paths and not directories and b) no video_type, then the effect of this function is as follows:
* Enumerate the suffixes _if any_ of the video files supplied
* Filter the list of files provided against that set of suffixes

This is a) 100% pointless and b) cuts off the not-exactly-unusual edge case of suffix-less files. It performs no useful filtering ("does this member of the set have a suffix matching any of the suffixes from the set?"). Ergo, this is a logical bug.

**For Tensorflow**
For reasons I don't fully understand, the logic path for Tensorflow uses a completely indepdent but _nearly_ identical function to enumerate videos. As of DLC3, it seems to have introduced the same logical bug.

`deeplabcut.pose_estimation_tensorflow.predict_videos.analyze_videos` calls (line 597) `deeplabcut.utils.auxiliaryfunctions.get_list_of_videos`

```python
>>> from deeplabcut.utils import auxiliaryfunctions as dlc_u_af
>>> videos = dlc_u_af.get_list_of_videos(
    videos= ["/home/user/videos/video_0001.mp4", "/home/user/videos/abcd1234"],
    videotype=None,
    in_random_order=False,
)
>>> print(videos)
["/home/user/videos/video_0001.mp4"]
#Again, same error. Only identifies one video out of two. 
```

This function implements the same filterslightly differently, and in a way that makes a bit more sense (it applies to a list of videos _either_ supplied by the user _or_ extracted from the directory), but to the same effect: it demands that video files _must have a suffix_, which is the root problem. 
```python
# deeplabcut.utils.auxiliaryfunctions
    if not videotype:
        videotype = auxfun_videos.SUPPORTED_VIDEOS
    # filter list of videos
    videos = [
        v
        for v in videos
        if os.path.isfile(v)
        and any(v.endswith(ext) for ext in videotype)
        and "_labeled." not in v
        and "_full." not in v
    ]
```
(As an aside, this function also does not support paths, but fairly trivially could do so, as and when addressing this more serious bug)

Summary: both functions have the same issue. The logic for enumerating an unknown set of videos from a directory is also applied, unnecessarily, to the list of files provided by the user, and applies a (probably unintentional) assumption by the author, that files necessarily have a suffix. 

My suggestion is that:
a) you can delete one of these functions. As far as I can tell, these two functions are _intended_ to do exactly the same thing (and very nearly do), but with subtly different signatures and duplicating the code into two places
b) Adjust the logic to handle the case where the user may provide a suffix-less file. The best way to handle this (imo!) is to apply the filter logic only to enumeration of a directory. If the user provides a list of files, and no directories, those should be returned immediately, without filtering. 
c) If keeping the more general function ( `deeplabcut.utils.auxiliaryfunctions.get_list_of_videos` ) update it to handle Path objects as well as strings. 

### Relevant log output

```shell

```

### Anything else?

This is the behaviour under Deeplabcut 2.3.11

```python
>>> from deeplabcut.utils import auxiliaryfunctions as dlc_u_af
>>> videos = dlc_u_af.get_list_of_videos(
    videos= ["/home/user/videos/video_0001.mp4", "/home/user/videos/abcd1234"],
)
>>> print(videos)
 ["/home/user/videos/video_0001.mp4", "/home/user/videos/abcd1234"],
# Correct outpout: hands back the same list of files provided by the user
```

Comparing to `deeplabcut.utils.auxiliaryfunctions` in fa2ebee, the 2.3.11 release commit (https://github.com/DeepLabCut/DeepLabCut/blob/fa2ebee4bf9b9723cf32e0a1ffe664de95b9b2e1/deeplabcut/utils/auxiliaryfunctions.py#L334), it looks like the bug was introduced by the addition of this:

```python
    if not videotype:
        videotype = auxfun_videos.SUPPORTED_VIDEOS
```
Although I can't actually find which commit changed this and introduced the bug. 

### Code of Conduct

deruyter92:
### Summary
Currently there exist different functions for collecting video paths from a list of directories and files, with slightly different behavior. See #3300. 

**Old behavior**
- `list_videos_in_folder`  **always** filters the input files by extension: either the default `SUPPORTED_VIDEOS` or custom specified `video_type`.
-  `get_video_list` only filters **directories** by `SUPPORTED_VIDEOS`, but not files.
- `get_list_of_videos` only filters **directories** by `SUPPORTED_VIDEOS`, but not files. 

**Issues**
 - Multiple functions with inconsistent behavior
 - No way to collect videos _without extension_
 - Ambiguous input parameters `video_type` and `videotype`: the meanings of `""` and `None` differ per-case. List of strings is not always accepted.
 - Code exists in deep branches where 

### Changes
- The current PR moves and renames the `list_videos_in_folder` function, to have a single centralized helper for this purpose
- The contract is changed to treat files and directories as follows
    - File paths are accepted as-is when ``extensions`` is ``None``; only filtered when
      ``extensions`` is explicitly set.
    - Directory contents are always filtered by extension: by ``SUPPORTED_VIDEOS`` when
      ``extensions`` is ``None``, or by the given value(s) otherwise.
    - ``exclude_patterns`` are always applied to both files and directory contents.

solves #3300 

### Details
**PR Status**
- [x] relocate `list_videos_in_folder` to auxfun_videos: more centralized instead of pytorch-specific
- [x] rename to `collect_video_paths`, which more accurately reflects function. **feedback on naming is welcome!**
- [x] conditional filtering: 
   - treat specified _video files_ differently from video paths collected from _specified directory_
   - files are not filtered by extension by default, only when _explicitly specifying_ extensions to filter for.
   - directories are scanned using `SUPPORTED_VIDEOS` by default or with specified _extensions_ filter if set.
   - All cases are filtered using `exclude_patterns` which is defaults to ["*_labeled.*", "*_full.*"] to match prior DLC outputs.
- [x] add basic testing
- [x] deprecate original `list_videos_in_folder` and map to new `collect_video_paths`
- [x] deprecate `get_list_of_videos` and map to new `collect_video_paths`
- [x] deprecate `get_video_list` and map to new `collect_video_paths`
- [x] keep `grab_files_in_folder`, `get_camerawise_videos` (different implementation) 
 
 
**New function signature**
```python
def collect_video_paths(
    data_path: str | Path | list[str | Path],
    extensions: str | Sequence[str] | None = None,
    shuffle: bool = False,
    exclude_patterns: Sequence[str] = DEFAULT_EXCLUDE_PATTERNS,
) -> list[Path]:
    """
    Collects video paths from a given set of data paths: directories, files, or a mix
    of both. Directories are scanned one level deep (non-recursively).

    Files and directories are treated differently with respect to extension filtering:
    - File paths are accepted as-is when ``extensions`` is ``None``; only filtered when
      ``extensions`` is explicitly set.
    - Directory contents are always filtered by extension: by ``SUPPORTED_VIDEOS`` when
      ``extensions`` is ``None``, or by the given value(s) otherwise.
    - ``exclude_patterns`` are always applied to both files and directory contents.

    Args:
        data_path: Path or list of paths to folders containing videos, or individual
            video files. Can be a mix of directories and files.
        extensions: Controls extension filtering for collected video files.
            - ``None`` (default): file paths are accepted without extension filtering;
              directories are scanned for files with a recognized video extension.
            - ``str`` or ``Sequence[str]`` (e.g. ``"mp4"`` or ``["mp4", "avi"]``):
              both file paths and directory contents are filtered to only include files
              matching the given extension(s).
            - Empty ``str`` ``""`` is treated as ``None`` (deprecated, keep for backwards
              compatibility).
        shuffle: Whether to shuffle the order of videos. If ``False``, videos are
            returned in sorted order for deterministic behavior.
        exclude_patterns: Patterns to exclude from the collection. Defaults to
            ``DEFAULT_EXCLUDE_PATTERNS``. Set to ``[]`` to disable pattern exclusion.

    Returns:
        The paths of videos to analyze. Duplicate paths are removed.

    Raises:
        FileNotFoundError: If any path in ``data_path`` does not exist.
        ValueError: If ``extensions`` is an empty sequence.
    """
 ...
    return unique_videos
```


C-Achard:
Note: CI seems to be failing due to tf-macos, merging #3292 may potentially help

C-Achard:
@deruyter92 Great to add deprecations, definitely agree. Do you think a separate PR for that specifically would be useful, or is it more efficient to merge this directly ?

deeplabcut/utils/auxfun_videos.py
```diff
@@ -643,3 +645,81 @@ def display_help(*args):
 
     plt.close(fig)
     return bbox
+
+
+def collect_video_paths(
+    data_path: str | Path | list[str | Path],
+    extensions: str | list[str] | None = None,
+    shuffle: bool = False,
+    exclude_patterns: list[str] | None = None,
```
C-Achard: While I don't mind having a default, and I understand why it is None, it could be nice to have more strongly advertised default behavior, maybe using a const:
```python
DEFAULT_EXCLUDE_PATTERNS: tuple[str, ...] = ("*_labeled.*", "*_full.*")
```
and switching to:

```suggestion
    exclude_patterns: Sequence[str] = DEFAULT_EXCLUDE_PATTERNS,
```
This is a breaking change if call sites use None, but we can keep None as an accepted value too if needed.

deeplabcut/utils/auxfun_videos.py
```diff
@@ -643,3 +645,81 @@ def display_help(*args):
...
+    Raises:
+        FileNotFoundError: If any path in data_path does not exist.
+    """
+    if isinstance(data_path, (str, Path)):
+        data_path = [data_path]
+
+    if exclude_patterns is None:
+        exclude_patterns = ["*_labeled.*", "*_full.*"]
+
+    def _coerce_extensions(extensions: str | None) -> set[str]:
```
C-Achard:
```suggestion
  def _coerce_extensions(
      extensions: str | list[str] | tuple[str, ...] | None,
  ) -> set[str] | None:

```

deeplabcut/utils/auxfun_videos.py
```diff
@@ -643,3 +645,81 @@ def display_help(*args):
...
+    - exclude patterns are ALWAYS applied for directory contents and supplied files. Set to `[]` to exclude no patterns.
+
+    Args:
+        data_path: Path or list of paths to folders containing videos, or individual
+            video files. Can be a mix of directories and files.
+        extensions: The types of videos to filter for (e.g., ".mp4", ".avi", etc.).
+            - If set: filter all videos with the given extensions. Both for directory contents and supplied files.
+            - If ``None``, provided files are not filtered, but directory contents are filtered by ``SUPPORTED_VIDEOS``.
+        shuffle: Whether to shuffle the order of videos. If False, videos are returned
+            in sorted order for deterministic behavior.
```
C-Achard: > in sorted order for deterministic behavior.

Are they sorted here ?

deeplabcut/utils/auxfun_videos.py
```diff
@@ -643,3 +645,81 @@ def display_help(*args):
 
     plt.close(fig)
     return bbox
+
+
+def collect_video_paths(
+    data_path: str | Path | list[str | Path],
+    extensions: str | list[str] | None = None,
```
C-Achard: Here also I'd argue for slightly clearer intent: if None, filtered by support, OK. 
Currently [] will also result in default behavior. If we want [] to be "accept no extensions", changes would be needed.
Depends on downstream use.

deeplabcut/utils/auxfun_videos.py
```diff
@@ -643,3 +645,81 @@ def display_help(*args):
...
+    for path in map(Path, data_path):
+        if not path.exists():
+            raise FileNotFoundError(f"Could not find: {path}. Check access rights.")
+
+        if path.is_dir():
+            # Discriminate videos from other files; skip excluded patterns (e.g. prior DLC outputs).
+            allowed = explicit_suffixes if explicit_suffixes else implicit_suffixes
+            videos.extend(
+                f
+                for f in path.iterdir()
```
C-Achard, line 733: Minor nitpick: This only scans a single level; maybe document that "directories" search does not include recursive search

deeplabcut/utils/deprecation.py
```diff
@@ -0,0 +1,78 @@
...
+            + (f" since {since}" if since else "")
+            + f"; use '{new}' instead."
+        )
+
+        @functools.wraps(fn)
+        def wrapper(*args, **kwargs):
+            if old in kwargs:
+                warnings.warn(msg, DeprecationWarning, stacklevel=2)
+                kwargs[new] = kwargs.pop(old)
+            return fn(*args, **kwargs)
```
C-Achard, line 150:
```suggestion
        @functools.wraps(fn)
        def wrapper(*args, **kwargs):
            if new in kwargs and old in kwargs:
                raise TypeError(
                    f"{fn.__qualname__} received both '{old}' and '{new}' parameters; use only '{new}'."
                )
            if old in kwargs:
                warnings.warn(msg, DeprecationWarning, stacklevel=2)
                kwargs[new] = kwargs.pop(old)
            return fn(*args, **kwargs)
```

C-Achard: (Otherwise it silently overwrites)

deeplabcut/utils/deprecation.py
```diff
@@ -0,0 +1,78 @@
...
+import functools
+import warnings
+from collections.abc import Callable
+
+
+def deprecated(
+    replacement: str | None = None,
+    since: str | None = None,
+    removed_in: str | None = None,
+) -> Callable:
```
C-Achard:
```suggestion
def deprecated(
    *,
    replacement: str | None = None,
    since: str | None = None,
    removed_in: str | None = None,
) -> Callable:
```
I would prefer forcing keyword use for future extensibility

tests/test_auxiliaryfunctions.py
```diff
@@ -48,6 +48,8 @@ def _create_fake_file(filename):
             auxiliaryfunctions.find_analyzed_data(fake_folder, "video" + str(ind), SCORER, filtered=True)
 
 
+@pytest.mark.deprecated
```
C-Achard, line 51: Nice to add those!

C-Achard (commented):
Really nice work overall! Definitely much better to have the video loading centralized, and deprecations are nice.

I added quite a few comments that I hope will help future us avoid mistakes and make things easier to use, happy to discuss if there are any concerns.

I will open a PR for the deprecations, I have a small design suggestion to ensure we can greatly extend and automate the system later if needed, while keeping current design and lightweight code.

deeplabcut/utils/deprecation.py
deruyter92: Agreed, sounds good

deeplabcut/utils/deprecation.py
deruyter92: tnx, good one. I'll merge your adjustments from #3326.

deeplabcut/utils/auxfun_videos.py
deruyter92: Agreed, this is better

deeplabcut/utils/auxfun_videos.py
deruyter92: The `extensions` parameter is now updated to yield the following groups of behavior:

1. **Default**
    - `None` (default) -> only filter directory contents using `SUPPORTED_VIDEOS`, but accept all provided files without filtering
    - `""` -> same behavior as `None` (Required for backward compatibility)
2. **Only select specified extensions**
    - Sequence of str, e.g. `(".avi", ".mp4")` -> filter both files and directories, collect only paths with ".avi" suffix or ".mp4" suffix
    - String e.g. `".avi"` -> identical to single-length sequence: `(".avi")`
3. **Only select no-extensions**
    - Empty list / tuple `[]` -> only select files without extension

deruyter92: My prior behavior just treated all Falsy values, including "" the same as None. But this would only yield option 1 and 2. 

Maybe the current update is a bit too complex, but at least it supports all cases.

AlexEMG (approved):
Looks good to me.

deeplabcut/utils/auxfun_videos.py
C-Achard: I think it's good, thanks !

C-Achard (approved):

PR #3310, opened 2026-05-06
Increase CircleCI timeout for tests to 30m

C-Achard:
Add `no_output_timeout: 30m` to the run step for the TestDLC job in .circleci/config.yml to prevent CircleCI from timing out during long-running testscript_cli.py executions, specifically during HF model download. Tokenless downloads may have been adjusted downwards, which makes the job consistently hit timeout limits that previously were not a problem.


C-Achard:
Download is still broken. Wait to see if the download issues are solved (works fine locally), consider caching model weights

Issue #3312, opened 2026-05-06
[task] `TrainingDatasetMetadata`: improve robustness and separation of concerns for incomplete projects


deruyter92:
**Problem**
`TrainingDatasetMetadata` and `ShuffleMetadata` in _deeplabcut/generate_training_dataset/metadata.py_ have grown into classes with unclear contracts and conflated responsibilities. This causes several problems:

- _Inference fails on incomplete projects_: For example, inference on [this example project](https://zenodo.org/records/7883589) is currently not possible and [this demo notebook](https://github.com/DeepLabCut/DeepLabCut/blob/main/examples/COLAB/COLAB_3miceDemo.ipynb) is therefore broken. 
- _Error messages can be uninformative_: e.g. when seeing `Could not find a shuffle with trainingset fraction 0.95 and index 0`, the user has no way to know whether the metadata is empty, stale, or simply using the wrong index.
- _No documented contract_ in  `ShuffleMetadata.split: DataSplit | None` for when None is valid. E.g. `save()` crashes on None-split entries (and `load_split()` on a frozen dataclass is also a bit awkward).

**Acceptence criteria**
- Separate focused classes with clear contracts, separating scanning, merging and registry persistence responsibilities.
- Engine correctly inferred from folder structure as fallback, instead of only pickle / yaml.
- Inference works on model-only projects.

deruyter92:
Partially solved by #3313, but keeping this issue open as a more durable solution is still requested for project-status management (see first criterium).

PR #3313, opened 2026-05-06
Improve `TrainingDatasetMetadata` and `get_shuffle_engine` for incomplete projects

deruyter92:
**Motivation**
see #3312: currently `get_shuffle_engine` is broken for projects with shuffles are not stored in metadata under training-datasets. While it can be a valid choice to make this metadata file the single source of truth, it currently breaks backwards compatibility, since for older DLC versions, only a model folder was sufficient for inference.

**Proposed changes:**
The current PR aims to implement small changes that 
- allow fallback for `get_shuffle_engine` to inferring the engine from the model folder. (This fixes the currently broken `analyze_videos` for projects that only have a model folder, but no training dataset)
- improve error messages, so users can diagnose if their project is incomplete / if they just specified a non-existent shuffle. 
- don't write an empty metadata file when the metadata was not found in the first place (this made the state of the project even more corrupt than it was)

Note that more refactors could be implemented to guarantee long-term maintainability of this code (e.g. separate responsibilities and clear contracts - again see #3312) but the scope of the current PR is just some easy wins. 

- [x] improve error messages in TrainingDatasetMetadata
- [x] implement fallback for `get_shuffle_engine` based on folder structure
- [x] add tests

### DIFF
--- a/deeplabcut/generate_training_dataset/metadata.py
+++ b/deeplabcut/generate_training_dataset/metadata.py
@@ -191,14 +191,24 @@ class TrainingDatasetMetadata:
             the shuffle with the given trainset index and shuffle index
 
         Raises:
-            ValueError if the shuffle is not present in the metadata
+            ValueError if trainset_index is out of bounds or the shuffle is not present
         """
-        train_fraction = self.project_config["TrainingFraction"][trainset_index]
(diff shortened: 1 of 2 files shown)


C-Achard (approved):
Very nice, I like the fact that this prevents writing partial metadata.

PR #3323, opened 2026-05-11
Robustness fix: Annotation file not dropping likelihood column if present from machine labels 

### LINKED ISSUE #3319: Shuffle creation failure
> I've just tried to create a new train/test split using the edited labels. It created an iteration-0 folder with CollectedData csv and h5 file and metadata yaml file but DLC is now hanging. Previously I saw it also created a Documentation data pickle file and Matlab file. When I go to train the network, it says "failed to read shuffle 1". 
> 
> FYI when I first tried to create the train/test split it complained that there were duplicates of the videos in the config file - not sure why, but I edited the pathing to fix it. 
> 
> Not sure what the issue is - all video folders have a CollectedData .csv and .h5 file now, plus some have machine labels too. Let me know if you'd like me to send over the config file or an example folder. 

 _Originally posted by @d45ha in [napari-deeplabcut #189](https://github.com/DeepLabCut/napari-deeplabcut/issues/189#issuecomment-4406805461)_

C-Achard:
### Issue

The latest napari-deeplabcut version retains the likelihood column when refining machine annotations (they are added to the CollectedData h5 if present).
The h5/csv concat & matlab file creation function always assumed only x and y are present, and did not filter out likelihood, leading to dataset creation failure.
Note: this raises in terminal but could cause the GUI to hang instead, as reported in #3319.

### Related

- [x] Should napari-deeplabcut always discard likelihood when saving to CollectedData ? -> Yes
  - [x] Fixed plugin-side in https://github.com/DeepLabCut/napari-deeplabcut/pull/204
- [x] Do we want to systematically filter at the training set h5/csv creation step ? -> not needed due to above point
- [ ] Aims to close #3319

### Fix

Improves the robustness of the training data formatting by ensuring that any "likelihood" columns present in the input DataFrame are dropped before .mat formatting, and adds a corresponding test to verify this behavior. 

**Data formatting fixes:**

* Updated `format_training_data` in `trainingsetmanipulation.py` to automatically detect and remove "likelihood" columns from the DataFrame before processing, ensuring only "x" and "y" coordinates are used. Added validation to require both "x" and "y" columns and to check that the number of coordinate values per row is even.
* Improved error handling for malformed data, raising clear exceptions when required coordinate columns are missing or when the data shape is unexpected.

**Testing:**

* Added a new test `test_format_training_data_ignores_likelihood_columns` in `test_trainingsetmanipulation.py` to verify that the presence of "likelihood" columns does not affect the output of `format_training_data`.


C-Achard:
Note: after discussion with @deruyter92, it has been decided not to retain likelihood columns in CollectedData, as previously. See https://github.com/DeepLabCut/napari-deeplabcut/pull/204.

deruyter92 (approved):

deruyter92:
> Should napari-deeplabcut always discard likelihood when saving to CollectedData?


For future reference we concluded yes for the following reasons: 
- Anything reviewed by human-labeling should be considered ground-truth labels 
- Keeping likelihood in the CollectedData leaves the wrong impression that the data is uncertain or unreliable, while these scores are only applicable to the NN-based judgements, not the human labels. 
- Since the likelihood scores for machine-labels are also stored separately, there is no need to also keep them here, and we can safely remove these from the CollectedData.

PR #3332, opened 2026-05-15
Resolve inconsistent parameter names via aliasing + deprecationwarning.

deruyter92:
## Standardize public API parameter names
This PR cleans up a series of inconsistent parameter names across the public DeepLabCut API. All renames are fully backward-compatible: the old names
are preserved as deprecated aliases via the existing `@renamed_parameter` decorator, which emits a
`DLCDeprecationWarning`. 

### Changes

- `Shuffles`, `shuffles` → `shuffles`
- `Snapindex`, `snapshotindex` → `snapshotindex`
- `videotype`, `video_type`, `extensions` → `extensions`
- `batchsize`, `batch_size` → `batch_size`
- `displayiters`, `display_iters` → `display_iters`
- `saveiters`, `save_iters` → `save_iters`
- `maxiters`, `max_iters` → `max_iters`
- `comparisonbodyparts`, `comparison_bodyparts` → `comparison_bodyparts`

All docstrings and inline examples have been updated to use the new names.


deeplabcut/compat.py
C-Achard, line 1: Definitely like this one, we have several inconsistencies throughout between snake/camel case and this makes the API easier imo.

C-Achard (commented):
Really nice overall !

I just have a few broader questions to make sure this stays "healthy" on the longer term:

-  Can we handle non-keyword args safely with this ? Are there still risks of collisions/improper usage for users ?
- I think we should disallow "chaining" renames if possible, ie. videotype -> video_type is fine, but videotype -> video_type -> extensions raises, if possible. Not 100% needed. docs could suffice
- I would favor having super-clear rules on the decorator itself. e.g. : Always use the new version in code and never the old. Do not chain rename or stack, etc.
- We could also systematically test these, but not crucial

I don't mean to keep this blocked though, I just think a bit of documentation would be perfect, no need to go much further yet (automation etc)

MMathisLab:
merge conflicts need fixing @deruyter92

deruyter92:
@MMathisLab I have resolved all conflicts and rebased on main now

deruyter92:
@C-Achard, currently addressing your review, valid points thanks!

deruyter92:
@C-Achard 

> 1. Can we handle non-keyword args safely with this ? Are there still risks of collisions/improper usage for users ?
> 2. I think we should disallow "chaining" renames if possible, ie. videotype -> video_type is fine, but videotype -> video_type -> extensions raises, if possible. Not 100% needed. docs could suffice
> 3. I would favor having super-clear rules on the decorator itself. e.g. : - - Always use the new version in code and never the old. Do not chain rename or stack, etc.
> 4. We could also systematically test these, but not crucial

1.  Positional arguments currently bypass the decorator by design, which is the correct behavior. Changing the order of positional arguments still breaks the API as it would normally (now documented).
2. I agree, this is potentially dangerous when misused. This case is easy to intercept and I've added a guard now. Also added an extra guard for typos in the name of the `new` parameter.
3. Agreed. Added rules in the docstring now. Let me know if you agree.
4. Added some tests!

C-Achard:
Amazing thank you @deruyter92 !

C-Achard (approved):

deruyter92:
(rebased again, removing empty diffs)

deeplabcut/pose_estimation_pytorch/apis/tracking_dataset.py
```diff
@@ -125,11 +126,12 @@ def extract_features_for_video(
     shelf_writer.close()
 
 
+@renamed_parameter(old="videotype", new="extensions", since="3.0.0")
 def create_tracking_dataset(
     config: str,
     videos: list[str] | list[Path],
     track_method: str,
-    videotype: str | Sequence[str] | None = None,
+    extensions: str | Sequence[str] | None = None,
```
MMathisLab: This feels like a bad change; video type is clear, extensions are broad?

MMathisLab (changes requested):

deeplabcut/pose_estimation_pytorch/apis/tracking_dataset.py
C-Achard: The main ambiguity we were trying to resolve is that we only know the extension on disk (the file name); but not what the actual container format of the video is. 
So videotype sounds like we reliably know the underlying media format, but extensions makes it clear this is filename suffix filtering only, with no other guarantees.

This may be more of a dev-side distinction though, let us know what you'd prefer

C-Achard: Maybe `videos_extensions` would be a good middle ground ?

MMathisLab: I'm in favor of leaving all naming as it was originally for full and easy backwards compatibility. Naming changes are a big pain for users that have automated pipelines

C-Achard: Of course, the decorator makes the aliasing automatic so this is fully backwards compatible:
`@renamed_parameter(old="videotype", new="extensions", since="3.0.0")`

Anyone who uses videotype as arg will be automatically mapped to the new one, this is true for all parameter name changes in this PR, as long as the decorator is present basically.

C-Achard: (Though this approach to renaming could become a problem if not used sparingly, [we tried to make it clear in docstrings](https://github.com/DeepLabCut/DeepLabCut/pull/3332/changes#diff-397d67fdc4cba1c3433d95cd11f145a5cc179a7725a94813677ea1373fd882b5R133))

deruyter92: @MMathisLab,  I understand the importance of not changing too much, that is a valid point. 

My personal stance is that the benefits outweigh the downsides in this case:
- the old parameter name is still accepted. (we can even suppress the warning that is currently emitted).
- `video_type` / `videotype` are inconsistently used throughout the codebase, so changing is needed anyway
- I believe `video_type` to be slightly misleading: 
    - singular despite accepting a list of extensions
    - suggests that the video type will be checked, which is not the case (as @C-Achard mentioned). 
    
My personal preference would go for `extensions`. But again, I understand your argument @MMathisLab. I'll leave the decision up to you. In case not going for `extensions`, would you prefer to stick with `video_type`? Or do you prefer `videotype`? Both are currently used throughout the codebase.

deruyter92: > (Though this approach to renaming could become a problem if not used sparingly, ...)

NB @C-Achard, I think this is actually not a problem now. As long as you don't chain renames - which is picked up by CI, you can basically rename every parameter with full backward compatibility. (Bit extreme, but technically no problem)

deeplabcut/compat.py
```diff
@@ -515,7 +521,7 @@ def evaluate_network(
     if engine is None:
         cfg = _load_config(config)
         engines = set()
-        for shuffle in Shuffles:
+        for shuffle in shuffles:
             engines.add(
                 get_shuffle_engine(
```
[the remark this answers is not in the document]
deruyter92: unrelated to this PR (the type hint was unchanged), but addressed now. It is a good suggestion.

deeplabcut/pose_estimation_pytorch/apis/tracking_dataset.py
C-Achard: My concern is more with renaming an already renamed parameter later

deruyter92: @C-Achard, yes that's valid point! In case we would keep renaming A -> B-> C, we can only support a single remapping B -> C or A -> C . No chaining is permitted.

C-Achard: This would be clean and okay to do though, right ?
```python
@renamed_parameter(old="oldestname", new="newest", since="3.0.0")
@renamed_parameter(old="older_name", new="newest", since="4.0.0")
def f(*, newest): ...
```

C-Achard: Is the check enforcing that the in-degree be strictly 1 (which is not the best imo) or that the sum of in-degrees of all previous nodes is not more than one (which then is okay) ?

C-Achard: Apparently this is formally called an antichain in DAGs, the check is simple:
`old_names` ∩ `new_names` == ∅
Will suggest a small upgrade in review

deeplabcut/utils/deprecation.py
C-Achard, line 119:
```suggestion
def _validate_renamed_parameters(
    *,
    fn: Callable[..., object],
    infos: tuple[DeprecationInfo, ...],
) -> None:
    old_to_new: dict[str, str] = {}

    for info in infos:
        old = info.old_parameter
        new = info.new_parameter

        if old is None or new is None:
            continue

        if old == new:
            raise ValueError(
                f"@renamed_parameter: invalid rename on {fn.__qualname__}: "
                f"{old!r} maps to itself."
            )

        existing_new = old_to_new.get(old)
        if existing_new is not None and existing_new != new:
            raise ValueError(
                f"@renamed_parameter: conflicting aliases on {fn.__qualname__}: "
                f"{old!r} maps to both {existing_new!r} and {new!r}."
            )

        old_to_new[old] = new

    old_names = set(old_to_new)
    new_names = set(old_to_new.values())
    chained = old_names & new_names

    if chained:
        details = ", ".join(sorted(repr(name) for name in chained))
        raise ValueError(
            f"@renamed_parameter: chaining renames is not allowed on "
            f"{fn.__qualname__}. The following name(s) are used as both old "
            f"and new parameters: {details}. "
            f"Aliases must point directly to the current newest canonical parameter."
        )
```

C-Achard, line 189:
```suggestion

def decorator(fn: Callable[P, R]) -> Callable[P, R]:
    sig = inspect.signature(fn)

    if new not in sig.parameters:
        raise ValueError(
            f"@renamed_parameter: {new!r} is not a parameter of "
            f"{fn.__qualname__}. Available parameters: {list(sig.parameters)}"
        )

    if old in sig.parameters:
        raise ValueError(
            f"@renamed_parameter: {old!r} must not appear in the signature of "
            f"{fn.__qualname__}; use only the new canonical parameter {new!r}."
        )

    existing = getattr(fn, "__deprecated_params__", ())

    info = DeprecationInfo(
        kind="parameter",
        target=fn.__qualname__,
        since=since,
        old_parameter=old,
        new_parameter=new,
    )

    infos = (*existing, info)
    _validate_renamed_parameters(fn=fn, infos=infos)

message = info.format_message()
```

deeplabcut/utils/deprecation.py
```diff
@@ -128,9 +129,43 @@ def renamed_parameter(
...
+
+    Rules:
+        - ``new`` must be the name used in the function signature and all
+          internal call-sites.  ``old`` must **not** appear in the signature.
+        - Do **not** chain renames.  If ``A`` was renamed to ``B`` and ``B``
+          is later renamed to ``C``, replace the ``A→B`` decorator with
+          ``A→C`` directly rather than stacking a second decorator.
+        - Multiple independent renames on the same function (e.g.
+          ``batchsize→batch_size`` *and* ``videotype→video_type``) are fine
+          as long as they do not form a chain.
```
C-Achard, line 146:
```suggestion
    - Do not chain renames. If ``A`` was renamed to ``B`` and ``B`` is later
      renamed to ``C``, update the old alias to point directly to ``C``:
      use ``A→C`` rather than ``A→B→C``.
    - Multiple deprecated aliases may point to the same canonical parameter,
      and multiple independent renames on the same function are allowed, as
      long as no alias is also used as a canonical replacement name.
```

C-Achard (commented):
@deruyter92 Let me know if you think this is okay!

deeplabcut/utils/deprecation.py
deruyter92: just pushed my changes in parallel, otherwise good suggestion. Note that old == new is also caught by current implementation as old *must not* be in signature and new *must* be in signature.

deruyter92:
Note: let's **not** merge before we agree on the parameter naming.  @MMathisLab, both of us understand your concerns; we'll leave the decision up to you. In case not going for `extensions`, would you prefer to stick with `video_type`? Or do you prefer `videotype`? Both are currently used throughout the codebase. 

 see discussion above, quoted here: 

@MMathisLab:
>I'm in favor of leaving all naming as it was originally for full and easy backwards compatibility. Naming changes are a big pain for users that have automated pipelines

@C-Achard:
> ... videotype sounds like we reliably know the underlying media format, but extensions makes it clear this is filename suffix filtering only, with no other guarantees.

@deruyter92:
> My personal preference would go for `extensions` ...  the benefits outweigh the downsides in this case:
> - the old parameter name is still accepted. (we can even suppress the warning that is currently emitted).
> - `video_type` / `videotype`are inconsistently used throughout the codebase, so changing is needed anyway
> - I believe `video_type` to be slightly misleading:
singular despite accepting a list of extensions suggests that the video type will be checked, which is not the case 




**EDIT 2026-05-20 Current status**: 
- `video_type` ,  `videotype` , `extensions` → **`video_extensions`**
- `Shuffles`, `shuffles` → `shuffles`
- `Snapindex`, `snapshotindex` → `snapshotindex`
- `batchsize`, `batch_size` → `batch_size`
- `displayiters`, `display_iters` → `display_iters`
- `saveiters`, `save_iters` → `save_iters`
- `maxiters`, `max_iters` → `max_iters`
- `comparisonbodyparts`, `comparison_bodyparts` → `comparison_bodyparts`

Current guards for renamed parameters (i.e. these cases would explicitly fail in CI):
- chaining renames is disallowed A->B->C (but A->C + B->C is allowed)
- new name must be in function signature
- old name must not be in function signature

deeplabcut/pose_estimation_pytorch/apis/tracking_dataset.py
deruyter92: Nice way to think about it! 


> This would be clean and okay to do though, right ?
>@renamed_parameter(old="oldestname", new="newest", since="3.0.0")
>@renamed_parameter(old="older_name", new="newest", since="4.0.0")
def f(*, newest): ..

Yes this is ok. I've added a test to explicitly confirm.

C-Achard:
@deruyter92 `video_extensions` is maybe another viable alternative

deruyter92:
Rebased again, and changed `video_type` -> `video_extensions` now. 

@C-Achard, could you support again with a thorough review? 

Note that in a few isolated places (such as the 3d pose triangulation modules) the parameter name `videotype` is kept unchanged, because they implement their own file searching logic that doesn't accept a sequence of strings. These locations are flagged for potential deduplication later, but that will require more than a simple rename.

deeplabcut/utils/deprecation.py
```diff
@@ -128,9 +129,51 @@ def renamed_parameter(
         old: The old parameter name that callers may still pass.
         new: The current parameter name the function actually accepts.
         since: Version when the rename happened.
+
+    Rules:
+        - ``new`` must be the name used in the function signature and all
+          internal call-sites.  ``old`` must **not** appear in the signature.
+        - Do **not** chain renames.  If ``A`` was renamed to ``B`` and ``B``
+          is later renamed to ``C``, replace the ``A→B`` decorator with
+          ``A→C`` directly rather than stacking a second decorator.
```
C-Achard, line 138:
```suggestion
          ``A→C`` directly rather than stacking a second decorator.
            Example:
                @renamed_parameter(old="A", new="C", since="12.4.0")
                @renamed_parameter(old="B", new="C", since="13.0.0")
                def func(*, C: int):
                    print(f"C={C}")
```

C-Achard (commented):
Thanks, went through this again!

I also like the flagging of other places this could be harmonized, in this vein I noticed:
- Plotting3D has unflagged duplication
- Same for auxiliaryfunctions_3d
- triangulation.py does miss the intended note

Not sure if they are truly exact duplicates, just noticed similar patterns.
But I don't think this is necessary to address now, a note could be nice if not too time consuming.

Docs also still show a lot of the previous arg name, do we want to update this in one of the docs audit PRs ?
Same for testscripts and notebooks.
Since this is backwards compatible it's no issue, but we should decide whether and how we want to systematically update all of these, also in the future.

deeplabcut/utils/deprecation.py
C-Achard: (I think the example makes it slightly easier to understand, lmk if you agree)

deruyter92: thanks, good suggestion! Also addressed the missing flags for future updating.

MMathisLab (approved):
thanks both! looks good to me now!

PR #3333, opened 2026-05-18
Add additional drop_likelihood_columns guards

### LINKED ISSUE #3319: Shuffle creation failure

C-Achard:
Rename the private _drop_likelihood_columns to drop_likelihood_columns and use it across the codebase to consistently sanitize annotation DataFrames (drop any coord level named 'likelihood'). 

Updated in :

- modelzoo (ma_dlc, ma_dlc_dataframe)
- **pose_estimation_pytorch (dlcloader)**
- utils (skeleton). 

Added a unit test (tests/pose_estimation_pytorch/data/test_dlc_dataloader.py) to ensure DLCLoader.to_coco ignores likelihood columns, as this is arguably the most important one right now.

### More context

As reported in #3319, the current guards allow for shuffle creation, but the DLCLoader path later fails.
This tries to more systematically prevent the problem throughout the codebase, with the by-product of tagging places that load project annotation data on their own without a centralized, validated I/O path (which should be introduced whenever possible)


deruyter92 (approved):
Good solution for now. Indeed we should soon go for a centralized I/O, where we can also normalize data if needed.

C-Achard:
Having this fix ASAP would be great @AlexEMG @MMathisLab; thanks!

PR #3335, opened 2026-05-19
Docs audit [April 2026]: update multi-animal guide

deruyter92:
in line with the main single animal guide, this PR updates the formatting of the multi-animal guide


deruyter92:
@C-Achard, maybe it is ready for a first round of review.

**Changes:** 
After first editing the single and multi-animal guides separately, I decided (after our discussion) that it would be worthwile to try to merge them. Thanks to your custom CSS dropdowns and admonitions, I think the user can easily skip and hide the information that is not relevant. 

In summary the docs now look like this: 
- 1 merged 'DeepLabCut user guide` with a shared workflow.
- The workflow is separated in phases (similar to the original maDLC docs), which are subdivided in steps.
- All multi-animal or single-animal-specific information is put inside a custom dropdown, to easily hide irrelevant info. (@C-Achard thanks for the custom CSS!)
- A special case is made for the multi-animal tracking: 
    - Since the default `video_analysis` step uses `auto_track=True` in version 3.0 onward, the standard user does not have to worry about tracking by default. This is now made very explicit in the documentation.
    - More information on the tracking steps and how to run them manually is contained in a separate doc page: multi-animal tracking. 
    


I think the multi-animal tracking docs still need quite some (formatting and correction) work, but the merged user guide is getting there.

Since all multi-animal-specific info is tagged in dedicated admonitions/dropdowns, we can easily go back to the old situation, in case we prefer to have two separate (largely duplicated) doc pages.

C-Achard:
@deruyter92 I had a few issues in the user_guide file, the frontmatter metadata was a bit malformed, and the dropdowns did not render correctly. Is it okay if I push a fixed version?
(My CSS example was also not fully correct, which is now fixed)

C-Achard:
Very important question: what is the font in the workflow chart actually?
image

docs/main-workflows/user-guide.md
```diff
@@ -0,0 +1,1666 @@
...
+##### Manual frame selection
+
+However, picking frames is highly dependent on the data and the behavior being studied. Therefore, it is hard to
+provide one-size-fits-all code that extracts frames to create a good training dataset for every behavior and animal.
+If the user feels specific frames are lacking, they can extract hand selected frames of interest using the interactive GUI
+provided along with the toolbox. This can be launched by using:
+
+```python
+deeplabcut.extract_frames(config_path, "manual")
+```
```
C-Achard, line 476: Note: This now uses napari and the docs are not up to date

docs/main-workflows/user-guide.md
```diff
@@ -0,0 +1,1666 @@
...
+**Data augmentation:** At this stage you can also decide what type of augmentation to
+use. Once you've called `create_training_dataset`, you can edit the
+[**pytorch_config.yaml**](dlc3-pytorch-config) file that was created (or for the
+TensorFlow engine, the [**pose_cfg.yaml**](https://github.com/DeepLabCut/DeepLabCut/blob/main/deeplabcut/pose_cfg.yaml) file).
+
+- PyTorch Engine: [Albumentations](https://albumentations.ai/docs/) is used for data
+  augmentation. Look at the [**pytorch_config.yaml**](dlc3-pytorch-config) for more
+  information about image augmentation options.
+- TensorFlow Engine: The default augmentation works well for most tasks (as shown on
+  www.deeplabcut.org), but there are many options, more data augmentation, intermediate
```
C-Achard: Is the raw URL really correct ?

deruyter92:
@C-Achard thanks for the review!

> Very important question: what is the font in the workflow chart actually? 
I used Courier 10 Pitch, which an online font search tool found most similar to the one in the logo

C-Achard:
@deruyter92 This is suffering from the outdated squashed commits mismatch as well, happy to fix it if needed!

C-Achard:
Cutoff should be ba8e9858504bb0d7df9d7ce50c7d77e16460dd2d^

deruyter92:
I'm not working on it a the moment, so feel free to rebase as you wish! Thanks for taking care of it!

docs/main-workflows/user-guide.md
deruyter92: Good point, we should add a note or change it probably

docs/main-workflows/user-guide.md
deruyter92: maybe we should remove the reference alltogether

docs/main-workflows/user-guide.md
C-Achard: Done!

PR #3345, opened 2026-05-22
Update deeplabcut-docker - version 0.1.0

deruyter92:
This PR aligns the downstream package `deeplabcut-docker` with the new DeepLabCut 3.0 release and fixes some final issues before release of the new package (see version bump PR #3351).  Note that these are minor additions to a larger previous update in PR #3291. 

*changes*:
- [x] Bugfix in dockerfile for jupyter image (NOTEBOOK_TOKEN was not evaluated)
- [x] Update dockerfile default deeplabcut version: 3.0.0
- [x] Allow local image name / tag
- [x] better error logging when providing wrong image name
- [x] warning for empty notebook token
- [x] version bump + changelog **version 0.1.0**

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -3,25 +3,27 @@
 ARG PYTORCH_VERSION=2.5.1
 ARG CUDA_VERSION=12.4
 ARG CUDNN_VERSION=9
-ARG DEEPLABCUT_VERSION=3.0.0rc14
+ARG DEEPLABCUT_VERSION=3.0.0
 
 # ── core ──────────────────────────────────────────────────────────────────────
 FROM pytorch/pytorch:${PYTORCH_VERSION}-cuda${CUDA_VERSION}-cudnn${CUDNN_VERSION}-runtime AS core
 
 ARG DEEPLABCUT_VERSION
+ARG CUDA_VERSION
+ARG PYTORCH_VERSION
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update -yy && \
     apt-get install -yy --no-install-recommends \
       libgl1 \
       libglib2.0-0 \
+      ffmpeg \
       build-essential \
       git \
-      ffmpeg \
+    && pip install --upgrade pip \
+    && pip install "deeplabcut[modelzoo,wandb]==${DEEPLABCUT_VERSION}" \
+    && apt-get purge -yy --auto-remove build-essential git \
     && rm -rf /var/lib/apt/lists/*
-
-RUN pip install --upgrade pip && \
-    pip install "deeplabcut[modelzoo,wandb]==${DEEPLABCUT_VERSION}"
 
 # Create pretrained weights directory and make it writable
 # (same default as HuggingFaceWeightsMixin in backbones/base.py: Path(__file__).parent / "pretrained_weights")
@@ -44,14 +46,15 @@ WORKDIR /app
 # (runs as root intentionally; deeplabcut-docker adds a host-matched user at runtime)
 FROM core AS jupyter
 
+# Default Jupyter token (override with -e NOTEBOOK_TOKEN=<secret>).
+# An empty value (-e NOTEBOOK_TOKEN=) disables token auth — see docker/README.md.
 ENV NOTEBOOK_TOKEN=deeplabcut
 
 RUN pip install "notebook<7"
 EXPOSE 8888
 HEALTHCHECK --interval=30s --timeout=5s --start-period=20s --retries=3 \
     CMD python -c "import urllib.request; urllib.request.urlopen('http://127.0.0.1:8888/api/', timeout=3).read()" || exit 1
-ENTRYPOINT ["jupyter", "notebook", \
-    "--no-browser", "--NotebookApp.token=${NOTEBOOK_TOKEN}", "--ip", "0.0.0.0"]
+ENTRYPOINT exec jupyter notebook --no-browser --NotebookApp.token="${NOTEBOOK_TOKEN}" --ip 0.0.0.0
 
 # ── test (CI only, not published to Hub) ─────────────────────────────────────
 FROM core AS test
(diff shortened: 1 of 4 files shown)


docker/package/deeplabcut_docker.py
```diff
@@ -149,6 +159,23 @@ def _parse_args() -> tuple[argparse.Namespace, list[str]]:
...
+    r = subprocess.run(_docker() + ["image", "inspect", remote], capture_output=True)
+    if r.returncode != 0:
+        try:
+            subprocess.run(_docker() + ["pull", remote], check=True)
+        except subprocess.CalledProcessError:
+            _log(
+                f"Failed to pull image {remote!r}. Please verify that you specified "
+                "a valid image name and tag, e.g. deeplabcut/deeplabcut:latest."
+            )
+            sys.exit(1)
```
C-Achard: To get better errors when docker is not available we could use:
```suggestion
    r = subprocess.run(
        _docker() + ["image", "inspect", remote],
        capture_output=True,
        text=True,
    )
    
    if r.returncode != 0:
        try:
            subprocess.run(_docker() + ["pull", remote], check=True)
        except subprocess.CalledProcessError as e:
            _log(f"Failed to pull image {remote!r}.")
            if r.stderr:
                _log(f"`docker image inspect` stderr: {r.stderr.strip()}")
            _log("Verify the image name/tag and that Docker is running and accessible.")
            sys.exit(e.returncode)
  ```

File: docker/Dockerfile [hunk 2]
```diff
@@ -50,8 +52,7 @@ RUN pip install "notebook<7"
 EXPOSE 8888
 HEALTHCHECK --interval=30s --timeout=5s --start-period=20s --retries=3 \
     CMD python -c "import urllib.request; urllib.request.urlopen('http://127.0.0.1:8888/api/', timeout=3).read()" || exit 1
-ENTRYPOINT ["jupyter", "notebook", \
-    "--no-browser", "--NotebookApp.token=${NOTEBOOK_TOKEN}", "--ip", "0.0.0.0"]
+ENTRYPOINT jupyter notebook --no-browser --NotebookApp.token="${NOTEBOOK_TOKEN}" --ip 0.0.0.0
```
C-Achard: While there is a default in the python file, I think that not having a default here could disable token auth.
Should we add a default above ?
Also adding exec and a default in case the token is empty (or should we allow users to not set the token to disable ?)

```suggestion
ENV NOTEBOOK_TOKEN=deeplabcut
ENTRYPOINT exec jupyter notebook --no-browser --NotebookApp.token="${NOTEBOOK_TOKEN:-deeplabcut}" --ip 0.0.0.0
```

docker/package/deeplabcut_docker.py
```diff
@@ -149,6 +159,23 @@ def _parse_args() -> tuple[argparse.Namespace, list[str]]:
     return parser.parse_known_args()
 
 
+def _pull_image_if_not_exists(remote: str) -> None:
```
C-Achard: Should we add an "always pull latest" option?

docker/package/deeplabcut_docker.py
```diff
@@ -116,6 +116,16 @@ def _build_user_image(remote: str, local: str) -> None:
     _log("Build succeeded")
 
 
+def _supplementary_group_args() -> list[str]:
+    """Return --group-add flags for each supplementary group of the current user."""
```
C-Achard, line 120: Can this get too long or face parsing issues for weird group names ?

C-Achard (commented):
Thanks for the changes! 
Depending on usefulness feel free to add some of the suggested changes below. The default token is probably the most important point there.

docker/Dockerfile
deruyter92: The default `ENV NOTEBOOK_TOKEN=deeplabcut` is already in place, which means that a user would only disable token auth if explicitly setting `NOTEBOOK_TOKEN` to an empty string, which is a valid choice. I've now documented this better.

Also I've added the `exec` statement as you suggested which is a good improvement.

docker/package/deeplabcut_docker.py
deruyter92: It uses GIDs, not group names so should be safe in this regard. os.getgroups() returns a list of integer group IDs, and each one is passed as str(gid) — so no group name parsing, no string quoting issues, and no risk of special characters.

docker/package/deeplabcut_docker.py
deruyter92: Good suggestion. Rather than making it an option, I would say this should be default behaviour when not a custom image or DLC version is specified.

I made this the new default behavior now.

docker/package/deeplabcut_docker.py
deruyter92: Thanks, incorporated your suggestions in the new `pull_image` function. Good improvements.

docker/Dockerfile
deruyter92: Do you agree that it is a valid option to disable notebook authentication with an empty string:  `NOTEBOOK_TOKEN=`? See also changes in the readme: https://github.com/DeepLabCut/DeepLabCut/pull/3345/changes/fd3153e5593bc0622ff4257b5629d7ab225b8380

C-Achard: Yes, we did warn explicitly in docs... Maybe a warning here too ?

C-Achard (approved):
I would suggest adding warnings for empty tokens/disabling auth given the risks documented, otherwise LGTM, thanks!

deruyter92:
All comments are addressed. Cleaned up the history so it can be merged **without squashing** (this preserves the isolated version bump commit, which is preferable).

deruyter92:
@stes, as the original author of this package we would love hear your feedback, if you have any! Let us know if you would be interested in reviewing this PR. Thanks in advance!

MMathisLab (approved):

PR #3354, opened 2026-06-01
[dev] C7 - Configs refactor: finalize and simplify infrastructure 

deruyter92:
This PR is part of the WIP (see #3198) for migrating from dictionary configs to typed & validated configurations (overview is tracked in https://github.com/DeepLabCut/DeepLabCut/issues/3193).

This PR finalizes, cleans and simplifies the infrastructure for structured configs before moving migrating to the structured config (schema V1).

### DLCBaseConfig, DLCVersionedConfig
The previous versions of the structured configs used OmegaConf (because of dict-like behaviour and easy merging capabilities), or Pydantic dataclasses (lightweight validation, easy to tailor). However the result was still a lot of back-and-forth switching between types and potentially difficult to understand collection of mixins. 

This PR settles for a simpler customized solution: a strict Pydantic BaseModel, subclassed by two DLC-custom classes. 
- **DLCBaseConfig**: for all structured configs in DeepLabCut. It has dict-like access functionality, aliasing for backward compatibility, simple construction and serialization methods for moving between yaml, dict, and canonical typed configs if necessary. 
- **DLCVersionedConfig**: for all top-level configurations that need versioning support. It includes basic change-tracking functionality to see if the state is the same as on disk, and migration between versions. The migration is now class-aware (migrating PoseConfigs, uses a different function than migrating ProjectConfigs)

All legacy references to OmegaConf are removed

### Hook-up in main API

**Normalization:** the function signatures for all main API are changed to accept `str | Path | dict | ProjectConfig` (or similar) which are all normalized to the typed  `ProjectConfig` at the beginning. This means that all downstream code now uses the validated schema. 

**Construction:** The construction of configs used to be a bit of a spaghetti of loading yamls, adjusting fields midway, and then updating them once more in the API calls, such as train_network.  The structured configs refactor improved this a bit, but there was a mixed state with validation halfway construction, requiring the need for nullable fields and moving a lot back and forth between dicts and types. see discussion on this topic in #3369 (merged/included in the current PR)

This PR settles on a two-step approach: 
- build the defaults dictionary from the yaml (given a net_type, derived from ProjectConfig + build args)
- construct a validated PoseConfig from the defaults

Note: API calls can still update the PoseConfig with overrides (e.g. train_network), but these overrides are never required: the validated PoseConfig is always considered a fully initialized model that can be used downstream.

### Testing
- Some functional tests were failing as a result of surfacing validation problems (e.g. dataset was created with a partially incomplete config, which happened to succeed by coincidence). The tests are adapted now to reflect correct (now validated) usage: 
   - always need to specify net_type for creating datasets, or the default_net_type will be used
   - specify ctd_conditions if creating a dataset for BUCTD model
- Some additional tests were added for the migration: testing the consistency between old and new configs. And tests that aim to really break the new structured configs (e.g. the DLCBaseConfig)

more tests will be added if needed [wip]


deruyter92:
@C-Achard, would be very helpful if you could give an intermediate review on the deeplabcut.core.config module and corresponding **tests**. (diff with main, there are some intermediate states that are irrelevant). Thanks a lot!!

deeplabcut/core/config/base_config.py
```diff
@@ -0,0 +1,426 @@
...
+        if not normalize:
+            return self.model_dump()
+        return normalize_for_serialization(self.model_dump())
+
+    def print(
+        self,
+        indent: int = 0,
+        print_fn: Callable[[str], None] | None = None,
+    ) -> None:
+        pretty_print(config=self.to_dict(), indent=indent, print_fn=print_fn)
```
C-Achard, line 149: Would a `__repr__` be useful ?

deeplabcut/core/weight_init.py
```diff
@@ -75,11 +76,12 @@ def __post_init__(self):
...
+            if len(self.conversion_array) != len(self.bodyparts):
                 raise ValueError(
-                    "There must be the same number of elements in the bodyparts list "
-                    "and conv. array; found {self.bodyparts}, {self.conversion_array}"
+                    f"There must be the same number of elements in the bodyparts list "
+                    f"and conv. array; found {self.bodyparts}, {self.conversion_array}"
                 )
+        return self
 
     def to_dict(self) -> dict:
```
C-Achard: This needs the `normalize` keyword, right? Since `super().to_yaml` calls `dict_data = self.to_dict(normalize=True)`

deeplabcut/pose_estimation_pytorch/config/data.py
```diff
@@ -35,8 +33,7 @@ class DataLoaderConfig(ConfigMixin):
     type: str
```
C-Achard: Is this needed? I don't think this is used anywhere now

deeplabcut/core/config/project_config.py
C-Achard, line 1: `_post_yaml_load_updates` calls the parent class function, but the parent just passes. Is this intended?

deeplabcut/core/config/utils.py
```diff
@@ -96,6 +90,62 @@ def write_config(config_path: str | Path, config: dict, overwrite: bool = True)
...
+                f"Config key '{k}' is deprecated, use '{canonical}' instead.",
+                DLCDeprecationWarning,
+                stacklevel=stacklevel,
+            )
+            k = canonical
+        resolved[k] = v
+    return resolved
+
+
+def normalize_for_serialization(obj: Any) -> Any:
```
C-Achard, line 165: Does this need to handle list/tuples/set recursively?

deeplabcut/core/config/project_config.py
```diff
@@ -165,23 +166,25 @@ class ProjectConfig(ChangeTrackingMixin, MigrationMixin, ConfigMixin):
...
     # @TODO @deruyter92 2026-02-13: These aliases are used in parallel. One of them should be removed.
-    with_identity: bool | None = field(
-        default=False, metadata={"comment": "Alias for 'identity'. Kept for backwards compatibility."}
+    with_identity: bool | None = Field(
+        default=False,
+        json_schema_extra={"comment": "Alias for 'identity'. Kept for backwards compatibility."},
     )
-    unique_bodyparts: list[str] = field(
-        default_factory=list, metadata={"comment": "Alias for 'uniquebodyparts'. Kept for backwards compatibility."}
+    unique_bodyparts: list[str] = Field(
```
C-Achard: This field appears twice here and on line 106, I guess we want the alias-defining one?

C-Achard: Is there a way to prevent duplicates in definitions?

deeplabcut/core/config/base_config.py
```diff
@@ -0,0 +1,426 @@
...
+                    object.__setattr__(self, name, value)
+                    return
+                canonical = self._resolve_alias(name, warn=False)
+                field_names = list(type(self).model_fields.keys())
+                dirty_fields = getattr(self, "_dirty_fields", None)
+                if dirty_fields is not None and canonical in field_names:
+                    old = getattr(self, canonical)
+                    original_setattr(self, name, value)
+                    if old != value:
+                        dirty_fields.add(canonical)
```
C-Achard: If setattr validates, could this mark dirty because validated != old ?

```suggestion
                    original_setattr(self, name, value)
                    new = getattr(self, canonical)
                    if old != new:
                        dirty_fields.add(canonical)
```

deeplabcut/core/config/base_config.py
```diff
@@ -0,0 +1,426 @@
...
+    # ------------------------------------------------------------------
+    # Change tracking
+    # ------------------------------------------------------------------
+
+    def model_post_init(self, __context: Any) -> None:
+        super().model_post_init(__context)
+        self._init_change_tracking()
+
+    def _init_change_tracking(self) -> None:
+        if getattr(self, "_change_tracking_initialized", False):
```
C-Achard: Can we somehow put this directly in setattr? Not sure if it's possible, if not some explanation could be nice. 
How much of an impact does this have on testing? Any particular pitfalls to avoid?

deeplabcut/core/config/base_config.py
```diff
@@ -0,0 +1,426 @@
...
+    # ------------------------------------------------------------------
+
+    def __setattr__(self, name: str, value: Any) -> None:
+        name = self._resolve_alias(name, warn=True, stacklevel=3)
+        super().__setattr__(name, value)
+
+    def __getattr__(self, name: str) -> Any:
+        if name not in type(self)._alias_map():
+            raise AttributeError(f"'{type(self).__name__}' has no attribute '{name}'")
+        return getattr(self, self._resolve_alias(name, warn=True, stacklevel=2))
```
C-Achard: Would this be safer?
```suggestion
        def __getattr__(self, name: str) -> Any:
            if name in type(self)._alias_map():
                return getattr(self, self._resolve_alias(name, warn=True, stacklevel=2))
            return super().__getattr__(name)
```

deeplabcut/core/config/versioning.py
```diff
@@ -94,21 +91,31 @@ def _log_field_changes(before: dict, after: dict, from_v: int, to_v: int) -> Non
         logger.debug("  [%s] No field changes", step)
 
 
-def register_migration(from_version: int, to_version: int):
-    """Decorator to register a migration function.
+def register_migration(
+    from_version: int,
+    to_version: int,
+    config_type: str,
```
C-Achard, line 97: Should this be more strongly typed, or should we have some way of listing config types in the project? This is fine as is, just wondering what would be useful. Might be overengineering.

deeplabcut/core/config/project_config.py
```diff
@@ -72,52 +67,58 @@ class ProjectConfig(ChangeTrackingMixin, MigrationMixin, ConfigMixin):
...
-    engine: str = field(
+    engine: str = Field(
         default="pytorch",
-        metadata={"comment": "\nDefault DeepLabCut engine to use for shuffle creation (either pytorch or tensorflow)"},
+        json_schema_extra={
+            "comment": "\nDefault DeepLabCut engine to use for shuffle creation (either pytorch or tensorflow)"
+        },
     )
 
     # Annotation data set configuration (and individual video cropping parameters)
```
C-Achard:
```suggestion
    # Annotation dataset configuration, individual video cropping parameters
```

deeplabcut/core/config/project_config.py
```diff
@@ -72,52 +67,58 @@ class ProjectConfig(ChangeTrackingMixin, MigrationMixin, ConfigMixin):
...
     bodyparts: list[str] | str = "MULTI!"
 
     # TODO @deruyter92 2026-02-06: The current pipeline requires at least one individual defined in the
     # default configuration. This will be removed in the future.
-    individuals: list[str] = field(default_factory=lambda: ["individual_1"])
-    uniquebodyparts: list[str] = field(default_factory=list)  # multi-animal project key
-    multianimalbodyparts: list[str] = field(default_factory=list)  # multi-animal project key
-    unique_bodyparts: list[str] = field(default_factory=list)  # metadata key; same as uniquebodyparts
+    individuals: list[str] = Field(default_factory=lambda: ["individual_1"])
+    uniquebodyparts: list[str] = Field(default_factory=list)  # multi-animal project key
```
C-Achard: This appears several times, I assume the one with aliases defs should stay?

deeplabcut/core/config/project_config.py
```diff
@@ -126,9 +127,9 @@ class ProjectConfig(ChangeTrackingMixin, MigrationMixin, ConfigMixin):
     colormap: str = "rainbow"
```
C-Achard, line 135: What would be the plan for validation/documenting valid values? This will come at a later time, right? (Apologies if we discussed that point already)

deeplabcut/core/config/project_config.py
```diff
@@ -165,23 +166,25 @@ class ProjectConfig(ChangeTrackingMixin, MigrationMixin, ConfigMixin):
...
+        json_schema_extra={"comment": "Alias for 'identity'. Kept for backwards compatibility."},
     )
-    unique_bodyparts: list[str] = field(
-        default_factory=list, metadata={"comment": "Alias for 'uniquebodyparts'. Kept for backwards compatibility."}
+    unique_bodyparts: list[str] = Field(
+        default_factory=list,
+        json_schema_extra={"comment": "Alias for 'uniquebodyparts'. Kept for backwards compatibility."},
     )
 
     # TODO @deruyter92 2026-02-06: These parameters are no longer used in the new pipeline.
```
C-Achard, line 182: Should we comment them out then? If it's really dead code

deeplabcut/core/config/project_config.py
C-Achard: Is it some sort of future proofing? What would be potential use cases if yes?

C-Achard:
```python
        if project_path.resolve() != self.project_path.resolve():
```
resolve() or absolute() here?

deeplabcut/core/weight_init.py
C-Achard, line 1: Most model config is in the pytorch backend, but this is "core" config. Is this desirable, or would it be good to move?

C-Achard (commented):
@deruyter92 Here are some first comments, mostly focusing on usability/possible refactor artifacts.

Note that I only reviewed the diff of C7 to the feature branch; happy to review the full version once this is finalized.

deeplabcut/core/config/base_config.py
deruyter92: Good suggestion. Updated accordingly!

deeplabcut/core/config/base_config.py
deruyter92: Pydantic already gives a useful short `__repr__`. The full `print()` method is different on purpose: full nested, human-readable output. I think both are fine as they are now.

deeplabcut/core/config/base_config.py
deruyter92: newer version not compares coerced old and coerced new values

deeplabcut/core/config/base_config.py
deruyter92: yes good point. Moving from dataclasses to subclassing BaseModel makes this possible.

I have now significantly simplified this

deeplabcut/core/config/project_config.py
deruyter92: Yes, I'd suggest we settle on a definite version in a separate PR. For now I use the old `uniquebodyparts` and keep it like that.

deeplabcut/core/config/project_config.py
deruyter92: I've added some basic validation now!

deeplabcut/core/config/project_config.py
deruyter92: Only via linting. Python class bodies never truly define the same name twice. I've added linting rule "PIE" now.

deeplabcut/core/config/project_config.py
deruyter92: agreed. Did that now

deeplabcut/core/config/project_config.py
deruyter92: should be absolute(), good catch!

deeplabcut/core/config/utils.py
deruyter92: Yes that's better. Done, thanks!

deeplabcut/core/config/versioning.py
deruyter92: I would say it is fine for now, but we can add it in the V1 migration PR if you want.

deeplabcut/core/weight_init.py
deruyter92: We might want to migrate the location. It is here for historical reasons, I just did not move it. I think that would deserve a separate PR though.

deeplabcut/core/weight_init.py
deruyter92: good point. These old serialization + construction methods need to be adjusted altogether.

I kept only the legacy load method and fixed everything else via coercion and validation instead.

deeplabcut/pose_estimation_pytorch/config/data.py
deruyter92: good catch. This parent class was orphaned, but is actually not even needed currently. I removed it.

deruyter92 (commented):
Thanks @C-Achard. Addressed all your comments. Let's discuss next week. 

I have a few extra adjustments in mind (better hook-up in codebase), but after that it's ready for proper stress-testing.

deeplabcut/core/config/base_config.py
```diff
@@ -0,0 +1,413 @@
...
+from collections.abc import Callable, Iterator
+from pathlib import Path
+from typing import Any
+
+from pydantic import BaseModel, ConfigDict, Field, PrivateAttr, model_validator
+from pydantic_core import ArgsKwargs
+from ruamel.yaml.comments import CommentedMap
+from typing_extensions import Self
+
+from deeplabcut.core.config import versioning
```
C-Achard:
```suggestion
import deeplabcut.core.config.versioning as versioning
```

deeplabcut/core/config/base_config.py
```diff
@@ -0,0 +1,413 @@
...
+from deeplabcut.core.config.utils import (
+    normalize_for_serialization,
+    pretty_print,
+    read_config_as_dict,
+    resolve_alias,
+    resolve_aliases_in_dict,
+    write_config,
+)
+from deeplabcut.core.config.versioning import migrate_config
```
C-Achard: Maybe use versioning.migrate_config given the above import, not very important though

deeplabcut/core/config/base_config.py
```diff
@@ -0,0 +1,413 @@
...
+            raise TypeError(
+                "Failure to load configuration: Expected a config instance, "
+                f"dictionary, string, or Path. Got {type(config)}"
+            )
+
+    @classmethod
+    def from_yaml(cls, yaml_path: str | Path, ignore_empty: bool = True) -> Self:
+        yaml_dict = read_config_as_dict(yaml_path)
+        if ignore_empty:
+            yaml_dict = {k: v for k, v in yaml_dict.items() if v is not None}
```
C-Achard, line 107: If yaml_dict is None, this could raise an AttributeError

deeplabcut/core/config/utils.py
C-Achard, line 78:
```suggestion
    return cfg or {}
```

deeplabcut/core/config/base_config.py
```diff
@@ -0,0 +1,413 @@
...
+        elif isinstance(config, dict):
+            return cls.from_dict(config)
+        else:
+            raise TypeError(
+                "Failure to load configuration: Expected a config instance, "
+                f"dictionary, string, or Path. Got {type(config)}"
+            )
+
+    @classmethod
+    def from_yaml(cls, yaml_path: str | Path, ignore_empty: bool = True) -> Self:
```
C-Achard, line 104: Should `ignore_empty` be recursive? Or is that not desirable?

deeplabcut/core/config/project_config.py
```diff
@@ -72,90 +74,95 @@ class ProjectConfig(ChangeTrackingMixin, MigrationMixin, ConfigMixin):
...
-    identity: bool | None = None
+    identity: bool | None = Field(default=None, json_schema_extra={"aliases": ["with_identity"]})
 
     # Project path
-    project_path: Path = field(default=Path(), metadata={"comment": "\nProject path (change when moving around)"})
+    project_path: Path = Field(
+        default_factory=Path,
+        json_schema_extra={"comment": "\nProject path (change when moving around)"},
+    )
     pose_config_path: Path = Path()
```
C-Achard: I think a better default would be None, right? This is cwd(), is there a need for this in particular?
```suggestion
    pose_config_path: Path | None = None
```

deeplabcut/core/config/utils.py
C-Achard, line 403: Do we want/have to keep this operating on a dict?

C-Achard, line 361: Is this validated?

deeplabcut/core/config/validation.py
```diff
@@ -0,0 +1,57 @@
...
+import numpy as np
+from numpy.typing import NDArray
+from pydantic import AfterValidator, BeforeValidator, Field, GetPydanticSchema, InstanceOf
+
+
+def _describe(value: float, name: str | None = None) -> str:
+    return f"{name} ({value})" if name else f"{value}"
+
+
+def greater_than(
```
C-Achard, line 13: The name implies it is strict but it is not here right?

deeplabcut/core/config/project_config.py
```diff
@@ -72,90 +74,95 @@ class ProjectConfig(ChangeTrackingMixin, MigrationMixin, ConfigMixin):
...
         default_factory=list,
-        metadata={"comment": "\nTraining,Evaluation and Analysis configuration"},
+        json_schema_extra={"comment": "\nTraining,Evaluation and Analysis configuration"},
     )
-    iteration: int | None = None
+    iteration: NonNegativeInt | None = None
     default_net_type: str = "resnet_50"
     default_augmenter: str | None = None
     default_track_method: str | None = None
     snapshotindex: str | int = "all"
```
C-Achard:
```suggestion
    snapshotindex: Literal["all"] | int = "all"
```

deeplabcut/core/config/project_config.py
```diff
@@ -72,90 +74,95 @@ class ProjectConfig(ChangeTrackingMixin, MigrationMixin, ConfigMixin):
...
-    alphavalue: float = 0.7
+    pcutoff: Fraction = 0.4
+    dotsize: NonNegativeInt = 12
+    alphavalue: Fraction = 0.7
     colormap: str = "rainbow"
 
     # Training, evaluation and analysis configuration
-    TrainingFraction: list[float] = field(
+    TrainingFraction: list[Fraction] = Field(
         default_factory=list,
```
C-Achard: Should this be empty by default?

deeplabcut/core/config/project_config.py
```diff
@@ -72,90 +74,95 @@ class ProjectConfig(ChangeTrackingMixin, MigrationMixin, ConfigMixin):
...
+    # Annotation dataset configuration (and individual video cropping parameters)
+    video_sets: dict[str, Any] = Field(
         default_factory=dict,
-        metadata={"comment": "\nAnnotation data set configuration (and individual video cropping parameters)"},
+        json_schema_extra={"comment": "\nAnnotation data set configuration (and individual video cropping parameters)"},
     )
     # VV TODO @deruyter92 2026-01-30: following the old original config.yaml template for now. VV
     # VV We should change this to a list[str] in the future. VV
-    bodyparts: list[str] | str = "MULTI!"
+    bodyparts: UniqueStrList | str = "MULTI!"
```
C-Achard:
```suggestion
    bodyparts: UniqueStrList | Literal["MULTI!"]= "MULTI!"
```

deeplabcut/core/config/project_config.py
```diff
@@ -72,90 +74,95 @@ class ProjectConfig(ChangeTrackingMixin, MigrationMixin, ConfigMixin):
...
+        },
     )
-    stop: float = 1.0
-    numframes2pick: int = 20
+    stop: Fraction = 1.0
+    numframes2pick: NonNegativeInt = 20
 
     # Plotting configuration
-    skeleton: list[list[str]] = field(
+    skeleton: list[UniqueStrList] = Field(
```
C-Achard: I think validation should also check that are exactly two bodyparts per connection? Or is this done somewhere else?

deeplabcut/core/config/project_config.py
```diff
@@ -72,90 +74,95 @@ class ProjectConfig(ChangeTrackingMixin, MigrationMixin, ConfigMixin):
...
         default=None,
-        metadata={"comment": "if cropping is true for analysis, then set the values here:"},
+        json_schema_extra={"comment": "if cropping is true for analysis, then set the values here:"},
     )
-    x2: int | None = None
-    y1: int | None = None
-    y2: int | None = None
+    x2: NonNegativeInt | None = None
+    y1: NonNegativeInt | None = None
+    y2: NonNegativeInt | None = None
```
C-Achard, line 162: Are the crop bounds validated?

deeplabcut/core/config/base_config.py
```diff
@@ -0,0 +1,413 @@
...
+    """
+
+    config_version: int = Field(
+        default=versioning.CURRENT_CONFIG_VERSION,
+        json_schema_extra={"comment": "Config schema version. Do not edit manually."},
+    )
+
+    _initialized: bool = PrivateAttr(default=False)
+    _dirty_fields: set[str] = PrivateAttr(default=set())
+    _change_notes: dict[str, Any] = PrivateAttr(default=dict())
```
C-Achard: Dumb question, is this a mutable default and does it have the same caveats as usual if it is?

deeplabcut/core/config/base_config.py
```diff
@@ -0,0 +1,413 @@
...
+        message: str,
+        *,
+        include_caller: bool = False,
+        _stack_depth: int = 1,
+    ) -> None:
+        if include_caller:
+            frame = sys._getframe(_stack_depth)
+            filename = frame.f_code.co_filename.rsplit("/", 1)[-1]
+            message = f"{message} [{filename}:{frame.f_lineno}]"
+        self._change_notes[field_name] = message
```
C-Achard, line 419: To avoid bad pairings, we may want:
```suggestion
  def record_change_note(
      self,
      field_name: str,
      message: str,
      *,
      include_caller: bool = False,
      _stack_depth: int = 1,
  ) -> None:
      field_name = self._resolve_alias(field_name)
  
      if field_name not in type(self).model_fields:
          raise KeyError(f"'{type(self).__name__}' has no field '{field_name}'")
  
      if include_caller:
          frame = sys._getframe(_stack_depth)
          filename = frame.f_code.co_filename.rsplit("/", 1)[-1]
          message = f"{message} [{filename}:{frame.f_lineno}]"
  
      self._change_notes[field_name] = message
```

tests/core/config/test_base_config.py
C-Achard, line 1: After working on #3366, I would suggest adding xfail tests for any desirable but non-yet-implemented feature if relevant. I think it helps make features/required changes clearly visible, adds future tests, and clarifies future expectations. Please skip if it is too time consuming!

deeplabcut/core/config/project_config.py
```diff
@@ -72,90 +74,95 @@ class ProjectConfig(ChangeTrackingMixin, MigrationMixin, ConfigMixin):
...
+    skeleton: list[UniqueStrList] = Field(
         default_factory=list,
-        metadata={"comment": "\nPlotting configuration"},
+        json_schema_extra={"comment": "\nPlotting configuration"},
     )
     skeleton_color: str = "black"
-    pcutoff: float = 0.4
-    dotsize: int = 12
-    alphavalue: float = 0.7
+    pcutoff: Fraction = 0.4
```
C-Achard: It seems the previous default for a new project was 0.6. This is kind of arbitrary in general, but curious to ask if there was a reason to change it here ?

deeplabcut/core/config/project_config.py
C-Achard: In general, is there a risk the MULTI sentinel might end up in a single animal config?

C-Achard (commented):
Amazing work so far! Quite a few comment split between here and files/changes in #3366, let me know if you have questions! Tried to focus on correctness, legacy preservation and edge cases so far, hope it helps.

deeplabcut/core/config/base_config.py
deruyter92: It is currently only used so we can strictly define non-nullable types for some top-level fields in the project config that are sometimes kept empty `identity` (bool), `engine`, etc. 

I would prefer on the long term to even remove the ignore_empty, so that empty is strictly interpreted as None, and yamls can just omit the field if they want defaults. 

So right now: preferably not recursive, and better documented + flagged for removal.

deeplabcut/core/config/base_config.py
deruyter92: Yes yaml_dict should always be a dict, which wasn't properly guaranteed by read_config_as_dict. I've added validation there now so it always returns a dict.

deeplabcut/core/config/base_config.py
deruyter92: No good point! PrivateAttr is specially by Pydantic, which is why it worked. Still, the idiomatic way is using default factories. I've fixed that now, thanks!

deeplabcut/core/config/base_config.py
deruyter92: Good suggestion. Is implemented

deeplabcut/core/config/project_config.py
deruyter92: hmm I don't know where I used the 0.4 from (probably found it somewhere else in the codebase). Changed it back to 0.6 now, thanks.

deeplabcut/core/config/project_config.py
deruyter92: good point. Added validation for that now (BodpartPair)

deeplabcut/core/config/project_config.py
deruyter92: No you are right, [0.95] would be a more appropriate default, as it seems the DLC standard everywhere.

deeplabcut/core/config/utils.py
deruyter92: Added validation. Loudly fails for None or non-mapping top-level.

deruyter92: No this is the fallback option after `validated_cfg.to_yaml`, which unfortunately is not validated. Do you think this should be more strict? e.g. loudly fail? I don't want to risk losing information. The written config might be patchable

deruyter92: Ideally not. The favorable alternative is `DLCBaseConfig.update()` now, but we cannot deprecate it until all call sites are adjusted as well (including some TF branches that use configs that fall outside the current schemas). 

I will revisit this after updating the main PoseConfig / TestConfig constructors. Probably most call sites can be fixed. I will also properly document that we are planning to remove this.

deeplabcut/core/config/validation.py
deruyter92: addressed

deeplabcut/core/config/project_config.py
deruyter92: added validation, so the risk is mitigated

deruyter92 (commented):
@C-Achard, thanks for the very helpful review! 

I addressed most comments and cherry-picked from #3366 what was needed. 

Remaining tasks for me: 
- address the remaining xfails and add more stress-tests if needed. 
- decide on arbitrary types (dd5b46950fb1538535255b9dffbba26e057ff2a8), I think allowing them might not be ideal and not needed per se. 
- improve constructors in make_pose_config, modelzoo etc. 
- re-visit core config utils and see what needs to operate on dict still, or can be changed to typed. 
- revisit all todos to see if they are/can be addressed or we will do that in next version.

deeplabcut/core/config/utils.py
C-Achard: I think the fallback is fine, it just feels a bit strange to fully skip the validation on any error, but then it is also hard to be granular without giant branching... Let's see in practice, we can rely on logging for now I think

tests/core/config/test_base_config.py
```diff
@@ -0,0 +1,504 @@
...
+class _ToyVersioned(DLCVersionedConfig):
+    config_version: int = _TOY_VERSION_NEW
+    toy_new_field: str = ""
+
+
+@register_migration(_TOY_VERSION_OLD, _TOY_VERSION_NEW, config_type="_ToyVersioned")
+def _toy_versioned_migrate_v98_to_v99(config: dict) -> dict:
+    if _LEGACY_FIELD in config:
+        config["toy_new_field"] = config.pop(_LEGACY_FIELD)
+    return config
```
C-Achard: Would it be better to make this a fixture and restore the state afterwards? Just to keep things clean

C-Achard: Same for test_versioning.py

tests/core/config/test_core_config.py
```diff
@@ -0,0 +1,457 @@
...
+    config_path = tmp_path / "config.yaml"
+    config_path.write_text("- item\n")
+    with pytest.raises(ValueError, match="must be a YAML mapping"):
+        read_config_as_dict(config_path)
+
+
+def test_read_config_as_dict_breaks_for_yaml_tags(tmp_path):
+    """read_config breaks for YAML tags like !!python/tuple"""
+    config_path = tmp_path / "config.yaml"
+    # NOTE @deruyter92 2026-02-03: This test is currently skipped, because
```
C-Achard: It is not xfailed so I think the comment may be outdated?

tests/core/config/test_core_config.py
```diff
@@ -0,0 +1,457 @@
...
+
+def test_read_config_breaks_for_yaml_tags(tmp_path):
+    """read_config raises for YAML files containing unsafe tags like !!python/tuple."""
+    config_path = tmp_path / "config.yaml"
+    config_path.write_text("project_path: /old/path\nengine: pytorch\nbodyparts: !!python/tuple [a, b, c]\n")
+    with pytest.raises(YAMLError):
+        read_config(config_path)
+
+
+def test_read_config_breaks_for_invalid_fieds(tmp_path):
```
C-Achard:
```suggestion
def test_read_config_breaks_for_invalid_fields(tmp_path):
```

deeplabcut/core/config/project_config.py
```diff
@@ -0,0 +1,254 @@
...
+    croppedtraining: bool | None = None
+
+    @property
+    def bodyparts_list(self) -> list[str]:
+        # Animal-count agnostic; Always return a list (never "MULTI!", None, etc.)
+        if self.multianimalproject:
+            return list(self.multianimalbodyparts)
+        if self.bodyparts == "MULTI!":
+            raise ValueError("bodyparts must be a list of strings when multianimalproject is False, got 'MULTI!'")
+        return list(self.bodyparts)
```
C-Achard, line 199: Should we add a few tests for this, since it is used in many places but not tested directly?
```python
class TestProjectConfigBodypartsList:
    def test_single_animal_bodyparts_list_returns_bodyparts(self):
        cfg = ProjectConfig(multianimalproject=False, bodyparts=["nose", "tail"])
        assert cfg.bodyparts_list == ["nose", "tail"]

    def test_multi_animal_bodyparts_list_returns_multianimal_bodyparts(self):
        cfg = ProjectConfig(
            multianimalproject=True,
            bodyparts="MULTI!",
            multianimalbodyparts=["nose", "tail"],
        )
        assert cfg.bodyparts_list == ["nose", "tail"]

    def test_single_animal_multi_sentinel_is_not_split_into_characters(self):
        cfg = ProjectConfig(multianimalproject=False, bodyparts="MULTI!")
        assert cfg.bodyparts_list == []
```

tests/pose_estimation_pytorch/apis/test_apis_training.py
```diff
@@ -0,0 +1,57 @@
...
+        "project_path": str(tmp_path),
+        "bodyparts": ["nose"],
+        "uniquebodyparts": [],
+        "individuals": ["mouse"],
+    }
+    cfg_path = tmp_path / "pytorch_config.yaml"
+    defaults = make_pytorch_pose_config(project_cfg, str(cfg_path), net_type="resnet_50")
+    if resume_from is not None:
+        defaults["resume_training_from"] = resume_from
+    return PoseConfig.from_dict(defaults)
```
C-Achard: Should this be from_dict since make_pytorch_pose_config now returns a config object?

tests/pose_estimation_pytorch/config/test_pose_config_creation.py
```diff
@@ -0,0 +1,207 @@
...
+@pytest.mark.parametrize("case", POSE_CONFIG_CASES, ids=lambda c: c.id)
+def test_pose_config_creation(case: PoseConfigCase, tmp_path: Path) -> None:
+    cfg = _as_dict(
+        build_pose_config(
+            case.project,
+            tmp_path / "pytorch_config.yaml",
+            net_type=case.net_type,
+            save=True,
+            top_down=case.build_kwargs.pop("top_down", False),
+            **case.build_kwargs,
```
C-Achard: Copying may be good here instead of mutating

tests/pose_estimation_pytorch/config/test_pose_config_creation.py
```diff
@@ -0,0 +1,207 @@
...
+    build_pose_config = PoseConfig.build
+except ImportError:
+    from deeplabcut.pose_estimation_pytorch.config.make_pose_config import make_pytorch_pose_config as build_pose_config
+
+
+FIXTURES_DIR = Path(__file__).resolve().parent / "fixtures"
+
+
+def _load_yaml(path: Path) -> dict:
+    data = yaml.safe_load(path.read_text())
```
C-Achard: Is this required, or should it use the centralized loader?

tests/pose_estimation_pytorch/modelzoo/test_generalized_data_converter_config.py
```diff
@@ -0,0 +1,171 @@
...
+        """Keys not overridden in Ma should match Single defaults."""
+        single_config = SingleDLC_config()
+        ma_config = MaDLC_config()
+        for key in SINGLE_DLC_DEFAULTS:
+            if key in MA_DLC_DEFAULTS_OVERRIDES:
+                continue
+            assert ma_config.cfg[key] == single_config.cfg[key], f"MaDLC_config.cfg[{key!r}] should match Single"
+
+
+class TestModelzooCreateProjectConfig:
```
C-Achard, line 106: Same comment, should this use the central YAML loading now?

tests/pose_estimation_pytorch/config/test_pose_config_creation.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+@pytest.mark.parametrize("case", POSE_CONFIG_CASES, ids=lambda c: c.id)
+def test_pose_config_creation(case: PoseConfigCase, tmp_path: Path) -> None:
+    cfg = _as_dict(
+        build_pose_config(
+            case.project,
+            tmp_path / "pytorch_config.yaml",
+            net_type=case.net_type,
+            save=True,
```
C-Achard: SHould we check if the file exists?

tests/pose_estimation_pytorch/modelzoo/test_generalized_data_converter_config.py
```diff
@@ -0,0 +1,171 @@
...
+        # Actual call requires compat.return_train_network_path to resolve paths;
+        # this test just documents the intended defaults for migration.
+        intended_train_test_defaults = {
+            "multi_stage": True,
+            "batch_size": 8,
+            "gradient_masking": True,
+        }
+        assert intended_train_test_defaults["multi_stage"] is True
+        assert intended_train_test_defaults["batch_size"] == 8
+        assert intended_train_test_defaults["gradient_masking"] is True
```
C-Achard: Should this be a comment rather than a test?

tests/pose_estimation_pytorch/modelzoo/test_generalized_data_converter_config.py
```diff
@@ -0,0 +1,171 @@
...
+#
+"""Tests capturing the current state of the generalized_data_converter configuration.
+
+These tests document default values and behavior so that a migration to a new
+configuration system can preserve existing behavior. Add or adjust assertions
+if you intentionally change defaults.
+"""
+
+import os
+import tempfile
```
C-Achard: SHould we favor the tmp fixture or keep this? (Not very important but consistency can be nice)

deruyter92:
@C-Achard pushed my latest commit for now, happy to take your review again. Let me know if anything is unclear. I have rebased on main for easier testing. Just so you know, you have reviewed everything up to 6812f3f4588018e72e09bb461c5e6a7d165891c5 before the rest is new. But any comments are welcome, also on prior code if you find anything. 

Most relevant commits are probably [2c3388c](https://github.com/DeepLabCut/DeepLabCut/pull/3354/commits/2c3388cb5627f852c187e9ebbd175e2d33dbe604) and  [ee524e5](https://github.com/DeepLabCut/DeepLabCut/pull/3354/commits/ee524e50ccbc7856d711aec0c45a787715429afd).

Thanks again!

deeplabcut/core/config/validation.py
```diff
@@ -0,0 +1,81 @@
...
+        raise ValueError(f"{_describe(value, name)} must be greater than {_describe(threshold, threshold_name)}")
+
+
+def less_than(
+    value: float,
+    threshold: float,
+    name: str,
+    threshold_name: str | None = None,
+) -> None:
+    if value > threshold:
```
C-Achard:
```suggestion
    if value >= threshold:
```

deeplabcut/core/config/base_config.py
```diff
@@ -0,0 +1,433 @@
...
+        return [(name, getattr(self, name)) for name in self._field_names()]
+
+    def select(self, path: str, default: Any = None) -> Any:
+        """Select a value from the config using dot notation for nested keys."""
+        try:
+            return self._navigate_nested(fields=path.split("."))
+        except (AttributeError, KeyError, TypeError):
+            return default
+
+    def set_nested(self, path: str, value: Any) -> None:
```
C-Achard:
```suggestion
    def set_nested(self, path: str, value: Any) -> Self:
```

deeplabcut/core/config/base_config.py
```diff
@@ -0,0 +1,433 @@
...
+        if "." not in path:
+            setattr(self, path, value)
+            return
```
C-Achard: SHould this return self as well?

tests/core/config/test_config_breakage.py
```diff
@@ -0,0 +1,287 @@
...
+    assert "identity" in cfg.dirty_fields
+    assert "with_identity" not in cfg.dirty_fields
+
+
+# -----------------------------------------------------------------------------
+# Cross-field validation vs bulk update
+# -----------------------------------------------------------------------------
+
+
+@pytest.mark.xfail(
```
C-Achard, line 244: I think this is passing now

tests/core/config/test_config_breakage.py
```diff
@@ -0,0 +1,287 @@
...
+# Migration skipped on typed construction
+# -----------------------------------------------------------------------------
+
+
+class _TypedConstructionCfg(DLCVersionedConfig):
+    config_version: int = 1
+    epochs: int = 50
+
+
+@pytest.mark.xfail(
```
C-Achard: I think this is passing now as well

C-Achard (approved):
Here are a few additional comments, nohting major really so feel free to skip some if they feel too tangential. Overall after some digging I could not really find fundamental design issues I think should be addressed, so I think we can roll this out soon!
Let me know if you would like further review as well, but it might mostly be minor details at this point.

tests/core/config/test_config_breakage.py
deruyter92: by coincidence it was, I updated the test. The principle still needs to be addressed: that bulk updates internally use sequential setattr calls, which causes invalid states midway updating. 

It would be better to do cross-field validation only after full initialization.

tests/pose_estimation_pytorch/modelzoo/test_generalized_data_converter_config.py
deruyter92: yeah.... good point this wasn't testing anything. Updated now.

tests/core/config/test_base_config.py
deruyter92: yes good suggestoin. This is implemented now

deeplabcut/core/config/utils.py
deruyter92: yes you are right, this is a bit odd and I don't know a good solution until we know for sure that the current ProjectConfig schema fully fits and works with all code branches, including some potentially more obscure tensorflow-era project config mutations that this PR may have missed. Since I cannot guarantee the last, we may have to stick with this and see in practice indeed. Since the round-trip is likely to fail, I assume that we will get reports on it, unless users adjust their configs manually.

What I did improve is the documentation (e.g. see docstring) and logging for this behavior. The logger emits an error message now instead of a warning, improving issue-reporting. Also added a test that documents the behavior.

tests/pose_estimation_pytorch/config/test_pose_config_creation.py
deruyter92: centralized loader is better indeed. Updated now.

tests/pose_estimation_pytorch/config/test_pose_config_creation.py
deruyter92: good point it was not tested fully. I've added a small separate test dedicated to cover the saving logic, and set save=False here.

deeplabcut/core/config/project_config.py
deruyter92: Agreed, good solution. Copied with one minor adjustment.

deruyter92:
@C-Achard, thanks for the review, very helpful! I've addressed all your comments, a few might things might be worth noting the rest is minor:

1. I moved the deprecation module to `deeplabcut/core`. do you agree? (helped to avoid circular imports),
[27af05d](https://github.com/DeepLabCut/DeepLabCut/pull/3354/commits/27af05d70f5a871c61375b29f75faf3a9468ef13)
2. I've added a convenience function to repair the project path for `PoseConfig.from_any()`
[48ac2b7](https://github.com/DeepLabCut/DeepLabCut/pull/3354/commits/48ac2b7c28aa08f3d53e3e5094e486054357d370)
3. I did not change the `edit_config` function for dicts in the end, but marked it as legacy, and that it is better to edit the typed configs directly. I have adjusted most callers accordingly: they edit and save the typed configs directly. 
4. I've hooked up the main functions in trainsetmanipulation.py as well, like other main API (the lower level functions still use cfg paths, but I think it's better not to recurse all the way down in the adjustments). 
[b0c3433](https://github.com/DeepLabCut/DeepLabCut/pull/3354/commits/b0c343395af385c73f3217af85f3eddf7e168276)

C-Achard:
@deruyter92 
1. Yes no problem!
2. Very nice
3. Sounds good, I think we should move away from this eventually (as well as the dict access as discussed)
4. That works, probably we can also hook up lower level API in subsequent PRs to help review and mitigate any risks. So yes I agree with not going lower rn

Thanks again!

PR #3358, opened 2026-06-05
Refactor VideoProcessor class

### LINKED ISSUE #3357: 'VideoProcessorCV' object has no attribute 'FPS'

### Operating System

Windows 11 + miniconda

### DeepLabCut version

3.0.0

pytorch

### DeepLabCut mode

single animal

### Device type

NVIDIA GeForce RTX 5070 Ti, Driver Version: 591.86, CUDA Version: 13.1

### Bug description 🐛

Can't create videos after training, error showed like this: 
AttributeError: 'VideoProcessorCV' object has no attribute 'FPS'.

### Steps To Reproduce

Run like this:
1. Created a new project [Copy videos to project folder]
2. Extracted frames
3. Labeled frames
4. Created training dataset
5. Trained metwork
6. Evaluate network
7. Analyzed videos
8. Create videos

### Relevant log output

```shell
(dlc) C:\Windows\System32>python -m deeplabcut
Starting GUI...
INFO:deeplabcut.gui:Use all bodyparts
INFO:deeplabcut.gui:All video files have been cleared.
INFO:deeplabcut.gui:Cleared selected videos
INFO:deeplabcut.gui:All video files have been cleared.
INFO:deeplabcut.gui:Cleared selected videos
INFO:deeplabcut.gui:All video files have been cleared.
INFO:deeplabcut.gui:Cleared selected videos
INFO:deeplabcut.gui:All video files have been cleared.
INFO:deeplabcut.gui:Cleared selected videos
INFO:deeplabcut.gui:All video files have been cleared.
INFO:deeplabcut.gui:Cleared selected videos
INFO:deeplabcut.gui:All video files have been cleared.
INFO:deeplabcut.gui:Cleared selected videos
INFO:deeplabcut.gui:All video files have been cleared.
INFO:deeplabcut.gui:Cleared selected videos
INFO:deeplabcut.gui:All video files have been cleared.
INFO:deeplabcut.gui:Cleared selected videos
INFO:deeplabcut.gui:Video type set to avi
INFO:deeplabcut.gui:Video type set to avi
INFO:deeplabcut.gui:Video type set to avi
INFO:deeplabcut.gui:Video type set to avi
INFO:deeplabcut.gui:Video type set to avi
INFO:deeplabcut.gui:Video type set to avi
INFO:deeplabcut.gui:Video type set to avi
INFO:deeplabcut.gui:Video type set to avi
INFO:deeplabcut.gui:Videos added to analyze:
{'C:\\Users\\JohnsonZHOU\\Desktop\\Johnson-2026-05-31\\videos\\20250603-1.avi'}
Current video files:
{'C:\\Users\\JohnsonZHOU\\Desktop\\Johnson-2026-05-31\\videos\\20250603-1.avi'}
INFO:deeplabcut.gui.console:Starting to process video: C:\Users\JohnsonZHOU\Desktop\Johnson-2026-05-31\videos\20250603-1.avi
INFO:deeplabcut.gui.console:Loading C:\Users\JohnsonZHOU\Desktop\Johnson-2026-05-31\videos\20250603-1.avi and data.
INFO:deeplabcut.gui.console:No unfiltered data file found in C:\Users\JohnsonZHOU\Desktop\Johnson-2026-05-31\videos for video 20250603-1 and scorer DLC_Resnet50_place_cellMay31shuffle1_snapshot_best-170.
INFO:deeplabcut.gui.console:Failed to create videos from C:\Users\JohnsonZHOU\Desktop\Johnson-2026-05-31\videos\20250603-1.avi.
  File "C:\Users\JohnsonZHOU\miniconda3\envs\dlc\lib\site-packages\deeplabcut\gui\tabs\create_videos.py", line 290, in build_skeleton
    SkeletonBuilder(self.root.config)
  File "C:\Users\JohnsonZHOU\miniconda3\envs\dlc\lib\site-packages\deeplabcut\gui\widgets.py", line 527, in __init__
    QtWidgets.QDialog.__init__(self, parent)
TypeError: SkeletonBuilder.__init__() missing 1 required positional argument: 'config_path'
INFO:deeplabcut.gui:Filtering predictions ENABLED
INFO:deeplabcut.gui:Analyzing 1 video(s) with extension .avi
INFO:deeplabcut.gui.console:Analyzing videos with C:\Users\JohnsonZHOU\Desktop\Johnson-2026-05-31\dlc-models-pytorch\iteration-0\place_cellMay31-trainset95shuffle1\train\snapshot-best-170.pt
INFO:deeplabcut.gui.console:Using scorer: DLC_Resnet50_place_cellMay31shuffle1_snapshot_best-170
INFO:deeplabcut.gui.console:Starting to analyze C:\Users\JohnsonZHOU\Desktop\Johnson-2026-05-31\videos\20250603-1.avi
INFO:deeplabcut.gui.console:Video metadata:
  Overall # of frames:    15429
  Duration of video [s]:  252.98
  fps:                    60.99
  resolution:             w=1280, h=1024

INFO:deeplabcut.gui.console:Running pose prediction with batch size 8
100%|████████████████████████████████████████████████████████████████████████████| 15429/15429 [07:57<00:00, 32.29it/s]
INFO:deeplabcut.gui.console:Saving results in C:\Users\JohnsonZHOU\Desktop\Johnson-2026-05-31\videos\20250603-1DLC_Resnet50_place_cellMay31shuffle1_snapshot_best-170.h5 and C:\Users\JohnsonZHOU\Desktop\Johnson-2026-05-31\videos\20250603-1DLC_Resnet50_place_cellMay31shuffle1_snapshot_best-170_full.pickle
INFO:deeplabcut.gui.console:The videos are analyzed. Now your research can truly start!
You can create labeled videos with 'create_labeled_video'.
If the tracking is not satisfactory for some videos, consider expanding the training set. You can use the function 'extract_outlier_frames' to extract a few representative outlier frames.

INFO:deeplabcut.gui.console:Filtering with median model C:\Users\JohnsonZHOU\Desktop\Johnson-2026-05-31\videos\20250603-1.avi
WARNING: QThread::wait: Thread tried to wait on itself
WARNING:vispy:QThread::wait: Thread tried to wait on itself
  File "C:\Users\JohnsonZHOU\miniconda3\envs\dlc\lib\site-packages\deeplabcut\gui\tabs\create_videos.py", line 257, in create_videos
    videos_created = deeplabcut.create_labeled_video(
  File "C:\Users\JohnsonZHOU\miniconda3\envs\dlc\lib\site-packages\deeplabcut\utils\deprecation.py", line 198, in wrapper
    return fn(*args, **kwargs)
  File "C:\Users\JohnsonZHOU\miniconda3\envs\dlc\lib\site-packages\deeplabcut\utils\make_labeled_video.py", line 784, in create_labeled_video
    results.append(func(video))
  File "C:\Users\JohnsonZHOU\miniconda3\envs\dlc\lib\site-packages\deeplabcut\utils\make_labeled_video.py", line 960, in proc_video
    create_video(
  File "C:\Users\JohnsonZHOU\miniconda3\envs\dlc\lib\site-packages\deeplabcut\utils\make_labeled_video.py", line 1048, in create_video
    CreateVideo(
  File "C:\Users\JohnsonZHOU\miniconda3\envs\dlc\lib\site-packages\deeplabcut\utils\make_labeled_video.py", line 111, in CreateVideo
    fps = clip.fps()
  File "C:\Users\JohnsonZHOU\miniconda3\envs\dlc\lib\site-packages\deeplabcut\utils\video_processor.py", line 79, in fps
    return self.FPS
AttributeError: 'VideoProcessorCV' object has no attribute 'FPS'
INFO:deeplabcut.gui.console:Starting to process video: C:\Users\JohnsonZHOU\Desktop\Johnson-2026-05-31\videos\20250603-1.avi
INFO:deeplabcut.gui.console:Loading C:\Users\JohnsonZHOU\Desktop\Johnson-2026-05-31\videos\20250603-1.avi and data.
INFO:deeplabcut.gui.console:Error: %s
INFO:deeplabcut.gui.console:
INFO:deeplabcut.gui.console:OpenCV(4.8.0) :-1: error: (-5:Bad argument) in function 'VideoCapture'
> Overload resolution failed:
>  - Can't convert object to 'str' for 'filename'
>  - VideoCapture() missing required argument 'apiPreference' (pos 2)
>  - Argument 'index' is required to be an integer
>  - VideoCapture() missing required argument 'apiPreference' (pos 2)
```

### Anything else?

opencv version: '4.8.0'

### Code of Conduct

C-Achard:
### Motivation

Fixes #3357.

The utils/make_labeled_video file uses the provided `video` as-is, but `video` may now be provided as a Path, which may not always be supported by OpenCV calls.
Otherwise, the GUI stores selected videos in `self.files`, which is backed by a `set` to avoid duplicates. 
However, labeled video creation expects a list-like sequence of video path strings.

### Fix

- ~~Convert all uses of `video` to `str` in utils/make_labeled_video~~
- ~~Convert the selected videos to a sorted `list[str]` before calling `deeplabcut.create_labeled_video`~~
- VideoProcessor:
  - Is now a proper abstract class
  - Properties are used where relevant, and callers updated accordingly
    - Coercion to `str` is now immediate at the property level for `fname` and `sname`
    - Other validation is not yet implemented for compatibility, like raising on missing input file
    - **NOTE**: this could break any external site still calling methods instead of attributes. In this codebase this is not a concern as only utils/make_labeled_video uses it. Due to how properties are implemented I'm not sure this is the best solution IF we are worried about potential other external callers.
  - Updated docstrings
- Added minimal tests for VideoProcessorCV


deruyter92 (commented):
@C-Achard, thanks for picking this up. A few remarks:

The GUI commit seems a bit misplaced to me, since `collect_video_paths` sorts as Paths right after. 

More general, to keep a clear internal policy (i.e. prefere `Path` over `str`), I would prefer to solve it at the boundaries of our own package rather than in intermediate modules like  _create_videos.py_ and _make_labeled_video.py._ Let me know if you agree. 

I think a good alternative would be to solve this in the `VideoProcessor` base class:

```
self.fname = str(fname) if fname else fname
self.sname = str(sname) if sname else sname
```

Same for `VideoReader` and `VideoWriter`


This keeps the solution to small isolated boundaries, where it is expected to fail (i.e. OpenCV can't handle pathlib, pass `str` to OpenCV, use `Path` elsewere).

Also let's add a test to pick this up in CI.

deruyter92:
BTW, let me know if you prefer me to implement these suggestions directly in the PR

C-Achard:
@deruyter92 I think you're right, the fix is a bit too minimal and does not go well with your recent efforts for internal standardization to pathlib.

I think I will instead spend a bit of time in the VideoProcessor class, I think it could use a careful overhaul and likely some tests.
I will add some docstrings as well, the class is a bit light on that currently.
Thanks!

deeplabcut/utils/video_processor.py
```diff
@@ -69,81 +74,184 @@ def load_frame(self):
...
     def fps(self):
         return self.FPS
 
+    @property
     def counter(self):
         return self.i
 
+    @property
     def frame_count(self):
         return self.nframes
```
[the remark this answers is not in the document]
C-Achard: As mentioned in PR desc

deeplabcut/utils/video_processor.py
```diff
@@ -21,18 +21,40 @@
...
     sh and sw are the output height and width respectively.
     """
 
-    def __init__(self, fname="", sname="", nframes=-1, fps=None, codec="X264", sh="", sw=""):
+    def __init__(
+        self,
+        fname: str = "",
+        sname: str = "",
+        nframes: int = -1,
+        fps: float = None,
```
deruyter92:
```suggestion
        fps: float | None = None,
```

deeplabcut/utils/video_processor.py
```diff
@@ -21,18 +21,40 @@
...
+        codec: str = "X264",
+        sh: int = "",
+        sw: int = "",
```
deruyter92:
```suggestion
        sh: int | None = None,
        sw: int | None = None,
```

Or at least maybe `int | Literal[""]`

deeplabcut/utils/video_processor.py
```diff
@@ -21,18 +21,40 @@
...
+    ):
+        self._fname = None
+        self._sname = None
+        self.FPS = None
+        self.vid = None
+        self.svid = None
+        self.sh = 0
+        self.sw = 0
+        ###
```
deruyter92:
```suggestion
```

deeplabcut/utils/video_processor.py
```diff
@@ -21,18 +21,40 @@
...
     """Base class for a video processing unit, implementation is required for video
     loading and saving.
 
     sh and sw are the output height and width respectively.
     """
 
-    def __init__(self, fname="", sname="", nframes=-1, fps=None, codec="X264", sh="", sw=""):
+    def __init__(
+        self,
+        fname: str = "",
```
deruyter92, line 73: Do we allow None-type as well?

deeplabcut/utils/video_processor.py
```diff
@@ -69,81 +93,184 @@ def load_frame(self):
             self.i += 1
         return frame
 
+    @property
+    def fname(self):
+        return self._fname
+
+    @fname.setter
+    def fname(self, value):
+        self._fname = str(value)
```
deruyter92: Maybe we should guard this against accidental None-type setting. OR allow the field to be `None`, but then adjust the typehint to make that explicit (see other comment). 


```suggestion
    @fname.setter
    def fname(self, value: str | Path):
        self._fname = "" if value in (None, "") else str(value)
```

deeplabcut/utils/make_labeled_video.py
```diff
@@ -106,9 +106,9 @@ def CreateVideo(
...
         ny, nx = y2 - y1, x2 - x1
     else:
-        ny, nx = clip.height(), clip.width()
+        ny, nx = clip.height, clip.width
 
-    fps = clip.fps()
+    fps = clip.fps
     if isinstance(fps, float):
         if fps * 1000 > 65535:
             fps = round(fps)
```
deruyter92, line 115: (review comment concerns the line below this one) 

We should try to consistently use it either as a property, or as a field (but then not define the property):
e.g. use `frame_count` property instead of internal field `nframes`

(also consider naming to `_nframes` to be more explicit)

deeplabcut/utils/video_processor.py
```diff
@@ -69,81 +93,184 @@ def load_frame(self):
...
+        Args:
+            frame (numpy.ndarray | None): RGB frame to write. ``None`` is ignored.
+
+        Notes:
+            This method preserves the historical behavior of silently ignoring
+            ``None`` frames. Non-``None`` frames are converted from RGB to BGR
+            before being passed to OpenCV.
+        """
         if frame is not None:
             self.svid.write(np.flip(frame, 2))
```
deruyter92, line 311: maybe do the same check as in read_frame: svid should not be None

deruyter92 (approved):
Overall great improvements and good fix to the issue! 

Some suggestions about consistency for types + private/public attributes

If those are addressed looks good to me!

deruyter92:
@C-Achard, regarding the API breakage, maybe it would actually be a good idea to keep transitional backward compatibility, since it is relatively cheap to implement, e.g.: 

use properties with a new name (e.g. `h`) and keep method accessors as they were
```python
@property
def h(self):
    return self._h

@deprecated(replacement="VideoProcessor.h", since="3.1")
def height(self):
    return self._h    
```

deeplabcut/utils/video_processor.py
C-Achard: I think I prefer this version, since historically "" was the empty value for this

deeplabcut/utils/make_labeled_video.py
C-Achard: Done, thanks!

C-Achard:
> @C-Achard, regarding the API breakage, maybe it would actually be a good idea to keep transitional backward compatibility, since it is relatively cheap to implement, e.g.:
> 
> use properties with a new name (e.g. `h`) and keep method accessors as they were
> 
> ```python
> @property
> def h(self):
>     return self._h
> 
> @deprecated(replacement="VideoProcessor.h", since="3.1")
> def height(self):
>     return self._h    
> ```

It should now be more of a middle ground. I'm not sure we really gain clarity from it but it is compatible.

deeplabcut/utils/video_processor.py
```diff
@@ -49,101 +104,254 @@ def __init__(self, fname="", sname="", nframes=-1, fps=None, codec="X264", sh=""
...
+        return self._w
 
+    @deprecated(replacement="VideoProcessor.video_fps", since="3.1")
     def fps(self):
-        return self.FPS
+        return self._fps
 
-    def counter(self):
-        return self.i
+    # @deprecated(replacement="VideoProcessor.video_fps", since="3.1")
```
C-Achard: Note: this does not work for properties yet.

We could fix it in the module to allow properties but then for that case attribute assignment such as ` __deprecated_info__` may fail, so there should be a deprectated_property decorator likely. I think this would be best in a separate PR.

deruyter92: agreed. and clearly documented in the comment I think.

deruyter92:
Thanks for addressing all the comments. Looks good to me!

> It should now be more of a middle ground. I'm not sure we really gain clarity from it but it is compatible.

Yes, for me this is also still a difficult balance - to what extent backward compatibility is favored over clarity. I think the middle ground you took is both clear and backward compatible. If you think we are overdoing it in this case, feel free to revert to your `VideoProcessor.width` and `VideoProcessor.height` properties, which I think are also a clear and viable option (and relatively very safe still). Whichever you prefer.

C-Achard:
I think here problems from external use are less likely to occur compared to a higher level API. This is not really meant to be user facing, so I think I would favor maintainability for the more private API. I will re-adapt this a bit, as discussed, to be more towards what I think is the (c)lean side.

deruyter92:
yes good consideration. I agree. Looks good now! Feel free to name the properties `height` / `width` as you originally intended, or keep them `h` and `w` as they are now.

deruyter92 (approved):

MMathisLab (approved):

PR #3360, opened 2026-06-09
Prepare migration to pandas 3.0

deruyter92:
Pandas released version 3.0 in Jan 2026 which introduces breaking changes for our codebase. It would be useful to address the places where it currently breaks, so we can remove the upperbound version we introduced in https://github.com/DeepLabCut/DeepLabCut/pull/3186. See the full tracking issue #3362, for an overview of related PRs. 

**Small fixes and flagged patterns:**
As exploratory preparation, this PR implements small fixes in preparation to 3.0 migration. Also it flags breaking instances based on a targeted search for these patterns:  
- Chained `.loc` assignment (copy on write; CoW) -> mutate parent object directly instead of a subset view
- groupby(..., axis=1) -> groupby on the transposed df
- Column-level groupby(level=...) -> groupby on the transposed df
- `select_dtypes(include="object")` -> string type migration
- deprecated options like `future.no_silent_downcasting`

**Future mode testing:**
Also this PR adds a new tool for CI or local testing, using the pandas 2.3 _future_ options that warns for CoW patterns and detects where `str` type should be included instead of only `object`.

### DIFF
--- a/deeplabcut/core/inferenceutils.py
+++ b/deeplabcut/core/inferenceutils.py
@@ -1046,7 +1046,7 @@ def parse_ground_truth_data_file(h5_file):
         pass
     # Cast columns of dtype 'object' to float to avoid TypeError
     # further down in _parse_ground_truth_data.
-    # TODO @deruyter92 2026-06-10: pandas migration 3.0 requires us to add support for string dtype columns.
+    # TODO @deruyter92 2026-06-10 (#3362): pandas migration 3.0 requires us to add support for string dtype columns.
     cols = df.select_dtypes(include="object").columns
     if cols.to_list():
         df[cols] = df[cols].astype("float")
--- a/deeplabcut/core/inferenceutils.py
+++ b/deeplabcut/core/inferenceutils.py
@@ -1046,6 +1046,7 @@ def parse_ground_truth_data_file(h5_file):
         pass
     # Cast columns of dtype 'object' to float to avoid TypeError
     # further down in _parse_ground_truth_data.
+    # TODO @deruyter92 2026-06-10: pandas migration 3.0 requires us to add support for string dtype columns.
     cols = df.select_dtypes(include="object").columns
     if cols.to_list():
         df[cols] = df[cols].astype("float")
--- a/deeplabcut/gui/tracklet_toolbox.py
+++ b/deeplabcut/gui/tracklet_toolbox.py
@@ -876,6 +876,7 @@ class TrackletVisualizer:
         df = data.iloc[inds]
 
         # Uncertain keypoints are ignored
+        # TODO @deruyter92 2026-06-10: pandas migration 3.0 forbids in-place mutations of dataframe subset views.
         def filter_low_prob(cols, prob):
             mask = cols.iloc[:, 2] < prob
             cols.loc[mask] = np.nan
--- a/deeplabcut/gui/tracklet_toolbox.py
+++ b/deeplabcut/gui/tracklet_toolbox.py
@@ -876,7 +876,8 @@ class TrackletVisualizer:
         df = data.iloc[inds]
 
         # Uncertain keypoints are ignored
-        # TODO @deruyter92 2026-06-10 (#3362): pandas migration 3.0 forbids in-place mutations of dataframe subset views.
+        # TODO @deruyter92 2026-06-10 (#3362):
+        # pandas migration 3.0 forbids in-place mutations of dataframe subset views.
         def filter_low_prob(cols, prob):
             mask = cols.iloc[:, 2] < prob
             cols.loc[mask] = np.nan
--- a/deeplabcut/gui/tracklet_toolbox.py
+++ b/deeplabcut/gui/tracklet_toolbox.py
@@ -876,7 +876,7 @@ class TrackletVisualizer:
         df = data.iloc[inds]
 
         # Uncertain keypoints are ignored
-        # TODO @deruyter92 2026-06-10: pandas migration 3.0 forbids in-place mutations of dataframe subset views.
+        # TODO @deruyter92 2026-06-10 (#3362): pandas migration 3.0 forbids in-place mutations of dataframe subset views.
         def filter_low_prob(cols, prob):
             mask = cols.iloc[:, 2] < prob
             cols.loc[mask] = np.nan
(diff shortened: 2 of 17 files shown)


deruyter92:
**edit:** @C-Achard, I have changed the scope of this PR to small preparatory fixes. The whole pandas migration to 3.0 will entail more work than I initially estimated. This PR flags just a few examples and adds a useful tool for CI if we decide to push forward with this. 

We might re-consider if the whole migration will be worth it in the first-place. Pandas 3.0 is relatively new and I doubt that users will rely on it soon, so I think we may be save with restricting to 2.x for the time being, until we eventually refactor the keypoint/dataset management in its entirety. See also tracking issue #3362

deeplabcut/core/inferenceutils.py
```diff
@@ -1046,6 +1046,7 @@ def parse_ground_truth_data_file(h5_file):
         pass
     # Cast columns of dtype 'object' to float to avoid TypeError
     # further down in _parse_ground_truth_data.
+    # TODO @deruyter92 2026-06-10: pandas migration 3.0 requires us to add support for string dtype columns.
```
C-Achard:
```suggestion
    # TODO @deruyter92 2026-06-10 (#3362): pandas migration 3.0 requires us to add support for string dtype columns.
```

deeplabcut/gui/tracklet_toolbox.py
```diff
@@ -876,6 +876,7 @@ def export_to_training_data(self, pcutoff=0.1):
         df = data.iloc[inds]
 
         # Uncertain keypoints are ignored
+        # TODO @deruyter92 2026-06-10: pandas migration 3.0 forbids in-place mutations of dataframe subset views.
```
C-Achard:
```suggestion
        # TODO @deruyter92 2026-06-10 (#3362): pandas migration 3.0 forbids in-place mutations of dataframe subset views.
```

deeplabcut/pose_estimation_tensorflow/core/evaluate_multianimal.py
```diff
@@ -326,6 +326,8 @@ def evaluate_multianimal_full(
                         # Form 2D array of shape (n_rows, 4) where the last dim is
                         # (sample_index, peak_y, peak_x, bpt_index) to slice the PAFs.
                         temp = df.reset_index(level="bodyparts").dropna()
+                        # TODO @deruyter92 2026-06-10: pandas migration 3.0 - this will break if bodyparts is a str type
```
C-Achard:
```suggestion
                        # TODO @deruyter92 2026-06-10 (#3362): pandas migration 3.0 - this will break if bodyparts is a str type
```

deeplabcut/utils/pandas_future_mode.py
C-Achard, line 1: This is really nice

deeplabcut/utils/make_labeled_video.py
```diff
@@ -135,7 +135,7 @@ def CreateVideo(
     else:
         nindividuals = len(Dataframe.columns.get_level_values("individuals").unique())
         map2bp = [bplist.index(bp) for bp in all_bpts]
-        nbpts_per_ind = Dataframe.groupby(level="individuals", axis=1).size().values // 3
+        nbpts_per_ind = Dataframe.T.groupby(level="individuals").size().values // 3
```
C-Achard, line 138: One small clarification: do we want this to run with `sort=False`? Since map2id[ind] must describe the individual for the ind-th bodypart slot in the original column order, we likely want to avoid sorting at that step

examples/testscript_pytorch_multi_animal.py
```diff
@@ -22,12 +22,14 @@
...
-import deeplabcut.utils.auxiliaryfunctions as af
-from deeplabcut.compat import Engine
-from deeplabcut.pose_estimation_pytorch.config.utils import (
-    is_model_cond_top_down,
-    is_model_top_down,
-)
+# Enable pandas future mode warnings if DLC_PANDAS_FUTURE env var is set
+from deeplabcut.utils.pandas_future_mode import configure_pandas_future_if_enabled
+
+configure_pandas_future_if_enabled()
```
C-Achard, line 28: This is import order sensitive, given how it is written, right? Is this worth documenting in the utils source file? How to ensure the import order is correct here, just import before any other deeplabcut module?

deeplabcut/utils/make_labeled_video.py
C-Achard: (The code seems very reliant on implicit assumptions about the dataframe structure, as we have often mentioned. Likely this is keypoint handling refactor territory though)

C-Achard (approved):
Very good start, just a few small comments but nothing blocking I think. Thanks for taking care of this!

deeplabcut/pose_estimation_tensorflow/core/evaluate_multianimal.py
```diff
@@ -364,7 +366,8 @@ def evaluate_multianimal_full(
 
                         coords_pred = pred["coordinates"][0]
                         probs_pred = pred["confidence"]
-                        for bpt, xy_gt in df.groupby(level="bodyparts"):
+                        for bpt, xy_gt in df.T.groupby(level="bodyparts"):
```
C-Achard: I think the transpose here is redundant since it was not previously using axis=1? This causes CI to fail

```suggestion
                        for bpt, xy_gt in df.groupby(level="bodyparts", sort=False):
```

C-Achard: Attempted a fix in c9d863511e20f86fa0196aa173a38834f21836a6 @deruyter92

C-Achard:
(Note: trying to fix current CI issues seen here in #3363 )

deeplabcut/utils/make_labeled_video.py
deruyter92: absolutely. It feels dangerous even touching this code right now tbh.

deeplabcut/pose_estimation_tensorflow/core/evaluate_multianimal.py
deruyter92: This is a good catch thanks. The df was stacked a few lines before, which is why bodyparts are in the index here not in the columns (which means that `groupby(level="bodyparts")` is safe in pandas 3.0).

I have now fully reverted the changes in these lines to the original state.

examples/testscript_pytorch_multi_animal.py
deruyter92: Good point. I've added a usage note now

deeplabcut/utils/make_labeled_video.py
deruyter92: > do we want this to run with `sort=False`

Yes good point. This was actually also a potential bug in the original code (which stays hidden as long as the individuals are in lexicographical order). Applied your suggestion.

deruyter92:
@C-Achard Thanks for the review. Good points, I've applied all your suggestions.

PR #3363, opened 2026-06-10
Add certifi SSL fallback for urlopen

C-Achard:
### Motivation

The latest Ci runner on Windows seems to be having issues with the default SSL context, currently it fails while loading certificates from the Windows certificate store with

```python
ssl.SSLError: [ASN1: NOT_ENOUGH_DATA] not enough data
```

### Fix

Import ssl and add _urlopen_with_certifi_fallback in tests/conftest.py to handle ssl.SSLError by falling back to an SSLContext using certifi's CA bundle. Replace direct urllib.request.urlopen call in unzip_from_url with the helper so test data downloads succeed on systems with incomplete SSL trust stores; re-raise the original error if certifi is not available.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -11,6 +11,7 @@
 
 import os
 import pickle
+import ssl
 import urllib.request
 import zipfile
 from io import BytesIO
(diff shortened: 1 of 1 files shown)


C-Achard:
This requires broader patching as in c4949456f0c3453639e5cca61f749e048e7b2a4e, since the failure affects everything including torch hub downloads. We might want to revert this patch once a fix is released for Win runners,

deruyter92 (approved):

deruyter92:
Good fix for now! Let's merge as soon as possible so we can use CI

C-Achard:
@deruyter92 I think this is fixed now since CI works in other PRs?

PR #3382, opened 2026-06-26
Prepare TensorFlow deprecation - add `deeplabcut.api`

deruyter92:
### Preparation for full TensorFlow deprecation
1. Remove `deeplabcut.compat`  and split into **PyTorch-only `deeplabcut.api`** and TF-only `deeplabcut.tensorflow_compat`
2. Add TensorFlow routing and deprecation utilities which warns the user.

### Background
TensorFlow support will be dropped soon and a clean preparation is needed to make the transition smooth. Although the pose_estimation code was already cleanly branched in `deeplabcut/compat.py`, this brought the disadvantage that the main entrypoint for DeepLabCut is actually mixed API, which was not ideal to test and remains a small obstacle to eventually remove TF cleanly.

This PR makes a large step towards clean TF deprecation, by splitting the `compat` module into our new clean user-facing API (Pytorch-only) with a fallback for TensorFlow instead of mixed API. It introduces some new packages and removes `deeplabcut.compat` as the entry point for the public API.

### New
**`deeplabcut.api`** : the new home for user-facing API functions. Currently covers all pose estimation functions (train_network, evaluate_network, analyze_videos, analyze_images, etc.), but can cover other main API as well.

**`deeplabcut.tensorflow_compat`** — an isolated compatibility layer that holds the TensorFlow-specific branches, split out of compat.py. This module is explicitly marked as temporary and will be deleted wholesale when TF support is dropped.

**`deeplabcut.api._tf_routing`** — a new decorator-based routing system (`@with_tensorflow_fallback`) that automatically intercepts calls to functions in deeplabcut.api, resolves the engine and delegates to `deeplabcut.tensorflow_compat` handling TF-only legacy kwargs. A `DLCDeprecationWarning` is emitted to inform user about the migration.

### Next steps
This PR intentionally scopes to pose estimation only. (I included `generate_training_dataset` as well first, it's probably better to work in incremental changes, rather than one big PR).

 Several other areas still go through compat or have TF branches inline and need similar treatment in follow-up PRs:
- Dataset creation (`generate_training_dataset`) — the largest remaining concern. 
- `deeplabcut.utils` — scattered TF-conditional helpers.
- `deeplabcut.modelzoo` — TF-specific download / inference paths.
- Other smaller callsites in the GUI and 3D/tracking modules.

The goal is that every user-facing function ultimately lives under deeplabcut.api with a clean, PyTorch-only signature, while all TF code is quarantined in deeplabcut.tensorflow_compat and removable in a single future commit.

### Test coverage
- `tests/api/test_pose_estimation.py` (480 lines) — smoke tests for the new deeplabcut.api pose estimation surface: parameter forwarding, engine routing, legacy kwarg handling.
- `tests/api/test_tf_routing.py` (360 lines) — unit tests for the _tf_routing machinery: engine resolution, renamed/dropped param warnings, gputouse normalisation, multi-shuffle validation.
- `tests/core/test_visualization.py` (191 lines) — new coverage for deeplabcut.core.visualization helpers.
- `tests/pose_estimation_pytorch/apis/test_apis_utils.py` (102 lines) — new coverage for PyTorch API utilities.


deruyter92:
(Needs formatted docstrings and typed configs still)

deruyter92:
- Rebased on main with new docstrings, pathlib internals etc. 
- Dissected TF from mixed functions and added to `deeplabcut.api`: modelzoo inference and dataset creation (`create_pretrained_dataset`), project creation ( SA and MA), post-processing, refinement
- Made the machinery for TF routing a bit more robust: no silent assumption that config is first arg, allow custom predicate for routing (e.g. for specific TF models, instead of shuffle/config/engine keywords)
- Added tests for deeplabcut.api

Will do a self-review tomorrow, and address failing tests. But otherwise nearly done.

deeplabcut/api/_tf_routing.py
```diff
@@ -0,0 +1,246 @@
...
+
+def _resolve_legacy_kwargs(
+    kwargs: dict,
+    renamed_params: dict[str, str],
+    dropped_params: list[str],
+    normalize_gputouse: bool = False,
+) -> dict:
+    """Resolve legacy TensorFlow kwargs to canonical (PyTorch) kwargs."""
+
+    if normalize_gputouse and (gpu := kwargs.get("gputouse")):
```
C-Achard: I think this handles GPU index 0 incorrectly since it's falsey

```suggestion
    if normalize_gputouse and "gputouse" in kwargs:
        gpu = kwargs["gputouse"]
        kwargs["gputouse"] = _normalize_gputouse(gpu)
```

deeplabcut/api/_tf_routing.py
```diff
@@ -0,0 +1,246 @@
...
+        # Normalize parameter "gputouse" to torch device string and rename
+        kwargs["gputouse"] = _normalize_gputouse(gpu)
+        renamed_params["gputouse"] = "device"
```
C-Achard: Also, renamed_params is from the decorator, so it mutates the decorator config
```python
effective_renames = dict(renamed_params)

if normalize_gputouse and "gputouse" in kwargs:
    kwargs["gputouse"] = _normalize_gputouse(kwargs["gputouse"])
    effective_renames["gputouse"] = "device"
```
and operate on `effective_renames`

deeplabcut/api/_tf_routing.py
```diff
@@ -0,0 +1,246 @@
...
+        )
+    ]
+    has_var_positional = any(p.kind is inspect.Parameter.VAR_POSITIONAL for p in sig.parameters.values())
+    if len(args) > len(params) and not has_var_positional:
+        raise TypeError(f"too many positional arguments (expected at most {len(params)}, got {len(args)})")
+    for param, value in zip(params, args, strict=False):
+        if param.name in unified:
+            raise TypeError(f"got multiple values for argument '{param.name}'")
+        unified[param.name] = value
+    return unified
```
C-Achard, line 103: _positionals_as_kwargs() creates a normalized view, but `when` bypasses it and inspects the original args/kwargs. So custom routing does not match between positional and keyword calls. For example, positional model_name="dlcrnet" is routed to PyTorch. Maybe send the unified mapping instead to `when`

deeplabcut/api/_tf_routing.py
```diff
@@ -0,0 +1,246 @@
...
+        @functools.wraps(fn)
+        def wrapper(*args, **kwargs):
+            unified = _positionals_as_kwargs(sig, args, kwargs)
+
+            if when is not None:
+                # Custom condition routing (e.g. modelzoo functions)
+                route_to_tf = when(*args, **kwargs)
+            else:
+                # Default: engine-based routing (from shuffle / config)
+                route_to_tf = _resolve_engine(unified) == Engine.TF
```
C-Achard, line 181: This should happen after legacy kwargs reolution, I think, otherwise things like export_model(cfg_path="config.yaml") may fail because it looks for "config" directly

deeplabcut/api/_tf_routing.py
```diff
@@ -0,0 +1,246 @@
...
+
+            kwargs.pop("engine", None)
```
C-Achard, line 183: Should this work on args as well?

deeplabcut/api/_tf_routing.py
```diff
@@ -0,0 +1,246 @@
...
+def _resolve_engine(unified_kwargs: dict) -> Engine:
+    """Resolve engine from explicit engine parameter or shuffle metadata.
+
+    Args:
+        unified_kwargs: Keyword arguments with positionals mapped to names via
+                        ``_positionals_as_kwargs``.
+    """
+    engine = unified_kwargs.get("engine")
+    if engine is not None:
+        return engine
```
C-Achard: Is this robust to passing both Enum and Enum values?

deeplabcut/api/_tf_routing.py
```diff
@@ -0,0 +1,246 @@
...
+
+    from deeplabcut.core.config.utils import read_config
+
+    shuffles = _shuffles_from_kwargs(unified_kwargs)
```
C-Achard, line 233:
```suggestion
    shuffles = _shuffles_from_kwargs(unified_kwargs)
    if not shuffles:
        raise ValueError("Shuffles must contain at least one index")
```
to prevent a less clear failure from pop()

docs/api/deeplabcut.analyze_videos.rst
C-Achard, line 1: SHould we rely on dev docs now?

tests/api/test_tf_routing.py
```diff
@@ -0,0 +1,512 @@
...
+        (0, "cuda:0"),
+        (1, "cuda:1"),
+        ("cuda:0", "cuda:0"),
+        ("gpu:1", "cuda:1"),
+        ("cpu", "cpu"),
+    ],
+)
+def test_normalize_gputouse(gputouse, expected):
+    assert tf_routing._normalize_gputouse(gputouse) == expected
+
```
C-Achard, line 42:
```suggestion
@pytest.mark.parametrize(
    "gputouse, expected",
    [
        (0, "cuda:0"),
        (1, "cuda:1"),
        ("cuda:0", "cuda:0"),
        ("gpu:1", "cuda:1"),
        ("cpu", "cpu"),
    ],
)
def test_normalize_gputouse(gputouse, expected):
    assert tf_routing._normalize_gputouse(gputouse) == expected


def test_resolve_legacy_kwargs_normalizes_gpu_zero():
    with pytest.warns(DLCDeprecationWarning, match="gputouse"):
        result = tf_routing._resolve_legacy_kwargs(
            {"gputouse": 0},
            renamed_params={},
            dropped_params=[],
            normalize_gputouse=True,
        )

    assert result == {"device": "cuda:0"}

```

tests/api/test_tf_routing.py
```diff
@@ -0,0 +1,512 @@
...
+
+# ---------------------------------------------------------------------------
+# _resolve_legacy_kwargs
```
C-Achard, line 80: Custom routing is only tested with model_name as a kwarg, maybe add a canonical signature and verify that both positional and keyword forms are OK as per comments in the routing file

tests/api/test_tf_routing.py
```diff
@@ -0,0 +1,512 @@
...
+
+def test_resolve_legacy_kwargs_normalize_gputouse_raises_when_both_given():
+    with pytest.raises(TypeError, match="Cannot specify both 'gputouse'"):
+        tf_routing._resolve_legacy_kwargs(
+            {"gputouse": 1, "device": "cuda:1"},
+            renamed_params={},
+            dropped_params=[],
+            normalize_gputouse=True,
+        )
+
```
C-Achard, line 171:
```suggestion
def test_resolve_legacy_kwargs_normalize_gputouse_raises_when_both_given():
    with pytest.raises(TypeError, match="Cannot specify both 'gputouse'"):
        tf_routing._resolve_legacy_kwargs(
            {"gputouse": 1, "device": "cuda:1"},
            renamed_params={},
            dropped_params=[],
            normalize_gputouse=True,
        )

def test_resolve_legacy_kwargs_does_not_mutate_renamed_params():
    renamed_params = {
        "keepdeconvweights": "load_head_weights",
    }

    with pytest.warns(DLCDeprecationWarning, match="gputouse"):
        result = tf_routing._resolve_legacy_kwargs(
            {"gputouse": 1},
            renamed_params=renamed_params,
            dropped_params=[],
            normalize_gputouse=True,
        )

    assert result == {"device": "cuda:1"}
    assert renamed_params == {
        "keepdeconvweights": "load_head_weights",
    }
```

deeplabcut/api/pose_estimation.py
```diff
@@ -0,0 +1,986 @@
...
+        plotting=plotting,
+        pcutoff=pcutoff,
+        bbox_pcutoff=bbox_pcutoff,
+        plot_skeleton=plot_skeleton,
+        ctd_conditions=ctd_conditions,
+    )
+
+
+@deprecated(replacement="deeplabcut.analyze_images", since="3.1")
+@with_tensorflow_fallback(normalize_gputouse=True, dropped_params=["frametype"])
```
C-Achard, line 608: Should this be a deprecated rename and forwarded instead?

tests/api/test_pose_estimation.py
C-Achard, line 1: Good PyTorch coverage, but could we add a small set of integration tests to mock read_config and get_shuffle_engine instead, covering one TF project, one PyTorch project, positional custom routing, and cfg_path? So as to avoid routing regressions

C-Achard (commented):
Good start,I really like the design overall, I just think some of the current routing might need to be made safer; happy to discuss if needed!

deeplabcut/api/_tf_routing.py
deruyter92: good catch. fixed now

deeplabcut/api/_tf_routing.py
deruyter92: fixed now! tnx

deeplabcut/api/_tf_routing.py
deruyter92: `when` uses `params` coming from the unified args after `_positionals_as_kwargs()`. As you suggested.

deeplabcut/api/_tf_routing.py
deruyter92: good point. Addressed now

deeplabcut/api/_tf_routing.py
deruyter92: fixed, uses `Engine(engine)` now

deeplabcut/api/_tf_routing.py
deruyter92: good suggestion

docs/api/deeplabcut.analyze_videos.rst
deruyter92: Yes. Let's remove the .rst files in a separate PR? For know I'll keep it consistent with the others.

deeplabcut/api/pose_estimation.py
deruyter92: it is deprecated and forwarding now right? The decorators combine well, is that the problem that you wanted to point out?

tests/api/test_pose_estimation.py
deruyter92: Added a small set of tests now!

deeplabcut/api/_tf_routing.py
deruyter92: yes should also be popped on the unified args. fixed now.

tests/api/test_tf_routing.py
deruyter92: added some tests for this. Good point

deruyter92:
@C-Achard thanks for your review. I've made the tf_routing more robust now. I've addressed all your comments, and added more tests.

I will do another iteration next week to see if there are more tests to be added, and try it out in practice a low. It is a high-impact PR, so I agree that it should be well-tested.

deeplabcut/core/config/utils.py
```diff
@@ -238,7 +238,7 @@ def wrapper(*args, **kwargs):
 # -----------------------------------------------------------------------------
 
 
-@deprecated(replacement="deeplabcut.core.config.ProjectConfig", since="3.1")
+@deprecated(replacement="deeplabcut.core.config.ProjectConfig", since="3.0.1")
```
C-Achard: A small question/design curiosity when I see this, should we have a file that defines several consts with version numbers mapping to deprecation, so that whenever there is a round of deprecating something, there is one single source for the version, minimizing the need to globally search for version numbers, and avoiding changing version numbers for an unrelated round of deprecation or on the wrong API?

deeplabcut/gui/tabs/manage_project.py
```diff
@@ -28,6 +28,12 @@
 class ManageProject(DefaultTab):
     def __init__(self, root, parent, h1_description):
         super().__init__(root, parent, h1_description)
+
+        self._reload_timer = QTimer(self)
```
C-Achard, line 32: Should these defs be moved to the base class?

deeplabcut/gui/tabs/analyze_videos.py
```diff
@@ -54,6 +54,11 @@ class AnalyzeVideos(DefaultTab):
     def __init__(self, root, parent, h1_description):
         super().__init__(root, parent, h1_description)
 
+        self._reload_timer = QTimer(self)
```
C-Achard, line 57: As mentioned elsewhere, might be best moved to the base class instead ? Even if disabled by default

C-Achard (commented):
Just a few preliminary comments while going through this again, let's perhaps discuss which areas would benefit from further review/testing as it has grown quite a bit.

deeplabcut/gui/tabs/manage_project.py
deruyter92: Done! see https://github.com/DeepLabCut/DeepLabCut/pull/3441

deeplabcut/core/config/utils.py
deruyter92: Done! see https://github.com/DeepLabCut/DeepLabCut/pull/3444

deeplabcut/api/_tf_routing.py
C-Achard: But modifying unified does not change the args which are then passed to TF right? And TF should not get engine

deeplabcut/api/_tf_routing.py
```diff
@@ -0,0 +1,264 @@
...
+    dropped_params: list[str],
+    normalize_gputouse: bool = False,
+) -> dict:
+    """Resolve legacy TensorFlow kwargs to canonical (PyTorch) kwargs."""
+
+    effective_renames = dict(renamed_params)
+
+    if normalize_gputouse and "gputouse" in kwargs:
+        # Normalize parameter "gputouse" to torch device string and rename
+        kwargs["gputouse"] = _normalize_gputouse(kwargs["gputouse"])
```
C-Achard, line 276: Does this handle None?

C-Achard:
```suggestion
        if normalize_gputouse and "gputouse" in kwargs:
            gpu = kwargs["gputouse"]
            if gpu is None:
                kwargs.pop("gputouse")
            else:
                kwargs["gputouse"] = _normalize_gputouse(gpu)
                effective_renames["gputouse"] = "device"
```

deeplabcut/api/_tf_routing.py
```diff
@@ -0,0 +1,264 @@
...
+        sig = inspect.signature(fn)
+
+        @functools.wraps(fn)
+        def wrapper(*args, **kwargs):
+            unified = _positionals_as_kwargs(sig, args, kwargs)
+
+            # Normalize renamed params in unified so routing sees canonical names
+            for old, new in (renamed_params or {}).items():
+                if old in unified and new not in unified:
+                    unified[new] = unified.pop(old)
```
C-Achard: Same check you did below for both being present in _resolve_legacy_kwargs may help here too
```suggestion
            for old, new in (renamed_params or {}).items():
                if old in unified:
                    if new in unified:
                        raise TypeError(
                            f"Cannot specify both '{old}' (deprecated) "
                            f"and '{new}'. Use '{new}' only."
                        )
                    unified[new] = unified.pop(old)
```

deeplabcut/api/_tf_routing.py
```diff
@@ -0,0 +1,264 @@
...
+            (PyTorch) function. A warning is emitted when they are dropped.
+            Note: applied **only on the PyTorch path**. The TF compat functions accept these parameters natively.
+        normalize_gputouse (bool): Resolve the old TF ``gputouse`` parameter to the new canonical PyTorch ``device``
+            parameter. Raises a TypeError if both are specified. Equivalent to ``renamed_params={"gputouse": "device"}``
+            with the additional normalization of legacy formats (``int``, ``"gpu:0"``) to ``"cuda:X"``.
+            Note: applied **only on the PyTorch path**. When ``True``, setting ``gputouse`` in ``dropped_params`` is
+            redundant (``normalize_gputouse`` always renames ``gputouse`` to ``device`` first).
+        when (Callable | None): A callable ``(*args, **kwargs) -> bool`` that determines whether to route to the
+            TensorFlow fallback.  When ``None`` (the default), the engine is resolved from shuffle metadata via
+            ``_resolve_engine``.  Supply a custom callable for engine-less routing (e.g. modelzoo functions).
```
C-Achard: SHould the type hint be `Callable[[dict[str, Any]], bool]` now ?

deeplabcut/api/_tf_routing.py
C-Achard: Actually it seems that if shuffles is a generator, it will always pass even if empty. So this should be typecast to list.

```suggestion
    shuffles = list(_shuffles_from_kwargs(unified_kwargs))
```

deeplabcut/__init__.py
```diff
@@ -128,9 +108,44 @@
     "transformer_reID": (".pose_tracking_pytorch", "transformer_reID"),
 }
 
+# API exports are lazily loaded from the pose_estimation API facade.
```
C-Achard, line 111: This will "break" the API docs in that the direct (still valid) imports like `deeplabcut.analyze_videos` will no longer be recognized.
I'm mentioning it, I doubt there is any easy clean fix for this, nor is it a big problem in the first place, but maybe worth mentioning.

C-Achard: (Dev docs are cleaner with the new version imo. I also checked again that all imports are still supported)

deeplabcut/api/_tf_routing.py
C-Achard, line 1: I checked with your new knowledge-index YAML generator tool that all symbols were correctly ported; seems all good!

deeplabcut/generate_training_dataset/multiple_individuals_trainingsetmanipulation.py
C-Achard, line 1: L.225 :             ``deeplabcut.compat.DEFAULT_ENGINE``.
Does this still exist ?

deeplabcut/generate_training_dataset/trainingsetmanipulation.py
```diff
@@ -920,7 +917,7 @@ def create_training_dataset(
         engine (Engine, optional): Whether to create a pose config for a Tensorflow or
             PyTorch model. Defaults to the value specified in the project configuration
             file. If no engine is specified for the project, defaults to
-            ``deeplabcut.compat.DEFAULT_ENGINE``.
+            ``deeplabcut.core.DEFAULT_ENGINE``.
```
C-Achard:
```suggestion
            ``deeplabcut.core.engine.DEFAULT_ENGINE``.
```

deeplabcut/modelzoo/video_inference.py
C-Achard, line 1: More of a discussion point, but is it correct that 
`deeplabcut.modelzoo.video_inference` 
is now PT only, whereas 

- `deeplabcut.api.modelzoo_inference.video_inference_superanimal`
- `deeplabcut.video_inference_superanimal`

have TF routing ? Maybe having a private torch only function and a single public TF wrapper would be cleaner ?
At the very least a direct import should warn if possible while keeping the current public call site, e.g.
```python
# deeplabcut/modelzoo/video_inference.py

def _video_inference_superanimal_pytorch(...):
    ...

# deeplabcut/api/modelzoo_inference.py

@with_tensorflow_fallback(
    when=lambda params: params.get("model_name") in _TENSORFLOW_MODELS,
    ...
)
def video_inference_superanimal(...):
    from deeplabcut.modelzoo.video_inference import (
        _video_inference_superanimal_pytorch,
    )

    return _video_inference_superanimal_pytorch(...)
# deeplabcut/modelzoo/video_inference.py

def video_inference_superanimal(*args, **kwargs):
    warnings.warn(
        "Import video_inference_superanimal from deeplabcut "
        "or deeplabcut.api instead.",
        DLCDeprecationWarning,
        stacklevel=2,
    )

    from deeplabcut.api.modelzoo_inference import (
        video_inference_superanimal as public_api,
    )

    return public_api(*args, **kwargs)
```

Also is dlcrnet the only TF model ?
Since the old implementation selected the backend through models_to_framework.json, while the new router appears to select TensorFlow only when model_name == "dlcrnet".

tests/api/test_tf_routing.py
```diff
@@ -0,0 +1,699 @@
...
+def test_positionals_as_kwargs_var_positional_allows_extra():
+    sig = inspect.signature(lambda a, *args: None)
+    result = tf_routing._positionals_as_kwargs(sig, (1, 2, 3), {})
+    assert result == {"a": 1}
+
+
+# ---------------------------------------------------------------------------
+# _resolve_legacy_kwargs
+# ---------------------------------------------------------------------------
+
```
C-Achard, line 82: Just to be sure, this tests  _resolve_legacy_kwargs, which is torch only right? Could we add tests that route to TF while both a legacy and current name are provided?

tests/api/test_pose_estimation.py
```diff
@@ -0,0 +1,701 @@
...
+        indices=None,
+        modelprefix="",
+        snapshot_index=None,
+        detector_snapshot_index=None,
+        dest_folder=None,
+    )
+
+
+# ---------------------------------------------------------------------------
+# TensorFlow-engine dispatch tests
```
C-Achard, line 427: Maybe add one or two integration tests using mocked read_config and get_shuffle_engine, for example analyze_videos and export_model(cfg_path=...)?
To catch issues with real API funcs.

tests/api/test_tf_routing.py
```diff
@@ -0,0 +1,699 @@
...
+from deeplabcut.core.deprecation import DLCDeprecationWarning
+from deeplabcut.core.engine import Engine
+
+# ---------------------------------------------------------------------------
+# _normalize_gputouse
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize(
+    "gputouse, expected",
```
C-Achard, line 30: Maybe add None case here ?

tests/api/test_tf_routing.py
C-Achard, line 1: Maybe add where relevant:
```py
import deeplabcut.api

@pytest.mark.parametrize("name", deeplabcut.api.__all__)
def test_migrated_export_available_from_public_facades(name):
    import deeplabcut
    import deeplabcut.api

    assert name in deeplabcut.__all__
    assert name in deeplabcut.api.__all__
    assert getattr(deeplabcut, name) is getattr(deeplabcut.api, name)
```

C-Achard (commented):
Some more comments!

Let me know if you would like any further review on any other points.

deeplabcut/api/_tf_routing.py
deruyter92: I've centralized and deduplicated this now!

deeplabcut/api/_tf_routing.py
deruyter92: good catch, `None` was missing from the normalization. I added it in `_normalize_gputouse` rather than dropping it, so users get feedback for incorrect usage of `gputouse` (+ `device`).

deeplabcut/api/_tf_routing.py
deruyter92: Thanks for staying alert. This was not fixed properly indeed for cases where `engine` is passed as positional arg. I've made it keyword-only now and added a validation step that runs at decoration time.

Note that making engine Keyword only changes the API! In fact I took the liberty to make more parameters keyword-only, since 12 positional args could get confusing easily. Let me know if you agree. 

see 4fc2d18d0022d93fd04d37e4a4643bef1619abef

deeplabcut/api/_tf_routing.py
deruyter92: good point. I've normalized the return type to list now.

docs/api/deeplabcut.analyze_videos.rst
deruyter92: see https://github.com/DeepLabCut/DeepLabCut/pull/3442/changes

deeplabcut/modelzoo/video_inference.py
deruyter92: good point, and I think in a broader sense we should streamline this for all functions. Currently pose_estimation_pytorch is a clear exported module on its own, which is re-exported in deeplabcut.api + at the top-level now (including the tf-routing decorator). Docstrings are largely duplicated. 

I think after the full stack is merged we should revisit this and see if we need to restructure the exports / module locations a bit. "

For now I've addressed your comment directly in 5cae12f8f692945ab215ea3dc6cb9fb1209461c7, but keeping this comment open for later discussion.

tests/api/test_pose_estimation.py
deruyter92: good suggestion. See 58bda12f57c94fbf05aecb6e1662477d6986e9d1

deruyter92:
merged #3442 and #3444  into this base branch

deeplabcut/__init__.py
deruyter92: ok I agree that this needs to be improved. I implemented manual `__getattr__` lazy loading following PEP 562, but that would have required some additions, like improving type checking to find modules before they are actually accessed as attribute. 

After some more research, I think that [lazy_loader](https://github.com/scientific-python/lazy-loader) is a good alternative (also used by napari, scikit-image) that keeps everything in one place (a .pyi stub) and works with type-checking. See #3459

C-Achard:
@deruyter92  While testing the GUI with TF, I had a small issue when switching backends that cause the Create training dataset menu to not switch properly to the TF models list despite confirming. Changing again seemed to solve it.
I haven't had time to find the exact source, let me know if you'd like a more thorough report

C-Achard:
Would it also be possible to cap the amount of deprecation warnings for tensorflow?

C-Achard (approved):
Tested again in the GUI, performance is equivalent, could not find other issues than those listed above

PR #3383, opened 2026-06-29
[Dev docs] Clean up workflows and mike usage

C-Achard:
## Summary

This PR cleans up and reduces duplication in the new versioned developer documentation setup.

## Scope

The docs deployment sequence should now be clearer and less duplicated:

1. Build and deploy the main Jupyter Book docs
2. Preserve existing versioned dev docs under `/dev/`
3. Deploy the current main-branch developer docs to `/dev/main/`
4. Manage release-specific dev docs versions manually, with optional `/dev/latest/` aliasing for intended latest

## Changes

- Renames workflow display names with clearer prefixes such as `Docs /` and `Code /` for easier scanning in GitHub Actions.
- Adds a reusable dev-docs setup action to avoid repeating steps.
- Simplifies the dev-docs build workflow by reusing the shared setup action and making the MkDocs config path configurable.
- Refactors the main docs deployment workflow so it builds and deploys the Jupyter Book first, then deploys the main branch developer docs through the reusable Mike workflow.
- Refactors manual dev-docs version manager workflow to call the shared Mike deployment workflow.
- Adds support for optionally marking a manually deployed dev-docs version as `latest` and setting the `/dev/` default version accordingly.
- Updates the Jupyter Book TOC so the Developer Documentation entry points to the versioned dev-docs site.

## Notes

The main branch deployment now publishes `/dev/main/` only. The `/dev/latest/` alias is managed through the manual dev-docs version workflow when `make_latest` is selected.


deruyter92:
@C-Achard very helpful! The cleaner workflow looks good and is very easy to use. 

Just to document what we discussed today here as well: 
- we prefer the default to point to the latest _release_, not to the latest bleeding edge code on main.
- we prefer to not add "latest" as dropdown item, to avoid confusion with the latest bleeding edge code on main. Instead I just added the "latest-release" alias which shows up next to the primary title in the dropdown.
- as you suggested, I added an outdated banner that helps to redirect the user to the default dev docs (when they end up on a non-latest-release page).

PR #3350, opened 2026-05-29
Migrate to `pathlib.Path` instead of `str`  (first phase)

deruyter92:
### Motivation
We currently use `pathlib.Path` and `str` mixed inconsistently throughout the codebase. Settling on `pathlib.Path` could make the codebase more uniform, and fix issues with brittle path resolvement, (e.g. see #3348). 

This PR builds on top of #3349, which introduces safer path resolution. 

### Policy

- **Internal:** use `pathlib.Path` for path math and state.
- **Public API:** accept `str | Path` at entry; normalize with `Path(...).absolute()` internally.
- **Boundaries:** convert with `str()` / `os.fspath()` only where required:
  - Qt widgets (`QLineEdit`, `QFileDialog`, list item text)
  - `video_sets` dict keys in `config.yaml`
  - OpenCV / PIL / `VideoReader` / subprocess / pandas paths
  - (legacy helpers that still expect `str` -> should be replaced on long-term as well)

### Scope

This PR starts with addressing many places to clearly enforce Path-centric API, but leaves quite some legacy helpers untouched for now, to avoid a single big switch. See details (autogenerated) below:

---

#### `create_project` (2D core)

- **`new.py`**, **`add.py`**, **`demo_data.py`**: `Path` internally; `os.fspath` at `video_sets` / `VideoReader` / copy-mklink boundaries
- **`create_new_project`**: returns `Path | Literal["nothingcreated"]` (was `str`)
- Removed legacy `realpath` / `readlink` fallback for `video_sets` keys; use `Path(video).absolute()`

#### GUI

- **`window.py`**: `config_path: Path | None`, `files: set[Path]`; derived folders (`project_folder`, `models_folder`, …) return `Path`; `os.fspath` for Qt line edits / recent files
- **`widgets.py`**: `ConfigEditor` stores `Path`; `DragDropListView` keeps **str in Qt items** (by design), normalizes on add
- **`components.py`**: video/snapshot/conditions selection uses `Path`; `os.fspath` at `QFileDialog`
- **Tabs updated:** `open_project`, `manage_project`, `create_project`, `analyze_videos`, `modelzoo`
- Other tabs mostly pass `root.config_path` (`Path`) through to DLC APIs — no per-tab refactor

#### `pose_estimation_pytorch` (partial)

- **Loaders / metadata:** `Loader.project_root`, `DLCLoader.project_path`, `PoseMetadata.project_path` are `Path`
[...]

PR #3370, opened 2026-06-15
Update GUI assets import

C-Achard:
As part of the efforts for path standardization in #3350, a few GUI assets-related calls could be improved.

This focuses on image and logo paths and aims to make the importing of these assets safer and more centralized.

## Automated summary

This pull request refactors how GUI assets (like icons and images) are loaded and used throughout the DeepLabCut GUI. The main improvement is the introduction of a centralized utility module, `gui_assets.py`, which provides functions to access resources bundled with the application. This change replaces all previous direct file path constructions and `QIcon`/`QPixmap` usage with these new helper functions, making resource management more robust and portable.

The most important changes are:

**Introduction of centralized GUI asset utilities:**
* Added a new module, `deeplabcut/gui/gui_assets.py`, which provides functions such as `icon_from_resource` and `pixmap_from_resource` to load icons and images from the application's assets directory in a consistent way.

**Refactoring icon and image loading throughout the GUI:**
[...]

PR #3385, opened 2026-07-07
GUI: use importlib.resources instead of manual path construction

deruyter92:
This PR builds on top of #3349 and is related to #3350. Together these PRs enforce more robust path resolution across the codebase. 

**Summary**
This PR replaces all manual path constructions in the GUI (for accessing the assets, like the logo, etc) with `importlib.resources`. This loads correctly regardless of the process working directory, editable installs, or where the user launched the GUI from and should be the canonical way to load GUI assets. It was already implemented in some places, but now made more consistent.


deruyter92:
ok nevermind, I forgot this was already taken care of in #3370, closing this PR.

PR #3387, opened 2026-07-10
Refactor `ctd_conditions` / `conditions_provider` + add `ConditionsConfig` + fix `InferenceConfig`

deruyter92:
**Issues**
This PR addresses two issues with the way conditions are specified for inference with a BUCTD model: 
1. With the introduction of typed configs, the InferenceConfig was too simplified: different schemas can be used to specify the conditions, but only a plain (unvalidated) `dict` was supported.
2. The implementation of `CondProvider` in _pose_estimation_pytorch/data/ctd.py_ was slightly incorrect and misleading: 
   - the two subclasses `CondFromFile` and `CondFromModel` serve a different thing (either a provider that can read conditions from a file, or just a config object that still needs downstream loading of the BU model). 
   - CondProvider was denoted as abstract base class, but both subclasses call super(). on the parent methods.
   - inference API required the CondFromModel configuration or a dict, which now should be replaced with a proper DLCBaseConfig instance. 

**Implementation**
- This PR introduces a `ConditionsConfiguration` schema with two subclasses `ConditionsFileConfig` and `ConditionsModelConfig` to solve the oversimplified schema design. 
- The implementation of `CondProvider` is refactored: only `CondFromFile` is kept, as it literally provides conditions from a file. The `CondFromModel` configuration class is replaced with `ConditionsModelConfig`. 
- inference API is refactored to accept `ConditionsConfig` or their raw arguments, and normalized + validated at the beginning.


deruyter92:
**Edit** 
now added a third schema subclass which was accepted in the original design as well: `ConditionsShuffleConfig` for shuffle references that are not yet resolved to a model snapshot on disk. I added a `resolve_from_conditions` method to do the canonical resolution to a `ConditionModelConfig` (which finds the snapshot on disk) for BUCTD live inference.

deeplabcut/pose_estimation_pytorch/apis/analyze_images.py
```diff
@@ -309,6 +306,10 @@ def analyze_images(
     model_cfg = PoseConfig.from_any(model_cfg_path)
     pose_task = Task(model_cfg["method"])
 
+    ctd_conditions = ctd_conditions or model_cfg["inference"]["conditions"]
```
C-Achard: Does this need to be resolved in all cases? Or only when the task is CTD?

deeplabcut/pose_estimation_pytorch/apis/videos.py
```diff
@@ -490,22 +486,15 @@ def analyze_videos(
...
-            )
-        # TODO @deruyter92: decide on typed / plain dict
-        elif isinstance(ctd_conditions, dict):
-            cond_provider = get_condition_provider(
-                condition_cfg=ctd_conditions,
-                config=config,
-            )
-        else:
-            cond_provider = ctd_conditions
+            raise ValueError("CTD conditions are required for video analysis with cond-top-down models")
```
C-Achard: Same comment as before, should it be conditional here?
```suggestion
    cond_provider = None
    ctd_conditions = None
    if loader.pose_task == Task.COND_TOP_DOWN:
        ctd_conditions = ctd_conditions or loader.model_cfg["inference"]["conditions"]
        
        if ctd_conditions is not None:
            ctd_conditions = ConditionsModelConfig.resolve_from_conditions(ctd_conditions, config=config)
        else:
            raise ValueError("CTD conditions are required for video analysis with cond-top-down models")
```

deeplabcut/pose_estimation_pytorch/apis/analyze_images.py
```diff
@@ -462,6 +453,10 @@ def analyze_image_folder(
     """
     model_cfg = PoseConfig.from_any(model_cfg)
```
C-Achard, line 450: Same here, you could grab pose_task and only compute conditions if the model is CTD

deeplabcut/pose_estimation_pytorch/apis/analyze_images.py
```diff
@@ -462,6 +453,10 @@ def analyze_image_folder(
...
 
+    cond_provider = cond_provider or model_cfg["inference"]["conditions"]
```
C-Achard: Is it always safe to fall back on falsey? Should this be strictly a None check only for the fallback, to be more explicit? Same in similar contexts

deeplabcut/pose_estimation_pytorch/apis/videos.py
```diff
@@ -490,22 +486,15 @@ def analyze_videos(
 
     snapshot = utils.get_model_snapshots(snapshot_index, loader.model_folder, loader.pose_task)[0]
 
-    # Load the BU model for the conditions provider
+    ctd_conditions = ctd_conditions or loader.model_cfg["inference"]["conditions"]
+    if ctd_conditions is not None:
+        ctd_conditions = ConditionsModelConfig.resolve_from_conditions(ctd_conditions, config=config)
```
C-Achard: The function accepts str or Path, but then the built ConditionsFileConfig from str/Path is rejected by `resolve_from_conditions`.
Should the input type annotations be changed? Is it always impossible to use these?

C-Achard: To me it would make sense that analyze_videos would be allowed to use files with pre-computed pose if available though, right?

deeplabcut/pose_estimation_pytorch/config/ctd_conditions.py
```diff
@@ -0,0 +1,230 @@
...
+
+        if isinstance(conditions, ConditionsFileConfig):
+            raise ValueError(
+                "A file config cannot be used for live BU inference. "
+                "Provide 'config_path'+'snapshot_path' or a shuffle shorthand."
+            )
+        if isinstance(conditions, cls):
+            return conditions
+
+        assert isinstance(conditions, ConditionsShuffleConfig)
```
C-Achard: Maybe best to follow the previous checks format?
```suggestion
        if not isinstance(conditions, ConditionsShuffleConfig):
            raise RuntimeError("Shuffle configs cannot be used for live BU inference")
```

deeplabcut/pose_estimation_pytorch/config/ctd_conditions.py
```diff
@@ -0,0 +1,230 @@
...
+
+    Use ``ConditionsConfig.build()`` to normalise any raw input into a typed subclass,
+
+    Subclasses:
+        - ``ConditionsFileConfig``    — pre-computed predictions file
+        - ``ConditionsModelConfig``   — resolved BU model (config + snapshot paths)
+        - ``ConditionsShuffleConfig`` — unresolved shuffle shorthand
+    """
+
+    source: Literal["file", "model"]
```
C-Achard: Wouuldn't `"shuffle"` be valid here?

C-Achard (commented):
I think this is a very good start, and indeed good to harden this a bit.

There are still a few things that may be worth clarifying:
- Should we wrap the old CondFromModel with a deprecated constructor that routes to the new builder? Or not worth it as internal API/too much effort for the payoff
- `resolve_from_conditions` is very nice to expose capabilities from the schema. It may be worth making the supported capabilities of each condition source more explicit, either in the documentation or through routing in the code, to make future refactors easier

Adding tests (which might already be done or planned) would be very helpful too.

deeplabcut/pose_estimation_pytorch/apis/analyze_images.py
deruyter92: good point, cleaner to move inside the conditional block (did that now)

deeplabcut/pose_estimation_pytorch/apis/analyze_images.py
deruyter92: good point, cleaner to move inside the conditional block (did that now)

deeplabcut/pose_estimation_pytorch/apis/analyze_images.py
deruyter92: Good point. Since the codebase still uses "" a lot as equivalent to None, I though maybe fallback all falsey values. But I think it is bad practice to silently fallback "", {}, etc and we should not keep doing it. This is just invalid input. I changed it to a proper `if cond_provider is None` check now.

deeplabcut/pose_estimation_pytorch/apis/videos.py
deruyter92: I agree that this design could be clearer. At the moment how it works is that a ctd-runner is spawned that always allows for live inference and caches the BU conditions on disk. If the cached conditions (same scorer) already exist on disk, these are used instead of new BU predictions.

If we would accept ConditionsFileConfig here, this means we cannot create a live buctd-runner and we would have add an "offline mode" more explicitly that skips this construction. Not sure if it will be worth the effort.

deruyter92: And good first point as well. This should not accept str | path in the first place, as it will never pass downstream validation. I removed it from the function signature.

deeplabcut/pose_estimation_pytorch/apis/videos.py
deruyter92: yes, applied that now. Thanks!

deeplabcut/pose_estimation_pytorch/config/ctd_conditions.py
deruyter92: good catch

deeplabcut/pose_estimation_pytorch/apis/videos.py
C-Achard: > If we would accept ConditionsFileConfig here, this means we cannot create a live buctd-runner and we would have add an "offline mode" more explicitly that skips this construction. Not sure if it will be worth the effort.

Okay so it's just not implemented. Thanks for the clarification! Fine to keep as is I would say then; not really within this PR's scope

C-Achard:
@deruyter92 I'm fixing the fast lane in another PR, sorry for the CI fails

deruyter92:
@C-Achard thanks again for your initial review. I've addressed your comments and made some extra adjustments:
- added tests
- improved public API that accepts raw input forms for `ConditionsModelConfig` / `ConditionsShuffleConfig` and its raw form (dict)
- improved docstrings and documentation for the distinction between `ConditionsFileConfig` (only used for evaluation) and `ConditionsModelConfig` / `ConditionsShuffleConfig` (used for live-inference; caches results in a file as well)


We should probably discuss whether it is worth to add the option of using precomputed conditions via `ConditionsFileConfig` / `CondFromFile` for inference as well. I completely agree with your earlier comment that this would feel natural, but I think it requires a substantial adjustment to the current implementation: now inference always assumes that unseen images can be passed and conditions are computed on the fly using the BU model (no-precomputed conditions available). The conditions are however cached, so the next time inference is run, the BU model will be skipped for existing images with the same scorer. 

Let me know if you think this adjustment would still be worth it, or if you think that caching is sufficient for now.

C-Achard (approved):
Thanks for all the updates!

C-Achard:
> I think it requires a substantial adjustment to the current implementation: now inference always assumes that unseen images can be passed and conditions are computed on the fly using the BU model (no-precomputed conditions available)

Let's leave it out then

MMathisLab (approved):

AlexEMG (approved):

PR #3390, opened 2026-07-11
feat: Adding exported value to extract_frames

### LINKED ISSUE #3188: Return frames2pick from extract_videos
A note at the top - I am happy to do the PR for this, I just wanted to verify it was even something the project wanted before dropping a PR.

Hello! I maintain a sister project (https://github.com/sam-delap/deepxromm) for integrating DeepLabCut and XMAlab. Currently, I'm implementing my own random shuffling algorithm for frame selection, however I'd like to make better use of the existing deeplabcut.extract_frames toolkit.

However, in order to then parse the subset of data I need in XMAlab format and transform it into DeepLabCut format to create my training dataset, I either have to:

- List the contents of each video directory and parse frame numbers from the file names.
- Use the frameselectiontools util function directly and re-implement error handling and image extraction.

Modify the function definition of extract_frames to add both input type hints and a return type (would still be list[int] or None given the other branches of the function) so that if frames are successfully picked using an algorithm without user input, the frames selected are returned to the caller.

sam-delap:
## Clearly document what changed and why

Added a consistent output type (and additional type hinting) to extract_frames function

## Link any related issue(s)

Closes https://github.com/DeepLabCut/DeepLabCut/issues/3188

## Update docstrings and documentation when behavior changes

Docstrings updated, from what I remember the API auto-generates the upstream docs here so no additional docs changes need to be made

## Add or update tests when appropriate

extract_frames was not directly under test in any test suite components at present

## Include a small usage example when it helps reviewers understand and/or test the change
```python
import deeplabcut

# extract_frames now RETURNS a mapping of video path -> picked frame indices
frames_by_video = deeplabcut.extract_frames(
    "/analysis/project/reaching-task/config.yaml",
    mode="automatic",
    algo="kmeans",
    userfeedback=False,
)
for video, indices in frames_by_video.items():
    print(video, "->", len(indices), "frames:", indices)
```


deeplabcut/generate_training_dataset/frame_extraction.py
```diff
@@ -62,20 +63,20 @@ def select_cropping_area(config, videos=None):
...
-    cluster_step=1,
-    cluster_resizewidth=30,
-    cluster_color=False,
-    opencv=True,
-    slider_width=25,
-    config3d=None,
-    extracted_cam=0,
-    videos_list=None,
-):
+    config: str,
```
C-Achard: Thanks for adding!

deeplabcut/generate_training_dataset/frame_extraction.py
```diff
@@ -173,7 +174,11 @@ def extract_frames(
 
     Returns
```
C-Achard, line 181: We may want to make returning the dict optional, so as to retain previous behavior where the success was reported (though the docstring return type was incorrect)

deeplabcut/generate_training_dataset/frame_extraction.py
```diff
@@ -441,7 +451,7 @@ def extract_frames(
 
         if all(has_failed):
             print("Frame extraction failed. Video files must be corrupted.")
-            return has_failed
+            return frames_by_video
```
C-Achard, line 458: This will likely have to change slightly to retain backwards compatibility.

I would suggest we make the base return type bool/None to get granularity (failed/success with result/success with no result), and add a flag to request the frames dict as output in the signature, so it preserves the old behavior as default.

Or we can make an extra function and rely on a private underlying implementation to provide both without making the return type conditional.

C-Achard (commented):
Hello @sam-delap,

Thanks a lot for the contribution, we might adapt it a bit to retain full backwards compatibility, no need to implement the changes yourself, we are happy to handle it.
We will discuss this internally and take care of the PR once we agree on the best implementation.

Best,
Cyril

deruyter92:
Merged with main to resolve some conflicts regarding the `str` / `Path` type hints

deeplabcut/generate_training_dataset/frame_extraction.py
sam-delap: I'm OK gating this behind a flag... will update on my side accordingly

deeplabcut/generate_training_dataset/frame_extraction.py
sam-delap: Will make sure I update the docstring, apologies

deruyter92:
Hi @sam-delap thanks again for contributing. I am putting this PR in draft-status for now, as you are working on it. Let us know when you are finished adding your changes, then we will help out as well to get this in a production-ready state! 💪

PR #3394, opened 2026-07-13
Matplotlib v3.9 compatibility

C-Achard:
## Scope

Migrates off `plt.cm.get_cmap`, which matplotlib removed in 3.9.

## Changes

- **`plt.cm.get_cmap` → `plt.get_cmap`** at the 8 call sites. Only change the 3.9 removal actually required (#2583, #2581).
- **`matplotlib_utils.silence_axes_logger()`** replaces `from matplotlib.axes._axes import _log` in the two 3D modules. Resolves the same logger by name instead of importing a private symbol
- **`erase_artists` / `clean_collections`** snapshot the `Axes` children sublists explicitly

## Notes

Earlier revisions of this PR added a `matplotlib_compat` layer. `plt.get_cmap` is undeprecated in 3.9, 3.10 and on `main`, and `ArtistList` concatenation was never deprecated, so there was no compatibility problem for a layer to solve.

`deprecated(name=, stacklevel=)` now has no in-repo caller. Kept as useful and tested, but happy to drop it.


deruyter92:
@C-Achard, great efforts, good plan. When working on this PR myself as well, I realized that adding a whole compat layer might not be necessary if the single solution `plt.cm.get_cmap` -> `plt.get_cmap` suffices (works for the full matplotlib quiver). The rest is anticipatory code which we currently may not need. But let's discus IRL, both have pro's and con's.

it might be helpful to merge #3430 first!

deruyter92:
(pushed my additional changes now but feel free to revert)

C-Achard:
> adding a whole compat layer might not be necessary

There was definitely some future-proofing/extra caution in doing it this way, as I was concerned we may uncover other matplotlib-related issues that would arise/need to be added later. Hopefully the current layer is not too bloated/redundant, we can definitely trim if you think there is no need for it. I do agree it should be "disposable" ideally and removed later, so that would add some weight towards making it leaner, but at the same time it's already mostly there.

Perhaps, on the other hand, would you see any extensions to it that could be needed, or is it also clear for you that only the cmap getter is needed ? If yes then I would definitely not shrink it. If no we can remove some.

Let me know what you thinkl!

C-Achard:
I agree with all latest changes, thanks a lot for the testing additions

deruyter92:
Let's merge https://github.com/DeepLabCut/DeepLabCut/pull/3430 first an then revisit this PR

C-Achard:
@deruyter92 Hopefully this version addresses everything related to the version pin, let me know what you think

deruyter92 (commented):
@C-Achard, thanks for the work on this and also looking into the extra cases (e.g. artists)

Sorry to push back a bit again: but do we really need a compatibility layer? As mentioned before I think it would be better to just do the migration now instead of adding a migration layer that we currently do not need. 

i.e. if  `plt.cm.get_cmap -> plt.get_cmap` fixes the current breakage + we add a useful `silence_axes_logger` I think we are more than set for releasing the upper-bound right? 

regarding `remove_artists`, to me it seems that there is really no compatibility requirement. But maybe I am overlooking things. Even then, I would prefer just adjusting the usage to the recommended pattern instead of adding helpers that redo that logic.

Let me know if you honestly think that it is really needed. Otherwise it's just extra code to maintain (and it pretends there is a compatibility problem when the real fix is already the upstream-supported drop-in).

deruyter92 (approved):

PR #3395, opened 2026-07-14
GUI: Fix redundant signal calls, streamline path selections

C-Achard:
## Scope

Improve GUI file selection and signal blocking, while preserving path types and workflows.

## Changes

- Added binding-safe `QFileDialog` helpers with consistent `Path`/string conversion
- Correctly handles single-file and multi-file dialog return values
- Normalized selected video, snapshot, and conditions paths
- Prevented feedback loops when synchronizing video type and shuffle widgets via signal blocking
- Made GUI handling robust to temporarily unavailable pose configuration
- Improved widget typing, editor lifetime, and noisy state-change logging

## Goal

File selection now works more reliably, selected paths remain consistent throughout the GUI, and synchronized controls no longer trigger repeated clearing, slow updates, or state sync exceptions


deruyter92 (approved):
Solid, well-scoped PR, thanks! 

I've added one commit to change `str()` conversion to `os.fspath` which is a bit more type-safe and consistent with the other places. 

LGTM

C-Achard:
> [!CAUTION]
> ~~Requires rebasing as the base was squashed~~

C-Achard:
> Solid, well-scoped PR, thanks!
> 
> I've added one commit to change `str()` conversion to `os.fspath` which is a bit more type-safe and consistent with the other places.
> 
> LGTM

Sorry, wouldn't Path usage make more sense given recent efforts in e.g. #3350 ? Should I adapt this file?

deruyter92:
>Sorry, wouldn't Path usage make more sense [...]?

`QFileDialog` expects a string, so at this boundary I think the conversion is warranted (which you did correctly). 

Regarding `os.fspath()` vs `str()` for the conversion: the first one is theoretically more robust for edge case conversions and fails for None instead of converting to a string `"None"` . However, these benefits are purely theoretical and in practice I think in hindsight, actually conversion via `str()` is clearer. Let's try to consistently use that everywhere.

I'm reverting my commit.

deruyter92 (approved):
Looks good. Ready to merge I think!

MMathisLab (approved):

AlexEMG (approved):

PR #3397, opened 2026-07-14
GUI: Config validation errors handler dialog & layout fixes

### LINKED ISSUE #3393: GUI: Wrap config validation errors in error message and avoid partial UI init

When opening an invalid config file, the schemas correctly report any errors.
However, the resulting exception arises midway through GUI initialization, resulting in rogue widgets and making the UI completely unresponsive.

A message box showing the error, and no loose widgets.

C-Achard:
Adds a more user-friendly error handling dialog for project configuration loading in the DeepLabCut GUI. 

- Reporting and recovery options for configuration validation errors, such as invalid files, missing files, or permission issues
- Adds guidance through resolving configuration problems, including options to edit the configuration, retry loading, or cancel. 
- The changes also improve the handling of recent projects and ensure the UI remains consistent even after failed loads, avoiding partial UI states and rogue widgets
- Brings several other GUI usability improvements:
  - Made tab layout scrollable to avoid collapsing widgets when the window is resized
  - Fixed the disabled Maximize action by uncapping window max size (which is monitor dependent and could break on multiple displays)
  - Fixed a warning in Analyze videos tab going out of bounds
  - Wrap video selection widget to avoid making fields unreadable

Closes #3393.

### TODO

- [x] Fix config edit menu
- [x] Merge #3399 in this PR
- [x] Testing (add or rely on tests in #3399)


deeplabcut/gui/dialogs/config_errors.py
deruyter92, line 1: Great approach for the error logging. I looked at it and it basically corresponds entirely how Pydantic recommends handling this.

deeplabcut/gui/dialogs/config_errors.py
```diff
@@ -0,0 +1,122 @@
...
+            error_type = detail.get("type")
+            message = detail.get(
+                "msg",
+                "Invalid value",
+            )
+
+            if error_type == "extra_forbidden":
+                message = "This setting is not supported by the installed DeepLabCut version."
+            elif error_type == "missing":
+                message = "This required setting is missing."
```
deruyter92: very minor - maybe this can be centralized in a mapping (e.g. constant at the top).

deeplabcut/gui/components.py
deruyter92, line 1: Thanks for fixing this at the same time as well!

deeplabcut/gui/window.py
```diff
@@ -184,11 +204,9 @@ def add_recent_filename(self, filename):
...
     @property
     def cfg(self):
-        try:
-            cfg = auxiliaryfunctions.read_config(self.config)
-        except TypeError:
-            cfg = {}
-        return cfg
+        if self.config is None:
+            return {}
+        return auxiliaryfunctions.read_config(self.config)
```
deruyter92: I know this was already the case before this PR, but I would strongly argue against returning an empty dict here. 

- This return type inconsistency between ProjectConfig and dict forces us to always use dict-access which is supported by both (unwanted).
- An empty dict conflates "no config configured" with "here is an object that still works (a bit)"
- The empty dict still does not work for direct fallback access, e.g. `self.cfg["bodyparts"]`, so it offers not more than `None` in this sense.

deruyter92: Also:  the naming of these variables was changed a bit in the [Pathlib PR (#3350)](https://github.com/DeepLabCut/DeepLabCut/pull/3350):
`self.config` -> `self.config_path`

I will make sure to handle the merging well

deeplabcut/gui/window.py
```diff
@@ -855,16 +888,193 @@ def _create_project(self):
...
+                return True
+
+            except (
+                ValidationError,
+                FileNotFoundError,
+                PermissionError,
+                OSError,
+                TypeError,
+                ValueError,
+            ) as error:
```
deruyter92: this seems currently duplicated with `load_config`. Probably good to put in a separate helper.

deeplabcut/gui/window.py
```diff
@@ -855,16 +888,193 @@ def _create_project(self):
...
+                            f"Open it manually:\n{self.config}"
+                        ),
+                    )
+
+    def _build_project_ui_from_current_config(self) -> bool:
+        """Build project tabs using the current config file on disk.
+
+        The configuration is reread for every attempt. No validated
+        configuration state is cached between attempts.
+        """
```
deruyter92, line 1130: I understand why it is necessary to read from disk after the config is adjusted because it was invalid. But is rereading the config on every property access not a bit excessive? 

Especially if manipulations (e.g. fixes) are happened in between, different parts of one operation may observe different configs.

deruyter92 (approved):
Some general comments / questions. To help implementing the suggestions, I've also created a small top-on PR #3399 with additions to this PR. Please review critically and only take what you need. I can also take care of cherry-picking/merging after you've looked at it.


**revisit extra='forbid' (in a new PR)?**
`extra="forbid"` makes config loading in general a hard-fail on an unknown field. I think this is good in principle, but it can become painful for forward compatibility (opening a project touched by a newer DLC). Maybe in a separate PR we could add an opt-in tolerant reader that just warns instead of raises (which can be used for the GUI as well). 

**`self.cfg` **
I've added some additions in the follow-up PR to improve the loading for `self.cfg`: consistently return a `ProjectConfig`, not (an empty) `dict`, use caching when possible instead of loading on every access.

deeplabcut/gui/window.py
C-Achard: @deruyter92 Agreed that re-reading on every property access is inefficient and can let different parts of one operation observe different file contents. However, do we really want to change the MainWindow.cfg into a persistent cache ? Because that changes its existing behavior and makes reload dependent on filesystem notifications.

Maybe reading and caching once at explicit operation boundaries, such as UI construction, reload, training, evaluation, and analysis would best, and cleanest overall in regards to all our other efforts (as well as a standard solution). Or we can use a scoped UI-build snapshot that is cleared after add_tabs() completes.

I think the goal really is consistency within each operation while avoiding stale global state and I'm not sure either approach works perfectly to that end right now.

deruyter92: understand your concern, but I think that is just a matter of invalidating the cache when needed. The problem with your proposed alternative is that properties like `all_bodyparts`, `is_multianimal` and `project_folder` are read in quite some different places, _after_ `add_tabs()` is completed. Each re-reading the config from disk (with possible ad-hoc config repairs that we inherited from legacy config readers.  Worst case: different configs for different GUI components). 

With the current cached config solution (introduced via #3399), no re-reads are required when the config was not changed. The cache is always invalidated when any project switch happens or the `config_path` is re-assigned, when `_build_ui_from_current_config` is called, and whenever `load_config` is triggered. I think these are clear boundaries where config re-reading is expected. The file monitor is just a bonus which is unrelated: whenever the user changes the file on disk, offer a GUI-notification + button for reloading the config from disk.

I am leaving this comment open so we can always return to it later, even after merging.

MMathisLab (approved):

PR #3402, opened 2026-07-15
Fix CondFromModel loading the config path as its snapshot

Denny-Hwang:
CondFromModel.__init__ set snapshot_path = Path(config_path), a copy-paste error that overwrote the snapshot path with the config path. When BU conditions were configured the documented way (inference.conditions.config_path + snapshot_path), the model tried to torch.load() the pytorch_config.yaml instead of the .pt snapshot and failed. Use snapshot_path for the snapshot.


deruyter92:
correct fix, but already addressed in #3387. Closing this in favor of that PR. Thanks for the contribution anyway.

PR #3410, opened 2026-07-15
Fix destfolder leaking across videos in filter/skeleton loops

Denny-Hwang:
filterpredictions() and analyzeskeleton() reassigned the destfolder parameter inside their 'for video in Videos' loop when it was None. After the first iteration destfolder stayed pinned to the first video's folder, so videos located in a different folder were looked up in the wrong place: load_analyzed_data raised FileNotFoundError and the video was silently reported as not analyzed (and any output would go to the wrong folder).

Use a per-iteration local 'videofolder' instead of mutating the parameter, matching plot_trajectories and extract_outlier_frames.


deruyter92:
Ignore the reverted merge with main, I just wanted to check pathlib compatibility, but restored to the original state for easier merging with the neighboring PRs

deruyter92 (approved):

Issue #3362, opened 2026-06-10
Migration to Pandas 3.0


deruyter92:
Tracking issue for migration to Pandas 3.0

Pandas released version 3.0 in Jan 2026 which introduces breaking changes for our codebase. It would be useful to address the places where it currently breaks, so we can remove the upperbound version we introduced in https://github.com/DeepLabCut/DeepLabCut/pull/3186. Even if we eventually drop pandas altogether this is probably a worthwhile effort.

Migration is prepared in separate PRs and will require pre-emptive regression testing before final removal of the upperbound. 
Some examples of the required changes are listed here. For a full overview see the Pandas [migration guide](https://pandas.pydata.org/docs/dev/user_guide/migration.html).

**Copy on write** (CoW)
This is arguably one of the most important ones as it removes the possibility of mutating parent dataframes via assignments in their subset views (e.g. chained `.loc[..]` assignment), for more predictable behavior. The worst-case to avoid is that data manipulations in our codebase might silently take no effect on the intended parent DataFrame. 

**String extension**
Pandas 3.0 introduces a new default `str` datatype that is more efficient than `object`. Unfortunately this default datatype does not work well with our multi-index HDFs: by default the index/column levels become the new `str` extension dtype instead of `object` and PyTables (`to_hdf`) cannot serialize a MultiIndex whose levels use extension dtypes:
```
NotImplementedError: Saving a MultiIndex with an extension dtype is not supported.
```
This would require us to change all dataframes back to `object` before saving. 

**PRs**
- [ ] https://github.com/DeepLabCut/DeepLabCut/pull/3185
- [ ] https://github.com/DeepLabCut/DeepLabCut/pull/3186
- [ ] https://github.com/DeepLabCut/DeepLabCut/pull/3360

PR #3416, opened 2026-07-19
Guard CoW read-only .to_numpy()/.values mutations under pandas 3

AxelNoun:
# Guard in-place mutations of `.to_numpy()` / `.values` results (Copy-on-Write, pandas 3.0)

Follow-up to #3360. Related to the tracking issue #3362.

## Summary

Under pandas 3.0's Copy-on-Write, `.to_numpy()` and `.values` return a
**read-only** array for single-dtype selections; mutating that array in place
raises `ValueError: assignment destination is read-only`. This PR guards the five
sites where such an array is mutated, using `to_numpy(copy=True)`.

The `pandas<3` upper bound is **intentionally left unchanged** â€” this is a
forward-compatibility fix, not the pin removal discussed in #3362. It complements
the preparatory work already merged in #3360 and is safe on the currently pinned
pandas 2.x.

## The pattern

```python
arr = df.to_numpy()      # or df.values  â†’ read-only view under CoW (single dtype)
arr[mask] = value        # ValueError: assignment destination is read-only
```

`copy=True` returns a writable array and is behavior-preserving on pandas 2.x, so
the change is correct under both major versions.

## Changes

`to_numpy(copy=True)` at the sites that mutate the extracted array in place:

- `pose_estimation_pytorch/data/ctd.py`
- `pose_estimation_3d/triangulation.py`
- `pose_estimation_3d/plotting3D.py`
- `refine_training_dataset/tracklets.py`
- `post_processing/filtering.py` â€” spline-filter branch (same interpolation
  pattern as `tracklets.py`; only the `tracklets` copy is covered by the current
  tests, so this one was located by a static scan rather than a failing test)

Considered and **left unchanged**: `pose_estimation_tensorflow/core/evaluate_multianimal.py`
extracts a mixed-dtype selection (`["sample", "y", "x", "bodyparts"]`), which
yields a writable object array and is therefore not affected.

## Testing

- Locally against **pandas 3.0.3 + PyTables 3.11.1**: the CTD-HDF tests pass after
  the fix, and the spline-filter path is smoke-tested (writable arrays, correct
  gap-filling).
- CI continues to run on pandas 2.x (unchanged pin), where these changes are
  behavior-preserving.

## Notes for #3362

- This is a class of pandas-3 break not covered by #3360 â€” it only surfaces at
  runtime under pandas 3.0, so it isn't caught by the 2.3 `future`-mode tooling.
- While validating against **released** pandas 3.0.3, the HDF concern described in
  #3362 (*"PyTables cannot serialize a MultiIndex whose levels use extension
  dtypes â†’ change all dataframes back to `object` before saving"*) does **not**
  reproduce: pandas 3.0 explicitly whitelists `StringDtype` in that guard
  (`io/pytables.py`, `write_multi_index`:
  `isinstance(lev.dtype, ExtensionDtype) and not isinstance(lev.dtype, StringDtype)`),
  and DeepLabCut's string index/column levels serialize fine (verified for the
  `df_with_missing` / `tracks` / `predictions` / `ass` keys). So no `to_hdf`
  object-conversion wrapper appears necessary on the released 3.0. (The pandas 2.3
  `future.infer_string` mode *does* raise there, since the exemption only lands in
  3.0 â€” so that mode shouldn't be used to validate HDF writes.)


AxelNoun:
Thanks for the review ? agreed on the remaining `load_tracklets_from_hdf` site.

Pushed a follow-up that guards the three remaining CoW-sensitive mutations we could reproduce under pandas 3:

1. **`refine_training_dataset/tracklets.py`** (the site you flagged): `self.data = df.values.reshape(...).swapaxes(...)` ? `df.to_numpy(copy=True)...`. `self.xy` / `self.prob` are views into `self.data` and are mutated in place (e.g. `swap_tracklets` / refine GUI). Verified end-to-end against a real HDF: `ValueError: assignment destination is read-only` before the fix, OK after.

2. **`utils/make_labeled_video.py`**: `xyp = df.values.reshape(...)` ? `df.to_numpy(copy=True)...`. `coords` is a view into `xyp` and is masked in place in **two** places (first-frame branch and per-frame loop); the single `copy=True` covers both.

3. **`pose_estimation_tensorflow/.../pose_multianimal_imgaug.py`**: `data.to_numpy()` ? `to_numpy(copy=True)`. A single-row float Series returns a read-only view under pandas 3 CoW; `kpts` is then masked in place when `mask_kpts_below_thresh=True`. Reproducible; included because it is the same failure mode as the rest of this PR.

Also ran `ruff format .` so the tree passes `ruff format --check` (the earlier CoW edits in `plotting3D.py` / `triangulation.py` / `ctd.py` needed reformatting for CI).

### Considered, left unchanged
- **`evaluate_multianimal.py:331`**: under pandas 3.0.2 the `.loc[:, [list]]` reordering path returns a writable array (consolidation/copy). Not reproducible ? no change. Adding `copy=True` would be defensive only.
- Other candidates from the AST audit (`calc_bboxes_from_keypoints`, `np.nan_to_num(..., copy=False)`, `columnwise_spline_interp` and callers) either create fresh writable arrays or never mutate their input.

### Out of scope (optional follow-up)
In `load_tracklets_from_hdf`, just above the fixed line, there is a redundant round-trip: a NumPy `data` array is re-wrapped as `pd.DataFrame(...)` then immediately re-extracted via `to_numpy(copy=True)`. It could be `self.data = data.reshape(...).swapaxes(...)` directly (`data` is already writable). Left out intentionally to keep this PR a minimal CoW guard; happy to open a separate cleanup if useful.

deruyter92 (commented):
@AxelNoun 
Apologies, I forgot to respond myself before requesting a CoPilot review. Thanks for addressing CoPilots suggestions as well.

 I've reviewed your changes, and they are definitely correct and very helpful! 

Thanks for catching these exceptions for the future-tooling in *utils/pandas_future_mode.py*. When extracting numpy arrays from the DataFrame, the warning is circumvented. I'm changing the mode from `"warn"` to `True`, to fully simulate the 3.0 behavior, and adding some tests as well.

AxelNoun:
Thanks Jaap, glad it's useful! And good call flipping the mode to `True` in `warn` mode extracting a numpy array sidesteps the CoW tracking entirely (no warning, no error), so none of these would ever have surfaced in CI. Full 3.0 enforcement is the right check.
For completeness on the one `.loc[:, cols].to_numpy()` + in-place write I deliberately left untouched (`evaluate_multianimal.py:331`): it's safe because that frame is multi-block, so the column selection consolidates into a fresh writable array rather than a view. I re-checked on pandas 2.3.3 with CoW enabled and it stays writable unlike the homogeneous single-block case your `column_subset` test covers, where even a subset is a read-only view. It comes down to block layout, not the `.loc[]` syntax. Happy to add a `copy=True` there too as cheap insurance if you'd rather not depend on the frame staying multi-block.

deruyter92:
Thanks @AxelNoun, good suggestion. Even though it is currently safe, I would lean towards adding `copy=True` as an explicit insurance. Probably worth the minimal performance cost.

I've already applied it now in [7c99777](https://github.com/DeepLabCut/DeepLabCut/pull/3416/commits/7c997771a5326588363563306be2d9a9cdea5ed0).

deruyter92 (approved):

AxelNoun:
@deruyter92 OK, perfect! Thank you, if you need me to change anything, I am available, thank you for your time!

PR #3417, opened 2026-07-20
Move imgaug from core dependencies to TensorFlow extras

natsukium:
First of all, thank you for building such a wonderful library and congratulations on the v3 release rewritten on PyTorch :tada:.

While looking through the v3 dependency tree, I noticed that `imgaug` can now be dropped from the core dependencies. It has had no release since 0.4.0 back in February 2020 and is effectively unmaintained. Carrying an abandoned package in the base install is a growing burden, both for long-term maintainability and from a supply-chain security standpoint.

Every remaining `import imgaug` lives under `deeplabcut/pose_estimation_tensorflow/`, so it is only used by the older TensorFlow-based workflow. This PR therefore moves `imgaug` from the core dependencies to the TensorFlow extras (`tf`, `tf-cu11`, `tf-cu12`, `tf-latest`) and regenerates `uv.lock` to match.

I ran pytest after removing it from the core dependencies and everything passed, so this should be a safe move.


deruyter92 (approved):
Yes, this absolutely an improvement. LGTM!

AlexEMG (approved):

PR #3418, opened 2026-07-20
Fix uncaught NetworkX exceptions in TrackletStitcher.stitch() fallback

deruyter92:
## Summary

- `TrackletStitcher.stitch()` crashes with `ValueError: Could not reconstruct N tracks` when the fallback ("black magic") path encounters infeasible flow problems, discarding all stitching work and leaving users with no `.h5`/`.csv` output after a completed detection run.
- The root cause is two uncaught `networkx` exceptions inside the `except NetworkXUnfeasible` handler, which prevent `self.paths` from ever being set.

**Reported by**: [image.sc forum](https://forum.image.sc/t/runtime-error-when-analysing-videos/121893) — user analyzing 4 visually similar mice in 1-hour videos at 60fps.

## Fix

Wrap both exception sources in `try/except` (matching the existing design intent):
- `node_disjoint_paths` → catch `NetworkXNoPath`, warn, continue with any paths already found.
- Second `capacity_scaling` → catch `NetworkXUnfeasible`, warn, keep existing paths.

### DIFF
--- a/deeplabcut/refine_training_dataset/stitch.py
+++ b/deeplabcut/refine_training_dataset/stitch.py
@@ -670,12 +670,18 @@ class TrackletStitcher:
             # Preflow push seems to work slightly better than shortest
             # augmentation path..., and is more computationally efficient.
             paths = []
-            for path in nx.node_disjoint_paths(self.G, "source", "sink", preflow_push, self.n_tracks):
-                temp = set()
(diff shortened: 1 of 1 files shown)


C-Achard (approved):
Nice, thanks!

AlexEMG (approved):

PR #3419, opened 2026-07-20
Skip TensorFlow-backed tests cleanly when the tf extra is absent

natsukium:
While working on #3417, I noticed that the test suite could no longer be collected in a base install (without the `tf` extra).

`pytest.importorskip("tensorflow")` is meant to skip these tests when TensorFlow is not installed. However, in the affected modules the guard is placed after the imports of TF-backed code from `deeplabcut.pose_estimation_tensorflow`. During pytest's collection phase those imports run first, so a base install raises an import error before the guard is ever reached, making the `importorskip` effectively a no-op.

This PR does three things. It moves `pytest.importorskip("tensorflow")` above the TF-backed imports in the affected modules so they skip cleanly instead of erroring, it adds the missing guard to `test_predict_supermodel.py` which had none, and it marks the relocated imports with `# noqa: E402` since they now intentionally follow module-level code.

### DIFF
--- a/tests/test_dataset_augmentation.py
+++ b/tests/test_dataset_augmentation.py
@@ -8,16 +8,17 @@
 #
 # Licensed under GNU Lesser General Public License v3.0
 #
-import imgaug.augmenters as iaa
 import numpy as np
 import pytest
-
-from deeplabcut.pose_estimation_tensorflow.datasets import augmentation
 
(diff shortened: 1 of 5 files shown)


deruyter92 (approved):
Good PR, correct fix. LGTM!

Issue #3240, opened 2026-03-12
Feature Request: Replace Albumentations with torchvision.transforms.v2 for PyTorch pose estimation


juan-cobos:
## Is your feature request related to a problem? Please describe.

Albumentations is an external dependency that has slowed down its development (last major release 1.4.3 in late 2023).

## Describe the solution you'd like

Replace the current Albumentations-based augmentation system in the PyTorch pose estimation module (`deeplabcut/pose_estimation_pytorch/data/transforms.py`) with `torchvision.transforms.v2`.
Benefits:
- **Dependency reduction**: Fewer external dependencies
- **Active development**: torchvision.transforms.v2 is actively maintained
- **Better integration**: Native support for `tv_tensors.KeyPoints`, `tv_tensors.BoundingBoxes`
- **Performance**: Optimized for PyTorch with potential GPU acceleration

## Describe alternatives you've considered

1. **Keep Albumentations**: Continue using it, but this relies on an external library with slower development
2. **Kornia**: Another option for differentiable augmentations, but less focused on standard image transforms
3. **Custom implementation**: Implement all transforms from scratch (too much work)

## Additional context

Current implementation uses ~15+ Albumentations transforms:
- Resize, LongestMaxSize, HorizontalFlip, Affine, PadIfNeeded
- Equalize, MotionBlur, GaussNoise, CoarseDropout, ElasticTransform
- Custom transforms: HFlip (keypoint-aware), KeypointAwareCrop, KeepAspectRatioResize, etc.
Relevant files:
- `deeplabcut/pose_estimation_pytorch/data/transforms.py` - main transforms implementation
- `deeplabcut/pose_estimation_pytorch/data/preprocessor.py`
- `deeplabcut/pose_estimation_pytorch/data/dataset.py`

I'd like to contribute to developing this feature but have some questions about the desired interface:
1. **Custom transforms vs. built-ins**: Should we implement custom DLC-specific transforms (e.g., `KeypointAwareCrop` for cropping around annotated keypoints) using PyTorch, or adapt to use built-in v2 transforms?
2. **API compatibility**: Should the config format remain the same (e.g., `hflip`, `affine`, `crop_sampling`), or take advantage of v2's declarative API?
3. **Backward compatibility**: Should we maintain Albumentations as a fallback, or fully migrate to v2?

Thank you in advance! 
Juan

C-Achard:
Hello @juan-cobos, thanks for the detailed request.

[...]

PR #2827, opened 2025-01-07
Pins `numpy<2`, fix svd for `scipy>=1.11.0`

n-poulsen:
This pull request pins the numpy version required for DeepLabCut to `numpy<2.0.0` and fixes SVD computation in `Tracklet.estimate_rank` for `scipy>=1.11.0`.

## scipy SVD computation

With `scipy<1.11.0`, computation of the SVD of an all-zero matrix would be successful, returning an all-zero array for the singular values. With `scipy>=1.11.0`, this fails with a `ValueError`. Hence, we first check if the matrice is the zero matrix before computing the SVD. If it is, we return a zero-vector to match the behavior of `scipy<1.11.0`.

This can be verified with the following script:

```python
import numpy as np
from scipy.linalg.interpolative import svd

mat = np.ones((100, 100))
u, s, v = svd(mat, 10)
print(s.shape)
print(s)

mat = np.zeros((100, 100))
u, s, v = svd(mat, 10)
print(s.shape)
print(s)
```

With `scipy==1.10.1` this succeeds with the output:

```
# Installed packages
# 
# Package    Version
# ---------- -------
# numpy      1.26.4
# pip        24.2
# scipy      1.10.1
# setuptools 75.1.0
# wheel      0.44.0

(10,)
[1.00000000e+002 2.00894856e-014 9.74976675e-030 1.87711978e-044
 3.36274201e-059 7.64750134e-074 6.78987388e-089 1.85753577e-103
 5.94236148e-118 5.94807826e-133]
(10,)
[0. 0. 0. 0. 0. 0. 0. 0. 0. 0.]
```

While with `scipy==1.15.0` this fails with the output:

```
# Installed packages
# 
# Package    Version
# ---------- -------
# numpy      1.26.4
# pip        24.2
# scipy      1.15.0
# setuptools 75.1.0
# wheel      0.44.0

(10,)
[1.00000000e+002 1.97841743e-014 2.33093502e-029 3.08938377e-044
 4.24604188e-059 6.43087871e-074 1.13626229e-088 8.36470319e-104
 1.62555162e-118 3.13271339e-133]
Traceback (most recent call last):
  File "sktest.py", line 11, in <module>
    u, s, v = svd(mat, 10)
  File "/miniconda/envs/sk15/lib/python3.10/site-packages/scipy/linalg/interpolative.py", line 905, in svd
    U, S, V = _backend.iddr_asvd(A, k, rng=rng)
[...]

PR #3420, opened 2026-07-20
Prepare for NumPy 2 forward compatibility

natsukium:
Much of the scientific-Python ecosystem has moved to NumPy 2, so the `numpy<2` cap in the core dependencies increasingly conflicts with other packages sharing an environment. This PR does the groundwork for eventually lifting that cap while leaving the cap itself in place: following the review, it is a preparatory step for forward compatibility rather than a switch to NumPy 2.

Each fix is a separate commit, with the details and rationale in its message.

## What changes

- Replace APIs NumPy 2 removed: `np.trapz` → `np.trapezoid` (keeping an `np.trapz` fallback while NumPy 1 is supported), `np.product` → `np.prod`, and `np.percentile(interpolation=)` → `np.percentile(method=)`. All three raise `AttributeError`/`TypeError` on NumPy 2.
- Estimate tracklet rank with a dense SVD instead of `scipy.linalg.interpolative.svd`, which raises "array must not contain infs or NaNs" on finite input under NumPy 2, breaking any stitching that computes tracklet affinities. The replacement is exact and deterministic where the interpolative routine was randomized, and it makes the all-zero guard added in #2827 redundant.
- Compare montblanc assemblies order-independently in the assembler test. The order of equally-scored assemblies within a frame is arbitrary and not fixed by NumPy's sorts; NumPy 2 emits two of them swapped, so the comparison failed although every coordinate matched.
- Raise the `numpy` lower bound to `>=1.22.4`. `method=` only exists from NumPy 1.22, and `pandas>=2.2` already requires `>=1.22.4` on Python 3.10, so nothing that could previously be installed is affected.

The `numpy<2` upper bound is unchanged.

## Why the cap stays

The `imgaug` concern raised in review is the blocker, and a hard one: `imgaug` 0.4.0 fails at *import* under NumPy 2.

```
imgaug/imgaug.py:45:  NP_FLOAT_TYPES = set(np.sctypes["float"])
AttributeError: `np.sctypes` was removed in the NumPy 2.0 release.
```

There has been no `imgaug` release since 0.4.0 (June 2020), so this will not be fixed upstream, and `imgaug` is currently a core dependency rather than a TF extra.

Moving `imgaug` to the TF extras (#3240) is necessary but not sufficient. `imgaug` declares no `numpy` upper bound of its own, and neither do all the pinned TensorFlow versions — only TF 2.15–2.17 constrain `numpy` to `<2`:

| extra | resolved TensorFlow | caps numpy to `<2`? |
| --- | --- | --- |
| `tf` | 2.15.1 (py≤3.11) / 2.17.1 (py3.12) | yes |
| `tf-cu11` | 2.14 | no |
| `tf-cu12` | 2.18 | no |
| `tf-latest` | ≥2.18 | no |
| `apple_mchips` | 2.14 (py<3.12) / 2.17.1 (py≥3.12) | no / yes |

So when the core cap is eventually lifted, `numpy<2` should move to the TF extras alongside `imgaug`. Otherwise `[tf-cu11]`, `[tf-cu12]` and `[tf-latest]` installs would pair `imgaug` with NumPy 2 and fail at import.

The other dependencies raised in review did not reproduce as blockers (checked on NumPy 2.4.6 / Python 3.11):

- `albumentations` 1.4.3 installs and runs under NumPy 2. Every transform used in `pose_estimation_pytorch/data/transforms.py` works with keypoint and bbox targets, except `CoarseDropout` (no bbox support) and `ElasticTransform` (no keypoint support) — albumentations' own limitations, which behave identically on NumPy 1.
- `filterpy` 1.4.5's `KalmanFilter`, as used by the SORT tracker in `core/trackingutils.py`, works under NumPy 2.
- `scipy` resolves to a NumPy 2 compatible build. Its `>=1.9` floor would need revisiting only if minimum-version installs are targeted, since NumPy 2 support starts at scipy 1.13.

## Testing

The core test suite passes identically on NumPy 1.22.4 and NumPy 2.

Running the test suite against NumPy 2 in CI as a non-blocking job would keep this from regressing while the remaining blockers are resolved. Happy to add that here or in a follow-up.


pyproject.toml
```diff
@@ -41,7 +41,7 @@ dependencies = [
   "matplotlib>=3.3,<3.9,!=3.7,!=3.7.1",
   "networkx>=2.6",
   "numba>=0.54",
-  "numpy>=1.18.5,<2",
+  "numpy>=1.22.4,<3",
```
deruyter92: If we could leave the upper bound unchanged for now, I think that would be safer. (We can integrally resolve the rest of the dependencies in the meanwhile).

```suggestion
  "numpy>=1.18.5,<2",
```

deruyter92 (commented):
@natsukium, thanks a lot for your recent valuable contributions. These are definitely improving the codebase!

Regarding this PR: 
I wouldn't be in favor of removing the upper bound just yet. Instead of directly jumping to full numpy 2 support I would prefer to accept all the improvements in this PR as a *preparatory step* for forward compatibility.

The reason I foresee issues:

- `albumentations` is currently still pinned at `<=1.4.3` 
- Most TF code still depends on `imgaug` (we could move the upperbound to TF extras to solve this)
- compatiblity with other packages like `filterpy`, `scipy` need to be verified and bounds adjusted accordingly

If we can leave the upperbound as is for the current PR, I strongly approve of your changes!

deruyter92:
By the way, would you mind if I push directly to this branch? I might add one or two commits if I spot any remaining cases. 
So far, it seems quite complete!

deruyter92:
Linking related issue https://github.com/DeepLabCut/DeepLabCut/issues/3240 here (dropping `albumentations`)

pyproject.toml
natsukium: Thanks for your suggestion. I've reverted the upper bound to <2.

natsukium:
Thanks for the careful review! I couldn't reproduce numpy 2 problems with albumentations, filterpy or scipy.
That's only at the unit test level though, so it doesn't rule out issues in real training runs.

Also, would it be worth adding a non-blocking CI job that runs the suite against numpy 2?
It would give us a signal for when the pin can be dropped.

Feel free to push directly to this branch. Note that I force-pushed just now, so please pull first.

deruyter92:
Regarding your recommendation: I agree that it will be worthwhile adding a non-blocking CI test for this, good suggestion! This holds also for other dependencies as well. I'm opening a separate PR for this.

deruyter92 (approved):
@natsukium, thanks again! I've pushed just a few minor additions. (moving the `hassatr` check out of the hot path; removing a duplicate function). Let me know if you agree. 

The PR looks great and I think it would be useful to merge soon!

AlexEMG (approved):

PR #3430, opened 2026-07-29
Add CI workflow lane for dependency testing (canary lane)

deruyter92:
**Motivation**
We currently enforce version pins for quite some dependencies. This PR introduces a non-blocking test lane that can help to identify future-incompatibilities when we release these upper bounds.

**Changes**
Add add a non-blocking "canary" test lane that tests future compatibility for specific pip-overrides. Currently, we run only the pytests and the torch functional tests for just one python version (3.12) and os (ubuntu). The following dependency overrides are tested:

- `albumentations >= 2`
- `numpy >= 2`
- `pandas >= 3`
- `matplotlib >= 3.9`

**Related PRs**
- This needs to be merged first to skip TF tests: https://github.com/DeepLabCut/DeepLabCut/pull/3419 
- Merge the current PR before below PRs. These PRs are expected to change the canary test outcomes by addressing future compatibility:
  - #3420 
  - #3416
  - #3174 
  - #3394

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -126,14 +126,14 @@ jobs:
       - name: Force-install upgrade packages
         if: ${{ matrix.pip_overrides != '' }}
         shell: bash -el {0}
+        env:
+            PIP_OVERRIDES: ${{ matrix.pip_overrides }}
         run: |
-          # Split on whitespace (not commas): pip version specs legitimately
-          # contain commas to combine constraints, e.g. "numpy>=2,<3".
-          read -ra pkgs <<< "${{ matrix.pip_overrides }}"
-          for pkg in "${pkgs[@]}"; do
+            read -ra pkgs <<< "$PIP_OVERRIDES"
+            for pkg in "${pkgs[@]}"; do
             echo "Force-installing (--no-deps): $pkg"
             python -m pip install --no-cache-dir --upgrade --no-deps "$pkg"
-          done
+            done
 
       - name: Install ffmpeg (Linux/macOS)
         if: runner.os != 'Windows'
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -83,6 +83,7 @@ jobs:
             || github.event_name == 'workflow_dispatch' && inputs.concurrency_key
             || github.run_id }}-
         ${{ matrix.os }}-${{ matrix.python-version }}
+        ${{ matrix.pip_overrides && format('-{0}', matrix.pip_overrides) || '' }}
       cancel-in-progress: true
 
     strategy:
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -80,6 +80,8 @@ jobs:
     # Use a stable concurrency key only when one is explicitly provided
     # (e.g. PR/workflow_call/manual dedupe). Otherwise fall back to github.run_id
     # so pushes to main never cancel each other.
+    # Use matrix.id (not full pip_overrides) so canary cells that share os/python
+    # do not cancel each other.
     concurrency:
       group: >-
         tests-${{ github.workflow }}-
@@ -87,7 +89,7 @@ jobs:
             || github.event_name == 'workflow_dispatch' && inputs.concurrency_key
             || github.run_id }}-
         ${{ matrix.os }}-${{ matrix.python-version }}
-        ${{ (matrix.pip_overrides || inputs.pip_overrides) && format('-{0}', matrix.pip_overrides || inputs.pip_overrides) || '' }}
+        ${{ matrix.id && format('-{0}', matrix.id) || '' }}
       cancel-in-progress: true
 
     strategy:
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -22,6 +22,14 @@ on:
         required: false
         type: boolean
         default: false
+      pip_overrides:
+        required: false
+        type: string
+        default: ""
+      continue_on_error:
+        required: false
+        type: boolean
+        default: false
 
   workflow_dispatch:
     inputs:
@@ -46,6 +54,16 @@ on:
         default: "[]"
       full_suite:
         description: "Run full pytest and default functional suite"
+        required: false
+        type: boolean
+        default: false
+      pip_overrides:
+        description: "Comma-separated pip specs to force-install after normal deps (e.g. numpy>=2)"
+        required: false
+        type: string
+        default: ""
+      continue_on_error:
+        description: "Allow test steps to fail without marking the job as failed"
         required: false
         type: boolean
         default: false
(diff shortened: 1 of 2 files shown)


.github/workflows/python-package.yml
```diff
@@ -104,6 +123,18 @@ jobs:
...
+        if: ${{ matrix.pip_overrides != '' }}
+        shell: bash -el {0}
+        run: |
+          # Split on whitespace (not commas): pip version specs legitimately
+          # contain commas to combine constraints, e.g. "numpy>=2,<3".
+          read -ra pkgs <<< "${{ matrix.pip_overrides }}"
+          for pkg in "${pkgs[@]}"; do
+            echo "Force-installing (--no-deps): $pkg"
+            python -m pip install --no-cache-dir --upgrade --no-deps "$pkg"
+          done
```
C-Achard, line 142: A potentially slightly safer suggestion, as the current version may allow for arbitrary command to be passed
```suggestion
      - name: Force-install upgrade packages
        if: ${{ matrix.pip_overrides != '' }}
        shell: bash -el {0}
        env:
            PIP_OVERRIDES: ${{ matrix.pip_overrides }}
        run: |
            read -ra pkgs <<< "$PIP_OVERRIDES"
            for pkg in "${pkgs[@]}"; do
            echo "Force-installing (--no-deps): $pkg"
            python -m pip install --no-cache-dir --upgrade --no-deps "$pkg"
            done
```

.github/workflows/python-package.yml
```diff
@@ -248,3 +283,24 @@ jobs:
...
+          echo "::warning title=Non-blocking test failure::[$MATRIX_DESC] pytest=${{ steps.run_pytest.outcome }}, functional_scripts=${{ steps.run_functional_scripts.outcome }}. This job uses continue-on-error and will still report success overall, but the failure should be investigated."
+          {
+            echo "### :warning: Non-blocking test failure"
+            echo ""
+            echo "This job runs with \`continue_on_error: true\`, so it will still report as **passing** overall -- but at least one test step actually failed and should be investigated."
+            echo ""
+            echo "| Matrix | pytest | functional scripts |"
+            echo "| --- | --- | --- |"
+            echo "| $MATRIX_DESC | ${{ steps.run_pytest.outcome }} | ${{ steps.run_functional_scripts.outcome }} |"
+          } >> "$GITHUB_STEP_SUMMARY"
```
C-Achard, line 312:
```suggestion
      - name: Report continue-on-error failures
        # continue-on-error hides failures behind a small icon on the step itself and
        # reports the job as passing overall, so make failures hard to miss: emit a
        # workflow annotation and a job-summary entry whenever this happens.
        if: >-
          ${{ inputs.continue_on_error &&
              (steps.run_pytest.outcome == 'failure' || steps.run_functional_scripts.outcome == 'failure') }}
        shell: bash
        env:
            MATRIX_PIP_OVERRIDES: ${{ matrix.pip_overrides }}
        run: |
          MATRIX_DESC="${{ matrix.os }}, py${{ matrix.python-version }}, pip_overrides='$MATRIX_PIP_OVERRIDES'"
          echo "::warning title=Non-blocking test failure::[$MATRIX_DESC] pytest=${{ steps.run_pytest.outcome }}, functional_scripts=${{ steps.run_functional_scripts.outcome }}. This job uses continue-on-error and will still report success overall, but the failure should be investigated."
          {
            echo "### :warning: Non-blocking test failure"
            echo ""
            echo "This job runs with \`continue_on_error: true\`, so it will still report as **passing** overall -- but at least one test step actually failed and should be investigated."
            echo ""
            echo "| Matrix | pytest | functional scripts |"
            echo "| --- | --- | --- |"
            echo "| $MATRIX_DESC | ${{ steps.run_pytest.outcome }} | ${{ steps.run_functional_scripts.outcome }} |"
          } >> "$GITHUB_STEP_SUMMARY"
```

.github/workflows/python-package.yml
```diff
@@ -49,6 +57,16 @@ on:
         required: false
         type: boolean
         default: false
+      pip_overrides:
```
C-Achard, line 60: I think this should account for both matrix.pip_overrides and the input-defined ones, see other comments

C-Achard: e.g. via

```yaml
env:
    PIP_OVERRIDES: ${{ matrix.pip_overrides || inputs.pip_overrides || '' }}
```
in the build: section

.github/workflows/python-package.yml
```diff
@@ -104,6 +123,18 @@ jobs:
           python -m pip install dependency-groups
           python -m pip install --no-cache-dir -e ".${{ matrix.extras }}" --group dev
 
+      - name: Force-install upgrade packages
+        if: ${{ matrix.pip_overrides != '' }}
```
C-Achard: and then use
```suggestion
            if: ${{ env.PIP_OVERRIDES != '' }}
```

.github/workflows/python-package.yml
```diff
@@ -104,6 +123,18 @@ jobs:
...
+        shell: bash -el {0}
+        run: |
+          # Split on whitespace (not commas): pip version specs legitimately
+          # contain commas to combine constraints, e.g. "numpy>=2,<3".
+          read -ra pkgs <<< "${{ matrix.pip_overrides }}"
+          for pkg in "${pkgs[@]}"; do
+            echo "Force-installing (--no-deps): $pkg"
+            python -m pip install --no-cache-dir --upgrade --no-deps "$pkg"
```
C-Achard, line 141: Do we want no-deps given that users would most often update their full stack? What do you think would be most useful here?

.github/workflows/python-package.yml
```diff
@@ -65,6 +83,7 @@ jobs:
             || github.event_name == 'workflow_dispatch' && inputs.concurrency_key
             || github.run_id }}-
         ${{ matrix.os }}-${{ matrix.python-version }}
+        ${{ matrix.pip_overrides && format('-{0}', matrix.pip_overrides) || '' }}
       cancel-in-progress: true
```
C-Achard, line 93: Could we simplify the concurrency key? Or do you prefer having all overrides be part of it?

C-Achard (approved):
Very nice, super useful for the version upgrades milestone.

One higher level design question, would you be in favor of removing those tests once we successfully increment a package and keep the infra for later? Or would this be more of a permanent CI lane by design?

.github/workflows/python-package.yml
deruyter92: yes I was not sure about this one. I would still lean towards keeping `--no-deps` because it helps us isolate the current issues in specific dependencies (e.g. pandas, numpy, matplotlib, etc). 

Maybe after we finish the current round of updating our dependency stack we can change it to a more generic canary lane where the full stack is updated

.github/workflows/python-package.yml
deruyter92: Ok good suggestion I've used more simple IDs now rather than the full keys. (They cannot be simply left-out because then the runs would cancel each other)

deruyter92:
> One higher level design question, would you be in favor of removing those tests once we successfully increment a package > and keep the infra for later? Or would this be more of a permanent CI lane by design?

@C-Achard, good question. Yes, as mentioned, I think this would indeed be very helpful as a permanent CI lane after we resolved the current dependency upper bounds (i.e. the first test layer that could find out if upstream packages become incompatible with our current API).

AlexEMG (approved):

PR #3436, opened 2026-08-05
Add detector_eval_interval to train_from_coco 

juan-cobos:
Enable to select eval interval for the detector when running adaption_train inside train_from_coco. Currently, there is an eval_interval param which only applies to the pose estimation head. I patched it by adding a new parameter which allows to change eval interval for the detector too. Alternatively, you could standardize eval_interval for both, but I followed the current approach of "detector_" prefix params.

### DIFF
--- a/deeplabcut/pose_estimation_pytorch/modelzoo/train_from_coco.py
+++ b/deeplabcut/pose_estimation_pytorch/modelzoo/train_from_coco.py
@@ -37,6 +37,7 @@ def adaptation_train(
     batch_size: int = 8,
     detector_batch_size: int = 8,
     eval_interval: int | None = None,
+    detector_eval_interval: int | None = None,
     skip_detector: bool = False,
 ):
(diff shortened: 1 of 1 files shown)


deruyter92 (approved):
Seems like a good addition to me! This parameter was missing. The implementation is consistent with other parameters.

deruyter92:
Side note: 
We could even consider adding this parameter to `train.py` and `compat.py`, but not strictly necessary as the eval interval can also be set via  `pose_cfg_updates`. As the main API is being revisited anyway in #3382 and related PRs, we can keep separate from the current PR.

juan-cobos:
@deruyter92 I think I would keep this minimal. The main issue here was the default detector `eval_interval=10` is greater than the default number of epochs in `adaptation_train` (4), resulting in never saving best. 
On the side note, I would lean towards keeping it an explicit field on the config rather than adding params to the functions. From what I've inspected, it touches multiple things so I agree it should be kept in mind for the main API.

PR #3441, opened 2026-08-12
GUI fixes (config editor + reload timer)

deruyter92:
## Summary

**Fix pathlib coercion & centralize editor opening**
- Move duplicated config-editor logic from five tabs into `DefaultTab._open_config_editor()` and `open_project_config_editor()`.
- After the pathlib migration, opening Edit config.yaml in the GUI raised TypeError because DictViewer passed a Path to QLabel, which only accepts str. This PR fixes that crash and consolidates how tabs open the YAML editor.

**Deduplicate deferred reload timer** 
- `ManageProject` and `AnalyzeVideos` each owned an identical single-shot `QTimer` to defer `reload_project_config` until
  after the config editor closed. That logic now lives on `DefaultTab` and is connected only when editing the project `config.yaml` (`reload_project=True`). Editing pose/inference configs still does not schedule a reload.

### DIFF
--- a/deeplabcut/gui/components.py
+++ b/deeplabcut/gui/components.py
@@ -13,7 +13,7 @@ from __future__ import annotations
 from pathlib import Path
 
 from PySide6 import QtWidgets
-from PySide6.QtCore import Qt, Slot
+from PySide6.QtCore import Qt, QTimer, Slot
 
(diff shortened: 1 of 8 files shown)


C-Achard (approved):
Great changes, thank you!

PR #3442, opened 2026-08-12
remove RST api docs references - replace with links to API docs

deruyter92:
## Summary
After moving to `mkdocs` API docs, some old sphinx references and RST files were still in place. This PR removes the redudant RST files, and replaces the old API references with links to the API-docs.

**Edit:**
Also some dead links were updated in the main docs: apparently some pages and notebooks still referenced the old separate multi-animal and single-animal guides. These dead links are are now replaced with the correct link to the [main user guide](https://deeplabcut.github.io/DeepLabCut/docs/main-workflows/user-guide.html).


C-Achard (approved):
Thanks for catching those, apologies for not converting them last time.

Should we delete:

- docs/maDLC_UserGuide.md
- docs/standardDeepLabCut_UserGuide.md

I think we fully reworked and replaced them in main-workflows, right?

PR #3444, opened 2026-08-12
Centralize deprecation version markers in `DeprecatedSince`

deruyter92:
**Motivation**
PR #3332  introduced a deprecation system for deprecated functions and parameter aliasing. Although this significantly centralized things, the actual decorators markering the deprecations are now spread throughout the codebase and the `since` parameter is redefined on every occurrence, making mistakes easy.

This PR centralizes the administration in a single `DeprecationRound` enum, that serves as the single source of truth, and references the specific deprecation rounds and corresponding pull requests.

**Example usage**
1. Register a new deprecation round in the `DeprecationRound` enum 
```
# ‎deeplabcut/core/deprecation.py

class DeprecationRound(Enum):
    """Collection of all deprecation rounds used in the deprecation markers - single source of truth."""
    ...
    
    PATHLIB_MIGRATION = DeprecationRoundInfo(
        since="3.0.1",
        pull_request=3421,
        summary="Migrate str path handling to pathlib.Path, including grab_files_in_folder",
    )
```

2. Mark the deprecated function, specifying `deprecation_round`
```python
# deprecated function 
from deeplabcut.core.deprecation import deprecated, DeprecationRound

@deprecated(replacement="deeplabcut.collect_video_paths", deprecation_round=DeprecationRound.PATHLIB_MIGRATION)
def grab_files_in_folder(folder, ext="", relative=True):
....
```

**Changes**
- Add and wire new `DeprecationRound` enum and `DeprecationRoundInfo` schema, listing existing deprecation rounds
- Add tests for the new deprecation logic

### DIFF
--- a/deeplabcut/core/deprecation.py
+++ b/deeplabcut/core/deprecation.py
@@ -69,7 +69,7 @@ class DeprecationRoundInfo(BaseModel):
 class DeprecationRound(Enum):
     """Collection of all deprecation rounds used in the deprecation markers - single source of truth."""
 
-    PARAMETER_CONSISTENCY = DeprecationRoundInfo(
+    INIT_PARAMETER_ALIASING = DeprecationRoundInfo(
         since="3.0.0",
         pull_request=3332,
         summary=(
--- a/deeplabcut/core/deprecation.py
+++ b/deeplabcut/core/deprecation.py
@@ -14,6 +14,7 @@ import functools
 import inspect
 import warnings
 from collections.abc import Callable
+from enum import Enum, unique
 from typing import Literal, ParamSpec, TypeVar
 
 from packaging.version import InvalidVersion, Version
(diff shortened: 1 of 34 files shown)


deeplabcut/core/deprecation.py
```diff
@@ -27,6 +27,39 @@ class DLCDeprecationWarning(DeprecationWarning):
     """Project-specific deprecation warning. Helps with filtering."""
 
 
+class DeprecatedSince:
```
C-Achard: Very nice! I like relying on the PR # for traceability; would enforcing it in the definition make sense? Or should we add a bit more context (e.g. date/author ?)

Not to overdo it, just curious how we can make this decently future-proof. 

e.g.
```python
@dataclass(frozen=True)
class DeprecationRound:
since: str
reference: str
summary: str
```

C-Achard: Or maybe just keep as is and paste the full PR link.

Let me know if any of this speaks to you, current version is also perfectly fine.

deeplabcut/core/deprecation.py
```diff
@@ -91,8 +124,11 @@ def deprecated(
     Args:
         replacement: Fully-qualified name of the replacement callable, e.g.
             ``"deeplabcut.utils.auxfun_videos.list_videos_in_folder"``.
-        since: Version in which the function was deprecated.
-        removed_in: Version in which the function will be removed.
+        since: Version in which the function was deprecated. Pass a
+            ``DeprecatedSince`` constant rather than a literal, so the round can
+            be retargeted in one place.
+        removed_in: Version in which the function will be removed, likewise a
```
C-Achard: I'm not sure if using the same for both makes sense ? The same round might have distinct "since" and "removed_in" fields, meaning we would have to duplicate or add headers

deeplabcut/core/deprecation.py
C-Achard: We could also retain the round dataclass here if introduced

C-Achard, line 127: Side note, I think this allows to construct invalid states like:
```python
DeprecationInfo(
    kind="parameter",
    target="func",
    replacement="not valid",
)
DeprecationInfo(
    kind="callable",
    target="func",
    old_parameter="old",
    new_parameter="new",
)
```

This can be fixed wih e.g.:


```suggestion
@model_validator(mode="after")
def _validate_kind_fields(self) -> DeprecationInfo:
    if self.kind == "callable":
        if self.old_parameter is not None or self.new_parameter is not None:
            raise ValueError(
                "Callable deprecations cannot specify parameter names."
            )

    elif self.kind == "parameter":
        if not self.old_parameter or not self.new_parameter:
            raise ValueError(
                "Parameter deprecations require old_parameter "
                "and new_parameter."
            )
        if self.replacement is not None:
            raise ValueError(
                "Parameter deprecations cannot specify replacement."
            )

    return self
```

C-Achard (approved):
A few comments, note that the suggestions for additional  dataclasses only matter if we want some sort of fancy automated ingestion, for maintaining it is likely redundant.

Feel free to cherry pick what you like and ignore the rest!

deeplabcut/core/deprecation.py
deruyter92: Good proposal! I've now put everything in a `DeprecationRoundInfo` schema and `DeprecationRound` enum. Hope we are not overdoing it, but so far it seems super helpful and organized to centralize these cases. I think the schemas may help.

deeplabcut/core/deprecation.py
deruyter92: Good point. Now improved with the `DeprecationRoundInfo` schemas

deeplabcut/core/deprecation.py
deruyter92: good one! applied!

deruyter92:
@C-Achard  Applied all your suggestions, thanks for the review.

Note, I've rebased the PR to make it better reviewable (and fixed the commit messages).

deeplabcut/core/deprecation.py
```diff
@@ -54,11 +53,104 @@ def _parse_version(cls, value):
...
+        if self.pull_request is None:
+            return None
+        return f"https://github.com/DeepLabCut/DeepLabCut/pull/{self.pull_request}"
+
+
+@unique
+class DeprecationRound(Enum):
+    """Collection of all deprecation rounds used in the deprecation markers - single source of truth."""
+
+    PARAMETER_CONSISTENCY = DeprecationRoundInfo(
```
C-Achard: Another quick comment, we may want to think about disambiguation, I can foresee many times where we might need to deprecate for parameter consistency. Maybe some sort of year/month marker ?

deruyter92: Ok good point, we need more unambiguous names. I would argue against a year/month marker as we already have `pull_request` and `since` describing the moment when it was introduced and dates can get a bit awkward (e.g. when multiple similar migrations happen in the same month).

I've renamed all rounds to a more descriptive name that is unique to that PR. I am sure we are creative enough to find non-ambiguous names for follow-up PRs. 

`PARAMETER_CONSISTENCY` -> `INIT_PARAMETER_ALIASING` (i.e. the PR that introduced the aliasing system)
`PRETRAINED_PROJECT_MIGRATION` -> `INIT_TF_DEPRECATION` (i.e. the PR that introduced the tensorflow deprecation, which is the main motivation of deprecating this old function)

C-Achard: Perfect thanks!

PR #3449, opened 2026-08-14
Fix frame extraction failure with pathlib video selections

### LINKED ISSUE #3156: Extracting frames from videos

### Operating System

**WSL Ubuntu 24.04**

### DeepLabCut version

**DLC 3.0.0rc9**

pytorch

### DeepLabCut mode

single animal

### Device type

RTX 2080 Ti

### Bug description 🐛

Hi there. I have just started to use DLC recently and I faced a issue when I want to extract the frames from the videos. The format of videos are H264 MP4 and I get this error in terminal:

**INFO:console:Frame extraction failed. Video files must be corrupted.
INFO:console:failed to extract frames: worker is None**

I get this error on my laptop and also on my PC.

### Steps To Reproduce

_No response_

### Relevant log output

```shell

```

### Anything else?

_No response_

### Code of Conduct

### LINKED ISSUE #3448: Extract frames failing on new and old previously successful projects
I encountered a sudden new issue today where all new and old projects are failing to extract frames. 
The popup in the GUI reads "Frame extraction failed. Please check your terminal output for more information.
Terminal output:
(DEEPLABCUT) C:\Windows\System32>python -m deeplabcut
Starting GUI...
INFO:deeplabcut.gui:Use all bodyparts
INFO:deeplabcut.gui:Videos added to analyze:
{WindowsPath('C:/Users/tensor/Desktop/3label-RZ-2026-08-06/videos/4_flies.MP4')}
Current video files:
{WindowsPath('C:/Users/tensor/Desktop/3label-RZ-2026-08-06/videos/4_flies.MP4')}
WARNING: QThread::wait: Thread tried to wait on itself
WARNING:vispy:QThread::wait: Thread tried to wait on itself
INFO:deeplabcut.gui.console:Config file read successfully.
INFO:deeplabcut.gui.console:Frame extraction failed. Video files must be corrupted.

This project has worked previously with these videos so the video shouldn't be corrupted (it also is happening on all new and old projects). 
Widows updated last night and this issue started today, unsure if it is related, but the timing is odd. 

I have tried upgrading DLC but keep running into this issue. I am able to open the window to crop, but when I click crop I also get the same error.
I attempted troubleshooting but have gotten stuck due to my lack of background in python.

Thanks for your help!

One thing to update - I had DeepLabCut run the most recent update to 3.0.1 on the computer that is no longer allowing me to load projects, another lab computer has not run this update and is allowing the creation of new projects and opening old ones. Could this be a bug in the recent update?
Apologies if I am using imprecise or inaccurate terminology.

C-Achard:
### Motivation

Frame extraction could process zero videos when `videos_list` contained `Path` objects and configuration keys were strings. 
This caused valid videos to be incorrectly reported as corrupted.

Closes #3448.

### Fix

- Paths are now normalized for comparison while preserving the original config keys
- Improved the empty selection error handling
- Added regression/test coverage


tests/generate_training_dataset/test_frame_extraction.py
```diff
@@ -0,0 +1,120 @@
+import os
+from pathlib import Path
+
+import numpy as np
+import pytest
+
+from deeplabcut.generate_training_dataset.frame_extraction import _filter_config_videos, extract_frames
+from deeplabcut.utils import auxfun_videos, auxiliaryfunctions, frameselectiontools, io
```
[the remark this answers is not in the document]
C-Achard: It is technically valid but brittle, it only works because skimage.io is being re-exported in init.py by a star import...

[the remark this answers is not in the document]
C-Achard: Patching skimage.io directly

deeplabcut/generate_training_dataset/frame_extraction.py
```diff
@@ -12,6 +12,28 @@
 from pathlib import Path
 
 
+def normalize_video_path(video: str | Path) -> Path:
+    return Path(video)
```
deruyter92, line 16: To me, this helper feels like an unnecessary abstraction of inline `Path` with is perfectly readable (even shorter).

deeplabcut/generate_training_dataset/frame_extraction.py
```diff
@@ -12,6 +12,28 @@
...
+
+
+def _filter_config_videos(
+    configured_videos,
+    selected_videos,
+) -> list:
```
deruyter92, line 22: would be helpful to add type hints. Also, let's add/adjust the type hints of the `extract_frames`  `videos_list` parameter.

deeplabcut/generate_training_dataset/frame_extraction.py
```diff
@@ -12,6 +12,28 @@
...
+
+    The original config keys are returned so they remain valid for subsequent
+    config dictionary lookups.
+    """
+    configured_videos = list(configured_videos)
+
+    if selected_videos is None:
+        return configured_videos
+
+    selected = {normalize_video_path(video) for video in selected_videos}
```
deruyter92, line 33: I would prefer normalizing `videos_list` at the beginning of `extract_frames` instead of doing it inside this helper.

deruyter92:
Thanks for addressing this! And great that you added tests 💪

deruyter92:
Side note: when do we think would be a good moment to update the config schema a bit? I think we all agree that having the file paths as mapping keys in the configs has been the cause of a lot of issues.

C-Achard:
@deruyter92 Thanks for the review !

As discussed, I think we're better off implementing additional validation for e.g project layout and keypoint data.
Then we could progressively migrate things, but maybe best if we start by first having a properly versioned baseline.

C-Achard:
All comments addressed, thank you!

tests/generate_training_dataset/test_frame_extraction.py
```diff
@@ -0,0 +1,121 @@
...
+        assert isinstance(result[0], str)
+
+    def test_filter_config_videos_matches_string_to_path(self):
+        configured = [Path(r"C:\project\videos\video.mp4")]
+        selected = [r"C:\project\videos\video.mp4"]
+
+        result = _filter_config_videos(configured, selected)
+
+        assert result == configured
+        assert isinstance(result[0], Path)
```
deruyter92: This test currently fails (we don't accept str as `selected`, only Path)


```suggestion
    def test_filter_config_videos_matches_path_config_keys(self):
        configured = [Path(r"C:\project\videos\video.mp4")]
        selected = [Path(r"C:\project\videos\video.mp4")]

        result = _filter_config_videos(configured, selected)

        assert result == configured
        assert isinstance(result[0], Path)
```

deeplabcut/generate_training_dataset/frame_extraction.py
```diff
@@ -407,7 +434,11 @@ def extract_frames(
             else:  # NO!
                 has_failed.append(False)
 
-        if all(has_failed):
+        if not has_failed:
+            raise RuntimeError(
+                "No videos were processed. Check that the selected video paths match the entries in config.yaml"
+            )
```
deruyter92, line 482: Consider changing the message here. If `video_list` was given and matched nothing, this is never reached because of earlier error-handling:
```
    if videos_list is not None and not videos:
        raise ValueError(
            "None of the selected videos matched the videos in the project "
            "configuration. Selected videos may use a different path representation."
        )
```

```suggestion
        if not has_failed:
            raise RuntimeError(
                "No frames were extracted: the project configuration lists no videos, or none of them could be opened."
            )
```

deeplabcut/generate_training_dataset/frame_extraction.py
```diff
@@ -9,9 +9,28 @@
 # Licensed under GNU Lesser General Public License v3.0
 #
 
+from typing import Collection, Iterable
 from pathlib import Path
```
deruyter92:
```suggestion
from collections.abc import Collection, Iterable
from pathlib import Path
```

deeplabcut/generate_training_dataset/frame_extraction.py
```diff
@@ -229,15 +248,23 @@ def extract_frames(
...
-    else:  # filter video_list by the ones in the config file
-        videos = [v for v in cfg["video_sets"] if v in videos_list]
+    configured_videos = list(cfg.get("video_sets_original") or cfg["video_sets"])
+    videos = _filter_config_videos(configured_videos, videos_list)
+
+    if videos_list is not None and not videos:
+        raise ValueError(
+            "None of the selected videos matched the videos in the project "
+            "configuration. Selected videos may use a different path representation."
+        )
```
deruyter92: minor suggestion: could be folded into the helper to avoid duplication maybe

deeplabcut/gui/tabs/extract_frames.py
```diff
@@ -222,7 +230,11 @@ def extract_frames(self):
             cluster_color=False,
             slider_width=slider_width,
             userfeedback=False,
-            videos_list=self.video_selection_widget.files or None,
+            videos_list=(
+                [str(video) for video in self.video_selection_widget.files]
+                if self.video_selection_widget.files
+                else None
+            ),
```
deruyter92: This `str` conversion is not required, I think `Path` is valid and better.
```suggestion
            videos_list=list(self.video_selection_widget.files) or None
```

deeplabcut/generate_training_dataset/frame_extraction.py
```diff
@@ -229,15 +248,23 @@ def extract_frames(
     from skimage.util import img_as_ubyte
```
deruyter92, line 299: The docstring of `extract_frames` should be updated to reflect `Path` support I think. 

```Python
"""
        ...
        videos_list (list[str | Path], optional): Full paths to the videos to extract
            frames for. If left as ``None``, all videos specified in the config file
            will have frames extracted. Defaults to None.
"""
```

deruyter92 (commented):
In a good state but still a few small fixes required I think. Let me know if you agree.

deeplabcut/generate_training_dataset/frame_extraction.py
C-Achard: I updated most of it to make it clearer and more consistent e.g. with defaults args, let me know what you think

deeplabcut/generate_training_dataset/frame_extraction.py
deruyter92, line 63: Effectively dead code now, since `ProjectConfig` has no field `video_sets_original`

```suggestion
        videos = list(cfg.video_sets)
```

deeplabcut/gui/tabs/extract_frames.py
```diff
@@ -51,24 +51,32 @@ def select_cropping_area(config, videos=None):
...
+
+        video_sets_name = "video_sets_original" if cfg.get("video_sets_original") else "video_sets"
+        video_sets = cfg[video_sets_name]
+
+        matching_keys = [key for key in video_sets if Path(key) == Path(video)]
+
+        if not matching_keys:
+            raise KeyError(f"Video is not present in the project configuration: {video}")
+
+        video_sets[matching_keys[0]] = temp
```
deruyter92, line 80: This is still risky right? If different Path representations of the same video differ, instead of changing the value for an existing key, this would write a new key. 

Not an issue for now, since you made the callers always pass a key directly from the config. But the whole design stays a bit fragile.

deruyter92 (approved):
Good PR and I think it can be merged in the current state! It already improves the state a lot!

Not relevant to the current scope, but by design Path-matching for the video set keys stays a bit of a fragile point. We could address that in the future by making them Pydantic models or so. 

Also there is still some existing dead code here and there that we should refactor in the future. Just commented here but no need to address now.

PR #3452, opened 2026-08-14
Fix incorrect rename mapping in modelzoo download

### LINKED ISSUE #3450: superanimal_humanbody_rtmpose_x download fails
The HumanBody checkpoint downloads successfully as rtmpose-x_simcc-body7.pt, but DLC expects superanimal_humanbody_rtmpose_x.pt. 

The downloaded file is currently not being renamed because of an incorrect rename mapping, and this is reported as an error when downloading (though it it downloaded correctly).

C-Achard:
Fix SuperAnimal HumanBody checkpoint downloads when the Hugging Face filename differs from DeepLabCut’s expected filename.

Pass the expected filename directly to download_huggingface_model, allowing it to map the actual downloaded filename, such as rtmpose-x_simcc-body7.pt, to superanimal_humanbody_rtmpose_x.pt. Existing models with matching filenames remain unaffected.

Closes #3450.


deruyter92 (approved):
Should we rename on HuggingFace actually to make it more consistent?
Regardless, this would be a good fix.

C-Achard:
@deruyter92 It would be more robust yes, but then we'd have to undo the fix. Maybe I can check if the filename already matches first?

MMathisLab:
> Should we rename on HuggingFace actually to make it more consistent?
> Regardless, this would be a good fix.

Let's keep hugging face as is; renaming breaks backwards compat

PR #3453, opened 2026-08-18
Add a knowledge indexing generator for an LLM-friendly docs index

deruyter92:
**Motivation**
Recently the DeepLabCut documentation got an overhaul. This includes the recent launch of the [api-docs](https://deeplabcut.github.io/DeepLabCut/dev/latest-release/reference/deeplabcut/). However for LLM agents, searching our documentation can still be a difficult task that easily fills the context window with irrelevant information. 

The current PR adds a module `tools/knowledge_indexing` for generating a LLM-friendly index of the DeepLabCut API and user docs, and deploys it alongside the existing docs. This helps to point out AI agents where to retrieve relevant information. 

**Implementation**
- All indexing is deployed on our gh-pages website. This makes it discoverable for LLMs (without adding the contents in the TOC). 
- A single discovery point, `llms.txt` is generated at the site root following the the [llmstxt.org](https://llmstxt.org/) convention and points at the structured index at `knowledge/<version>/..` .
- The index itself uses JSONL format: `knowledge/<version>/{api,docs}.jsonl`. It contains one JSON object per record (with a `content_hash`)
- A per-version `manifest.json` and a top-level `knowledge/manifest.json` enumerates every version 
- Versioning is asymmetric: `api.jsonl` is versioned per dev-docs release, mirroring mike. The main user docs knowledge index always indexes the currently deployed user docs in `docs.jsonl` (the user docs have no versioning).
- *Dependencies*: A dedicated `knowledge-index` extra (docutils, griffelib, markdown-it-py) is added to make the CI workflow lean (no other installs needed)

**How to use**
More information is added in tools/knowledge_indexing/README.md.
To locally build the full knowledge-index and `llms.txt`, run:
```
python -m tools.knowledge_indexing
```

The main knowledge-index is always deployed to `gh-pages` together with the user docs and dev-docs, triggered by updates on the `main` branch.  To **deploy a specific version label**, just run the pre-existing manual dispatch workflow for versioned dev-docs API docs, this now automatically adds / removes the corresponding llm-index as well. 

**Status**
- [x] Add generation tool for llm-friendly indexing of the docs + api-docs
- [x] Add CI workflow to auto-generate index for each new version / each deployed api-docs


tools/knowledge_indexing/__main__.py
```diff
@@ -0,0 +1,115 @@
...
+from .docs_index import build_docs_nodes
+from .schemas import API_DIRECTORY, DOCS_DIRECTORY
+from .toc import TOC_FILE
+from .write import write_index, write_symbol_table
+
+PACKAGE = "deeplabcut"
+
+# Default output, relative to --repo and per version, so the index is a build
+# artifact of the repository it describes. `build/` is already gitignored.
+DEFAULT_OUTPUT_ROOT = Path("build/knowledge-index")
```
C-Achard, line 25: Should this go in _build with the rest of the docs? Or should it be separate?

tools/knowledge_indexing/schemas.py
```diff
@@ -0,0 +1,267 @@
...
+            "extractor_version": EXTRACTOR_VERSION,
+            "generated_by": GENERATED_BY,
+            "generated_at": self.generated_at,
+            "base_urls": self.base_urls,
+            "node_counts": {directory: len(nodes) for directory, nodes in self.nodes.items()},
+            "nodes": {
+                directory: [{"id": node.id, "file": f"{directory}/{node.filename}"} for node in nodes]
+                for directory, nodes in self.nodes.items()
+            },
+        }
```
C-Achard, line 267: "main" and other dev docs builds may still point to a version, perhaps we could add a "package_version" field that tracks from the DLC pyproject/`__version__` ?
And make "version" more explicitely relate to the dev docs versioning by changing the name, it can stay as main, v3.0.1, etc.
It might be more granular that way.

tools/knowledge_indexing/schemas.py
```diff
@@ -0,0 +1,267 @@
...
+    generated_at: str
+    nodes: dict[str, Sequence[Node]] = field(default_factory=dict)
+    revision: str = ""
+    base_urls: dict[str, str] = field(default_factory=dict)
+
+    def to_dict(self) -> dict[str, Any]:
+        return {
+            "schema_version": SCHEMA_VERSION,
+            "version": self.version,
+            "revision": self.revision or None,
```
C-Achard, line 257: Maybe use the full SHA for this kind of thing

tools/knowledge_indexing/__main__.py
```diff
@@ -0,0 +1,115 @@
...
+        ),
+    )
+    return parser.parse_args(argv)
+
+
+def _git_revision(repo: Path) -> str:
+    """Short commit hash of `repo`, or "" if it is not a git checkout."""
+    try:
+        result = subprocess.run(
+            ["git", "-C", str(repo), "rev-parse", "--short", "HEAD"],
```
C-Achard, line 66:
```suggestion
            ["git", "-C", str(repo), "rev-parse", "HEAD"],
```

tools/knowledge_indexing/schemas.py
C-Achard, line 1: A general comment is that library version versus docs could use a little more granularity and separation to ensure the source is perfectly clear always. E.g. which API revision is used vs which user docs revision

tools/knowledge_indexing/__main__.py
C-Achard, line 1: Curiosity, does Amadeus know about the current local commit and is it compared to the knowledge system's commit? How should it handle mismatches?
(This is mostly a concern for testing with local DLC versions, and likely only affects us, but I wonder)

tools/knowledge_indexing/README.md
```diff
@@ -0,0 +1,126 @@
...
+## Known gaps
+
+- **No concepts or workflows.** These need to be authored rather than derived
+  from headings.
+- **Notebooks are not indexed.** `_toc.yml` lists notebooks under `examples/`,
+  which would need a reader for `.ipynb` markdown cells.
+- **Nothing validates the index.** There is no `--check` mode, so anchors,
+  dangling references, duplicate ids and unresolvable URLs are not caught.
+- **The user docs are not versioned** upstream, so `docs-pages/` is identical
+  across index versions and only stamped with the revision it came from.
```
C-Achard, line 126: Could we make use of the frontmatter metadata for this?

tools/knowledge_indexing/README.md
```diff
@@ -0,0 +1,126 @@
...
+| `write.py` | Nodes → YAML files, symbol table and manifest |
+| `__main__.py` | CLI |
+
+## Known gaps
+
+- **No concepts or workflows.** These need to be authored rather than derived
+  from headings.
+- **Notebooks are not indexed.** `_toc.yml` lists notebooks under `examples/`,
+  which would need a reader for `.ipynb` markdown cells.
+- **Nothing validates the index.** There is no `--check` mode, so anchors,
```
C-Achard, line 123: This might be an important item. At least check for duplicates?

tools/knowledge_indexing/schemas.py
C-Achard: Should we consider file content hashes for Nodes ? For correctness checks and maybe more optimized building (mainly the former)

C-Achard (commented):
Great work so far! 
A few questions and some hopefully useful suggestions below, but nothing major besides the revision granularity

tools/knowledge_indexing/README.md
deruyter92: the problem is: we could perfectly make a versioned index for user docs (we know the commit when the index is generated), but the ACTUAL docs on gh-pages are always just the latest. 

We could choose to index historical references (the doc source files on git) but I think it would make more sense to point to the rendered live docs (which can also be opened by the user). These live docs are unfortunately not versioned, but I don't think it is a huge problem: the main user guidance doesn't change that much. The API docs are versioned anyway.

deruyter92: BTW, the `status`, `last_verified` and other metadata fields are still very useful. They are included in the indexing description so downstream LLMs can make use of it. Super helpful product of the docs audit. 

Meanwhile I've restructured the output a bit (see 83c85878860d9797f60fea51051ee744823f9d0d) such that dev API-docs are listed per version, and the main docs are explicitly marked as unversioned (only latest: "main")

tools/knowledge_indexing/schemas.py
deruyter92: Done now. Changed to `api_version_label` (i.e. the dev-docs tag) and `package_version` (the corresponding deeplabcut package version) to avoid ambiguity.

tools/knowledge_indexing/schemas.py
deruyter92: Good suggestion, added hashes now.

deruyter92: > A general comment is that library version versus docs could use a little more granularity and separation to ensure the source is perfectly clear always. E.g. which API revision is used vs which user docs revision

Dev-docs api knowledge index can now be build separately from the main docs index using `--version-label`, to build separate granular knowledge indices. For the latest main docs and the api-index tagged "main": these are build together, with a shared `revision` timestamp. This matches how the CI currently really works: deploy latest (all) or add/remove a specific version (dev-docs only).

deruyter92: hooking it up in CI now. Please let me know what you think when I'm done (if it's granular enough in practice).

tools/knowledge_indexing/__main__.py
deruyter92: Good question. I don't have a definite answer yet.

Per your suggestion I've added content hashes and the `revision` field allows to compare the specific source. When it mismatches, that means that no specific docs are available for this version (as would be the case when a human would go and look at our versioned documentation, not finding the specific version they were looking for). 

The user can either change their DLC version to one of the documented versions, or Amadeus could assume the latest *available* version and reason from there.

C-Achard: Maybe a warning and a fallback is nicer indeed?

tools/knowledge_indexing/__main__.py
```diff
@@ -0,0 +1,189 @@
...
+        return 1
+
+    for required in (repo / TOC_FILE, repo / PACKAGE):
+        if not required.exists():
+            print(f"Error: {required} not found; is --repo the repository root?", file=sys.stderr)
+            return 1
+
+    output: Path = args.output or repo / DEFAULT_OUTPUT_ROOT
+    knowledge_dir = output / KNOWLEDGE_DIR
+    api_base_url = API_BASE_URL.format(version=args.version_label)
```
C-Achard, line 190: Should the label be checked a bit more thoroughly if it's going to end up in the URL?

tools/knowledge_indexing/__main__.py
```diff
@@ -0,0 +1,189 @@
...
+        result = subprocess.run(
+            ["git", "-C", str(repo), "rev-parse", "HEAD"],
+            capture_output=True,
+            text=True,
+            timeout=10,
+            check=True,
+        )
+    except (OSError, subprocess.SubprocessError):
+        return ""
+    return result.stdout.strip()
```
C-Achard, line 121: Is this stricter in CI?

.github/workflows/deploy-knowledge-index.yml
```diff
@@ -0,0 +1,143 @@
...
+            --output _build/knowledge-index \
+            --version-label "${{ inputs.version_label }}" \
+            --revision "${{ steps.revision.outputs.sha }}" \
+            ${{ inputs.skip_api && '--skip-api' || '' }} \
+            ${{ inputs.skip_docs && '--skip-docs' || '' }}
+
+      - name: Remove deleted version
+        if: inputs.action == 'delete'
+        shell: bash
+        run: rm -rf "_build/knowledge-index/knowledge/${{ inputs.version_label }}"
```
C-Achard: It seems that since the deploy runs with keep_files this may not reliably remove the published version

.github/workflows/deploy-dev-docs-mike.yml
```diff
@@ -127,3 +127,15 @@ jobs:
             --config-file "${{ inputs.config-file }}" \
             --deploy-prefix "${{ inputs.deploy-prefix }}" \
             "${{ inputs.version_label }}"
+
+  knowledge-index:
```
C-Achard, line 159: This may need a concurrency guard shared with other workflows operating on gh-pages, as having several workflows modifying concurrently may be a problem

.github/workflows/deploy-docs.yml
```diff
@@ -52,3 +52,14 @@ jobs:
       config-file: dev-docs/mkdocs.yml
       deploy-prefix: dev
     secrets: inherit
+
+  deploy-user-docs-knowledge-index:
```
C-Achard, line 56: This may need a concurrency guard shared with other workflows operating on gh-pages, as having several workflows modifying concurrently may be a problem

.github/workflows/deploy-knowledge-index.yml
C-Achard, line 1: This may need a concurrency guard shared with other workflows operating on gh-pages, as having several workflows modifying concurrently may be a problem

.github/workflows/deploy-knowledge-index.yml
C-Achard: Also deletion deltes both api and docs JSONs, which affects both at once, whereas building seems to be independent. Maybe worth separating the two if possible

.github/workflows/deploy-knowledge-index.yml
```diff
@@ -0,0 +1,143 @@
...
+
+      - name: Install knowledge-index dependencies
+        shell: bash
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install ".[knowledge-index]"
+
+      - name: Check out existing knowledge index from gh-pages
+        uses: actions/checkout@v6
+        continue-on-error: true
```
C-Achard: Should this be more error-specific? Since checkout failing != does not exist

tools/knowledge_indexing/write.py
```diff
@@ -0,0 +1,206 @@
...
+def _write_json(path: Path, data: dict[str, Any]) -> None:
+    """Write a single, human-readable JSON object."""
+    path.write_text(json.dumps(data, ensure_ascii=False, indent=2) + "\n", encoding="utf-8")
+
+
+def _read_json(path: Path) -> dict[str, Any] | None:
+    """Read a JSON object, or None if `path` doesn't exist or isn't valid JSON."""
+    try:
+        return json.loads(path.read_text(encoding="utf-8"))
+    except (OSError, json.JSONDecodeError):
```
C-Achard: This might be a tad permissive if there is a malformed output, since None means "keep that side of the index", being unable to read an existing file makes it as if it should be ignored safely

tools/knowledge_indexing/write.py
```diff
@@ -0,0 +1,206 @@
...
+    file and manifest provenance are kept exactly as already on disk under
+    `knowledge_dir/version_label`. Returns the number of api and docs records
+    written this run (0 for an untouched half).
+    """
+    version_dir = knowledge_dir / version_label
+    version_dir.mkdir(parents=True, exist_ok=True)
+
+    existing = _read_json(version_dir / VERSION_MANIFEST)
+    api_provenance = ApiProvenance.from_dict(existing["api"]) if existing and existing.get("api") else None
+    docs_provenance = DocsProvenance.from_dict(existing["docs"]) if existing and existing.get("docs") else None
```
C-Achard, line 56: Is there a check that the files exist rather than only being listed by the manifest?

C-Achard (commented):
A few CI-related comments, really liking the new version! 
I also started on the write.py side a bit.

Were you thinking of adding tests as for the other tools as well?

.github/workflows/deploy-dev-docs-mike.yml
deruyter92: The concurrency guards are already in place for the entry points. That should be sufficient right? Individual steps are ordered.

.github/workflows/deploy-docs.yml
deruyter92: The concurrency guards are already in place for the entry points. Or am I overlooking something?

.github/workflows/deploy-knowledge-index.yml
deruyter92: Good point. actually, there is no reason it should ever continue on error. It should always exist in this stage, and a checkout failing should surface.

deruyter92:
@C-Achard, I'm implementing tests now and addressing your other review comments (most of it is implemented but I hadn't replied yet). I'll tag you for re-review when done

tools/knowledge_indexing/__main__.py
deruyter92: Done. default is now _build/knowledge-index, alongside the other docs builds

tools/knowledge_indexing/schemas.py
deruyter92: Done. We now record the full commit hash

tools/knowledge_indexing/__main__.py
deruyter92: Agreed. That belongs in Amadeus. I will think about the design in the next revision round there

tools/knowledge_indexing/README.md
deruyter92: done! Good point.

build fails if duplicate record ids appear

tools/knowledge_indexing/__main__.py
deruyter92: Done! Tnx, good point.

tools/knowledge_indexing/__main__.py
deruyter92: In CI we always pass --revision from git rev-parse (tag or HEAD), so a failed revision resolution fails the workflow rather than writing an empty one.

.github/workflows/deploy-knowledge-index.yml
deruyter92: Good catch! Now delete doesn't go through peaceiris/keep_files. It checks out gh-pages, runs --delete, and pushes the knowledge-tree update directly. And delete only removes the API index, the docs index always stay in line with the latest `main` build

.github/workflows/deploy-dev-docs-mike.yml
deruyter92: Entry points (deploy-docs, manage-dev-docs) already share concurrency: group: gh-pages-deploy. 

The reusable knowledge-index workflow is only called from those, so I think this should be fine.

.github/workflows/deploy-docs.yml
deruyter92: see prior comment. Using only as re-useable workflow now (with the same concurrency guards that are already there)

.github/workflows/deploy-knowledge-index.yml
deruyter92: Fixed now!

tools/knowledge_indexing/write.py
deruyter92: agreed. Fixed now: malformed JSON now raises instead of being treated as absent

tools/knowledge_indexing/write.py
deruyter92: Yes, the top-level manifest only lists a version if api.jsonl exists and the version manifest has an api provenance block.

deruyter92:
@C-Achard, let me know what you think. I've addressed all your comments. I am quite happy with the current version. When we actually start using it ourselves with Amadeus, we might come up with further improvements (e.g. better content summarizing from docs).

tests/tools/knowledge_indexing/conftest.py
```diff
@@ -0,0 +1,17 @@
...
+from pathlib import Path
+
+import pytest
+
+
+def _repo_root() -> Path:
+    return Path(__file__).resolve().parents[3]
+
+
+@pytest.fixture(scope="session", autouse=True)
```
C-Achard, line 13: This is imported by pytest on collection, but the body only runs once it starts setting up tests. So the imports have already executed by then.

Adding 
```toml
[tool.pytest.ini_options]
pythonpath = ["."]
```
in the pyproject would allow to skip this fixture.

C-Achard: Or moving to the general tools conftest.py

C-Achard: (Note that in CI the pythonpath is already handled by the pytest -m call)

tools/knowledge_indexing/docs_index.py
```diff
@@ -0,0 +1,426 @@
...
+
+        if heading.tag == "h1":
+            # The first H1 titles the page; its lead paragraph summarises it.
+            title = title or text
+            summary = summary or excerpt
+            continue
+
+        anchor = _anchor(text)
+        seen[anchor] += 1
+        # Sphinx registers only the first occurrence of a repeated heading, so
```
C-Achard, line 247: This assumption is worth a regression check against the real docs build, since we have a lot of duplicate headings, so if it's wrong it mislinks a lot of sections silently.
The solution below may be complex so feel free to research alternatives or ignore + document it.

Concretely, docutils' set_id uses make_id(name) for the first occurrence and falls back to auto-generated id1, id2, … on collision. Maybe write a regression check and update based on the real behavior:

- Build one minimal page with two identical headings using the real _config.yml
- Read the generated HTML ids
- Encode that observed behaviour in _anchor() / _read_structure

For the exact assertion, anchor and docs_url should ideally carry the observed published anchor. id needs to stay unique for _check_unique_ids, so derive it from the resolved anchor rather than suffixing the first occurrence's, and if the build does assign unique ids per heading, the seen Counter and the suffix are not needed anymore, since they only exist to work around the shared assumption.

One more thing that was flagged: _config.yml sets no myst_heading_anchors, which is why make_id is the right function, but that's an undocumented dependency on a jupyter-book default, and if anyone sets it later every anchor in docs.jsonl goes wrong at once. Same test covers it; a one-line comment in _anchor() naming that config would pin it.

tools/knowledge_indexing/write.py
```diff
@@ -0,0 +1,219 @@
...
+                    url=symbol.docs_url,
+                    signature=symbol.signature,
+                    summary=symbol.summary,
+                    source=symbol.source,
+                )
+            )
+    return records
+
+
+def _check_unique_ids(filename: str, ids: Iterable[str]) -> None:
```
C-Achard, line 191: Symbol and submodule IDs cannot collid, and _check_unique_ids safely detects that, but a technically valid package containing both a symbol and a same-named submodule (which is arguably bad practice but may occur) aborts the build. Should we disambiguate module and symbol IDs by type, or accept and explicitly  document this as a limitation?

tools/knowledge_indexing/write.py
```diff
@@ -0,0 +1,219 @@
...
+    return api_count, docs_count
+
+
+def write_top_manifest(knowledge_dir: Path, docs_version_label: str) -> None:
+    """Rebuild `knowledge/manifest.json` from the version directories under `knowledge_dir`."""
+    versions = sorted(child.name for child in knowledge_dir.iterdir() if child.is_dir() and _has_api(child))
+    has_docs = (knowledge_dir / docs_version_label / DOCS_FILE).is_file()
+    manifest = TopManifest(
+        docs_path=f"{docs_version_label}/{DOCS_FILE}" if has_docs else "",
+        api_latest=docs_version_label,
```
C-Achard, line 98: api_latest is always "main" here, so knowledge/manifest.json tells an agent that the latest API index is the rolling dev build, but manage-dev-docs.yml already has a notion of "latest stable": mark_latest_release passes latest-release to mike as an alias and as the /dev/ default. That alias never reaches the knowledge index as it's keyed on version_label alone, so an agent asking "what's the current release API?" gets main, whose api.jsonl URLs point at unreleased signatures.

"Known gaps" frames this as version-numbering work left for later, but the ordering problem is already solved upstream: mike is told which version is the release. Could deploy-knowledge-index.yml take aliases (as deploy-dev-docs-mike.yml already does) and record them per version, so api.latest resolves to whatever carries latest-release and falls back to main? That also makes knowledge/latest-release/... addressable without knowing the version number.

tools/knowledge_indexing/api_index.py
```diff
@@ -0,0 +1,254 @@
...
+    """
+    is_root = module.path == package
+    symbols = []
+
+    for name, member in module.members.items():
+        if name.startswith("_"):
+            continue
+
+        target, kind = member, _own_kind(member)
+        if kind not in ("function", "class"):
```
C-Achard, line 165: Does this retains methods signatures for classes?
Can we record one level into class members with kind: "method" (skipping private and inherited methods)?

C-Achard (commented):
Quite a few new finds; let me know how severe/actually import you think these are for our use

C-Achard:
> Quite a few new finds; let me know how severe/actually import you think these are for our use

Some extras:

1. **`docs_index.py:361` — `_split_frontmatter` splits on the first two `---`.**
   A `---` inside the frontmatter, or a page whose body opens with a horizontal
   rule, silently truncates everything before it and we index a partial page with
   no error. Matching a closing delimiter at line start would be safer.
2. **`write.py:106` — `shutil.rmtree(..., ignore_errors=True)`** while `_delete`
   already checks `is_dir()` first. The flag can no longer mean "absent", only
   "removal failed", and we then print "Deleted the ... API index".
3. `test_write.py` builds nodes with `SimpleNamespace`, so a field rename in
   `ApiNode` / `DocsPageNode` won't be caught. Importing the real dataclasses only
   costs the griffe/docutils import at module load.
4. The delete path in `deploy-knowledge-index.yml` installs the whole
   `knowledge-index` group, including `griffelib`, which it never imports.

tests/tools/knowledge_indexing/conftest.py
deruyter92: good point. This never ran before collection. I realized now that the same problem actually pre-existed for the test-selector, and was only working 'by coincidence' since `python -m pytest` adds cwd to path. 

See #3484 for a fix to the pre-existing problem (namely, adding the root dir via pyproject.toml). For now, I've cherry-picked the fix from there, but I'd suggest we just merge that PR before this one.

deruyter92: (also removed the dead code that I had written in the current PR) thanks for spotting

tools/knowledge_indexing/docs_index.py
deruyter92: This is an important comment. Indeed I checked and we mis-linked the duplicate headings to the first occurrence..  (and I can confirm we have a lot of duplicate headings).

I've tried a few approaches but it is difficult to reliably render the correct links from the markdown source (modelling docutils and sphinx behavior). So instead I now switched to building from the built HTML. The site is built anyway in the same workflow, so passing it as an artifact is not a big issue.

The code for extracting anchors etc is now changed a lot. And I added a regression test. See 1934cade

tests/tools/knowledge_indexing/conftest.py
deruyter92: see 48812d92 and e5643949

tools/knowledge_indexing/write.py
deruyter92: Good suggestion. Symbol IDs are now disambiguated using `#` 

see 15ebb8d2

tools/knowledge_indexing/write.py
deruyter92: good suggestion. Implemented as you proposed! see 34fde37a

tools/knowledge_indexing/api_index.py
deruyter92: Added. A class's own public methods are recorded as kind: "method" rows named Class.method, so the agent does not have to fetch our API docs to read the methods now. b3b3b740

deruyter92:
> Some extras:
> 
> 1. **`docs_index.py:361` — `_split_frontmatter` splits on the first two `---`.**
>    A `---` inside the frontmatter, or a page whose body opens with a horizontal
>    rule, silently truncates everything before it and we index a partial page with
>    no error. Matching a closing delimiter at line start would be safer.
> 2. **`write.py:106` — `shutil.rmtree(..., ignore_errors=True)`** while `_delete`
>    already checks `is_dir()` first. The flag can no longer mean "absent", only
>    "removal failed", and we then print "Deleted the ... API index".
> 3. `test_write.py` builds nodes with `SimpleNamespace`, so a field rename in
>    `ApiNode` / `DocsPageNode` won't be caught. Importing the real dataclasses only
>    costs the griffe/docutils import at module load.
> 4. The delete path in `deploy-knowledge-index.yml` installs the whole
>    `knowledge-index` group, including `griffelib`, which it never imports.

Addressed now!

1. regex matching in `_split_frontmatter`  is improved. 30cfa07a, tests in e3ae6b90
1. `shutil.rmtree` ->  `ignore_errors=True` is now dropped. Anything reaching there is a real failure and raises now. e7355d74
1. Using real dataclasses now instead of a fake replacement. They are moved to schemas.py alongside the published shapes to make importing cheap. f06cfd6e, c27ecac5
1. This is guarded by a conditional now. The delete branch installs pyyaml only, not the entire group. 9b4f2232

C-Achard:
@deruyter92 Are the CI failures related to rebasing?

C-Achard:
Let me know if you'd like me to investigate

deruyter92:
@C-Achard, the problem was that I was trying too hard to avoid installing the new knowledge-index dependencies during test time. They are actually required now after the latest additions in tests/tools. 

Meanwhile, I think I addressed all your comments. Let me know if anything else comes up!

PR #3456, opened 2026-08-19
Update command line interface

### LINKED ISSUE #3381: Update (or remove) command line interface
## Summary
`deeplabcut/cli.py` is currently stale/unmaintained/outdated and not integrated with the PyTorch-first API. The Click-based CLI exposes workflow commands (`create_new_project`, `extract_frames`, `train_network`, `analyze_videos`, etc.) that currently bypass `deeplabcut.compat` and call `pose_estimation_tensorflow` directly, so they do not follow engine selection used by the public API. 

In addition:
- `cli.py` is not really documented or tested
- `click` is not declared in pyproject.toml dependencies
- version banner prints deeplabcut v0.0., and --verbose is unused

## Proposed direction
- As our documentation primarily mentions the python API and GUI as entry points, we should consider removing cli.py. It is currently not functional and can lead to confusion. 
- We could decide (e.g. after dropping TensorFlow) to create a new command line interface that matches the PyTorch API. This does not necessarily have to use Click - there are good alternatives (e.g. [Typer](https://typer.tiangolo.com/)).

deruyter92:
**Motivation**
The old `deeplabcut/cli.py` and root `dlc.py` became stale and outdated as they have not been maintained since version 2.x. With the official deprecation of TensorFlow (see #3382), this is the right moment to re-introduce an updated, PyTorch centric CLI. 

In this PR, `deeplabcut/cli.py` is rewritten as a [Typer ](https://typer.tiangolo.com/) app whose commands are thin
wrappers around deeplabcut.*. This should facilitate easy maintenance.
Until TensorFlow is fully removed, engine routing (PyTorch vs. TensorFlow) is automatically handled by
the API layer. 

Required parameters are explicit CLI arguments; optional ones can be specified via a repeatable `--set key=value`

**Considerations:**
[Typer ](https://typer.tiangolo.com/) seems like the right trade-off between control and automation, being just a thin wrapper around existing API (less risk of drift compared to bare `click` commands). Currently also still experimenting with alternatives, such as https://cyclopts.readthedocs.io/en/latest/, which focus even more on automation (Also reads docstrings and allows Union types).

image


deeplabcut/cli.py
```diff
@@ -1,803 +1,462 @@
-#
 # DeepLabCut Toolbox (deeplabcut.org)
-# © A. & M.W. Mathis Labs
-# https://github.com/DeepLabCut/DeepLabCut
-#
-# Please see AUTHORS for contributors.
-# https://github.com/DeepLabCut/DeepLabCut/blob/master/AUTHORS
```
C-Achard: Should that part be removed?

PR #3459, opened 2026-08-21
add stub-driven lazy imports using `lazy_loader`

deruyter92:
**Summary**
Replace DeepLabCut's hand-maintained lazy export maps (_API_EXPORTS_MAP, _OPTIONAL_EXPORTS, and the __all__ export lists) with Scientific Python's `lazy_loader.attach_stub`.

`deeplabcut/__init__.pyi` is now the source of truth for public API.  Type checkers can use use it, and at runtime 
`lazy_loader` parses it to install `__getattr__`, `__dir__`, and `__all__`, as well. 

Each implementation module is only imported when its attribute is first accessed. This avoids eagerly importing heavy or optional modules such as `pose_estimation_pytorch`, `gui`, and `pose_tracking_pytorch` during `import deeplabcut`.

Note that this shifts the import-time drastically to lower-level heavy modules
import_times

**Changes**
- add `lazy_loader` as a runtime dependency,
- add `__init__.pyi` and `py.typed`
- add a type-checking tool that checks the stubs in CI see [6b3c8e5](https://github.com/DeepLabCut/DeepLabCut/pull/3459/changes/6b3c8e5de9e0758620f96f70aa81b15ca5ad73f0)
- add runtime tests that check if all exports can be resolved
- add GUI warm-up to eagerly load modules at launch (see [411237](https://github.com/DeepLabCut/DeepLabCut/pull/3459/changes/411237334701b312f5b621c07881bdbd76dbe43c))
- fix the star-all imports in deeplabcut.utils
- move the `DEBUG` constant to it's own module
- **Behavior change:** for exports whose dependencies are not installed, attribute access now raises `ModuleNotFoundError` instead of `AttributeError` which is also propagated through `hassattr`

> [!NOTE]
> **For future reference**: this PR is scoped to change only the **import strategy** while fixing some big issues like circular imports. During the work, I found that within the package a lot of import hygiene can still be applied at lower levels. For instance: many modules import heavy dependencies in _core_, such as pandas, numpy or numba. Fixing these would yield another big win.


deeplabcut/__init__.py
```diff
@@ -13,272 +13,82 @@
...
+        "refine_labels",
+        "refine_tracklets",
+        "SkeletonBuilder",
+    }
+)
 
-# -----------------------------------------------------------------------------
-# Public API
-# -----------------------------------------------------------------------------
+_TORCH_EXPORTS = frozenset({"transformer_reID"})
```
C-Achard, line 64: How do you rate this in terms of maintainability?
Could we parse from the pyproject somehow?
Or make a shared helper that is called by all e.g. GUI modules that tries to import gui deps,
and raises directly an import error asking to install the optional dep? This way each module with optional deps has clear guidance messages, we have a centralized file with all of those at a glance, and it only couples it to local source code imports without risking drift with the pyproject

deeplabcut/__init__.py
```diff
@@ -13,272 +13,82 @@
...
 
-_VERSION_EXPORTS = [
-    "__version__",
-    "VERSION",
-    "DEBUG",
-]
+_GUI_DEPENDENCY_MODULES = frozenset({"PySide6", "napari", "qdarkstyle"})
```
C-Achard, line 66: Should this be broader? What is napari-deeplabcut is missing?

tests/typing/top_level_api.py
C-Achard, line 1: Will this ever run in CI or is it meant as local only?

tests/test_top_level_api.py
```diff
@@ -0,0 +1,185 @@
...
+    def fake_lazy_getattr(name):
+        raise ModuleNotFoundError("No module named 'some_unrelated_module'", name="some_unrelated_module")
+
+    monkeypatch.setattr(deeplabcut, "_lazy_getattr", fake_lazy_getattr)
+    with pytest.raises(ModuleNotFoundError):
+        _ = deeplabcut.launch_dlc
+
+
+@pytest.mark.skipif(
+    not (_module_available("torch") and _module_available("PySide6")),
```
C-Achard, line 180: The skip condition does not represent the full optional dependency set, right? With PySide6 installed but missing another GUI dep, this would run. Can we mark this test in the CI test selector maybe? Or maybe there are other minimal solutions that don't require copying the deps from pyproject at all

deeplabcut/__init__.py
```diff
@@ -13,272 +13,82 @@
...
-    _VERSION_EXPORTS
-    + _CORE_EXPORTS
-    + _PROJECT_EXPORTS
-    + _DATASET_EXPORTS
-    + _API_EXPORTS
-    + _UTIL_EXPORTS
-    + _THREE_D_EXPORTS
-    + _OPTIONAL_API_EXPORTS
-)
+        raise
```
C-Achard, line 94: Maybe good to mention: this changes e.g. hasattr(deeplabcut, "launch_dlc") from returning False to raising.
I would assume this is fine in most cases but still a bit of a sneaky change

deeplabcut/__init__.pyi
```diff
@@ -0,0 +1,155 @@
...
+from .pose_estimation_3d.triangulation import triangulate as triangulate
+from .pose_tracking_pytorch import transformer_reID as transformer_reID
+from .utils import auxfun_videos as auxfun_videos
+from .utils import auxiliaryfunctions as auxiliaryfunctions
+from .utils.auxfun_multianimal import convert2_maDLC as convert2_maDLC
+from .utils.auxfun_videos import CropVideo as CropVideo
+from .utils.auxfun_videos import DownSampleVideo as DownSampleVideo
+from .utils.auxfun_videos import ShortenVideo as ShortenVideo
+from .utils.auxfun_videos import check_video_integrity as check_video_integrity
+from .utils.auxfun_videos import collect_video_paths as collect_video_paths
```
C-Achard, line 136: This was not present before, is adding intentional?

C-Achard: (Not in `__all__`)

C-Achard:
Note: In `deeplabcut\gui\utils.py::is_latest_deeplabcut_version`, as well as window.py, we may want to change the import to deeplabcut.version instead of top-level deeplabcut. Or is that not a concern generally?

C-Achard:
Also the import all calls in `.utils.__init__` should change, I think it's a good PR to do this; happy to make another PR on top of this one also

C-Achard:
For the GUI, could we disable lazy imports ? Having the interface freeze on imports makes it seem unresponsive, in this case the previous behavior of importing all on launch may be a bit more user-friendly

deruyter92:
@C-Achard, thanks for your review comments. 

> Note: In deeplabcut\gui\utils.py::is_latest_deeplabcut_version, as well as window.py, we may want to change the import to deeplabcut.version instead of top-level deeplabcut. Or is that not a concern generally?

With the current implementation, deeplabcut.version is eagerly bound at `__init__.py`, so this is not a concern.

> Also the import all calls in .utils.__init__ should change, I think it's a good PR to do this; happy to make another PR on top of this one also

Done, thanks

> For the GUI, could we disable lazy imports ? Having the interface freeze on imports makes it seem unresponsive, in this case the previous behavior of importing all on launch may be a bit more user-friendly

I've added a GUI warm-up now, that imports all public deeplabcut exports + heavy imports on a deamon thread.

deeplabcut/__init__.py
deruyter92: yeah good point. This was over-engineered a bit. We don't need special error handling here for GUI or Torch dependencies. Torch is a main dependency and should be treated like all others. GUI components are normally never interacted with from scripts, and even then: the raw `ModuleNotFoundError` + stack trace is informative enough. 

Nothing warranted this special error handling and it is removed now

deeplabcut/__init__.py
deruyter92: see other comment. Removed now

deeplabcut/__init__.py
deruyter92: You are right to point this out, although do you foresee any risk? It only changes for modules that _fail_ to import, namely for exports whose dependencies are not installed, attribute access now raises `ModuleNotFoundError` instead of `AttributeError`. BTW, the previous behavior also swallowed genuine import errors inside DeepLabCut, reporting them as missing attributes.

deruyter92: I've included it in the PR description now.

deeplabcut/__init__.pyi
deruyter92: It actually was. I've double checked to assert that `set(deeplabcut.__all__)` is exactly the same in the current version and `main`. 


**edit:** it was not _exactly_ the same but that is corrected:
-  `collect_video_paths` was exported but missing in `__all__` on `main`. It is now added to `__all__` as well. 
- `DEBUG` was exported in this PR but not added to `__all__` which it is now.

tools/typing/top_level_api.py
deruyter92: A few notes:
 - this fixture is now moved to tools instead of tests, together with the actual tool `test_type_stub.py` that uses it. This enables running it as a separate tool without deeplabcut or any dependencies installed.
 - it is now hooked into CI via a separate lightweight type checking test see [6b3c8e5](https://github.com/DeepLabCut/DeepLabCut/pull/3459/commits/6b3c8e5de9e0758620f96f70aa81b15ca5ad73f0)
 
 
 There are also some tests that check if the imports resolve at _runtime_ defined in _test_top_level_api.py_. See other comment below. Let me know if you think this is excessive.

tests/test_top_level_api.py
deruyter92: I am reading the GUI deps from pyproject.toml now and removed the `skipif` everything runs now, and those modules that require GUI dependencies are excepted in case these dependencies are missing.

Note that we now have _two_ runtime tests that resolve all exports (+ a type-checking tool that can run in CI to test the stubs).

let me know if you think this is excessive

deeplabcut/__init__.py
C-Achard: I don't think it will be a problem, thanks for adding it

deeplabcut/utils/__init__.py
C-Achard: This may not re-export the full utils as previously, now each submodule needs to be called.
Example:
`from deeplabcut.utils import VideoReader`

PR #3463, opened 2026-08-28
update deprecation marker versions (TF deprecation since 3.1, removed in 3.2)

deruyter92:
Placeholder PR to revisit the deprecation markers before 3.1 release. To be decided: official removal date?


deruyter92:
merging into #3382, to avoid excessive constant rebasing

C-Achard:
For the record: we decided that 3.2 could be kept as a provisional target, and shifted/updated later if needed.

PR #3466, opened 2026-08-28
Run trajectory plots after analysis completes

### LINKED ISSUE #3462: DLC Freezing Issue

### Operating System

Windows 11

### DeepLabCut version

3.0.1

pytorch

### DeepLabCut mode

single animal

### Device type

CPU

### Bug description 🐛

Hello,
We have been experiencing a problem with DeepLabCut since the most recent update. The program freezes consistently, either while analyzing a video or when trying to create a video with the predicted poses.

We have tried reinstalling DeepLabCut, but the issue continues. Before the latest update, everything was working well without these problems.

Has anyone else experienced this issue, or is there a recommended solution or workaround? 

Any help would be greatly appreciated.

Mona

### Steps To Reproduce

Open DeepLabCut 3.0.1.
Open an existing DeepLabCut project.
Select a video for analysis.
Start Analyze Videos.
DeepLabCut begins processing the video but eventually freezes and becomes unresponsive.

The same issue also occurs when attempting to create a video with the predicted poses after analysis.

We have reproduced the issue multiple times, and it occurs consistently. We also tried reinstalling DeepLabCut, but the freezing continued.

### Relevant log output

```shell

```

### Anything else?

_No response_

### Code of Conduct

C-Achard:
Store plotting options before starting the worker and defer `plot_trajectories(showfigures=True)` to a main-thread finished slot. 

The analysis pipeline now always runs with `showfigures=False`, only opens trajectory plots when analysis succeeds, and centralizes button/progress-bar reset in the completion handler.

This fixes the GUI freezing due to plots opening in the wrong thread.

Closes #3462.


deeplabcut/gui/tabs/analyze_videos.py
```diff
@@ -53,12 +53,74 @@ class AnalyzeVideos(DefaultTab):
...
+            and self._pending_plot_options.plot_trajectories
+            and self._pending_plot_options.show_trajectory_plots
+        )
+
+        self._analysis_failed = False
+
+        if should_show_plots:
+            self._pending_plot_timer.start()
+        else:
+            self._clear_pending_trajectory_plots()
```
deruyter92, line 119: One behavioral change: 
This PR moves the plotting step from *during successful batch processing* to *after the whole pipeline succeeds.* Now, a late failure such as CSV conversion can suppress plots that otherwise would have been shown.

Do yo think this is justified or should we keep the previous behavior where plots are visualized as each batch succeeds?

deruyter92 (commented):
Good PR! 

Just one remark about a behavioral change. Not sure how relevant it is, but worth having your explicit opinion.

deruyter92 (changes requested):
after our discussion, we decided that this needs further changing ideally. To not change the existing behavior.

PR #3467, opened 2026-08-28
fix faulty str / Path conversions

deruyter92:
minor fixes where str() conversion was missing


C-Achard:
Ah, glad you could reopen, sorry again

C-Achard (approved):
Thanks!

PR #3468, opened 2026-08-28
Fix latent bugs and misleading errors in `create_videos`

deruyter92:
Recent testing of the dev branch before 3.1 release undisclosed a few minor bugs or misleading error messages. These are fixed in the current PR:

- The error log did not show correctly which videos were failed to create (the actual create video call uses `collect_videos(videos)` and not the logged `videos`, which in some cases may have a different order or count).
- `collect_videos` emits a warning for unsupported video file formats, but we use the function internally for `.h5` files as well. This PR introduces a silencing flag to not warn in these cases.
- a latent bug caused `analyze_videos_converth5_to_csv` to only work for `.mp4` videos this is now changed to all supported video types.


deeplabcut/gui/tabs/create_videos.py
```diff
@@ -272,11 +273,16 @@ def create_videos(self):
             color_by=color_by,
             overwrite=self.overwrite_videos.isChecked(),
         )
-        if all(videos_created):
+        analyzed_videos = collect_video_paths(videos)
```
C-Achard, line 276: Should this filter by extension in case the folder has more than one supported video type?

deruyter92: `videos` comes from the GUI file picker and is an explicit file list rather than a directory path. 

So here collect_video_paths(videos) just normalizes the selected files in the same way that the create_labeled_videos() backend does. (Both use the default SUPPORTED_VIDEO_EXTENSIONS) So I don’t think extra filtering is helpful here. Unless I'm overlooking something or didn't understand your comment correctly.

C-Achard (approved):

PR #3469, opened 2026-08-28
fix WandbLoggerConfig include additional kwargs in pydantic schema

deruyter92:
The Weights and Biases logger accepts a variety of extra arguments, such as `notes`, `tags`, etc. These were mediated through `wandb_kwargs` which could be defined at top-level. However since the introduction of the stricter pydantic schemas these top-level arguments with `extra="forbid"` would raise a `ValidationError`.

This PR:
- fixes a dead `wandb_kwargs` field in the `WandBloggerConfig` (was never passed correctly to the logger)
- ~adds common top-level fields such as `notes`, `tags`, `group` to the top-level config.~

**Edit:**
To avoid breaking the current config schema I refrained from adding common WandB args like `notes`, `tags`, `group` as top-level fields. Since they may appear in older unversioned config yamls, I've added a before validator that scoops them into the `wandb_kwargs` field. I've also added testing, so that the following YAML forms are accounted for:

```yaml
# Allow both top-level and nested args (even combined, when no conflicts):
logger:
  project_name: "deeplabcut"
  ...
  tags: ["mytag"]  # <-  top-level (will be moved into wandb_kwargs by the validator)
  wandb_kwargs:
     group: "mygroup"  # <- nested


# Don't allow configs like this:
logger:
  project_name: "deeplabcut"
  ...
  project: "myproject"  # <- Reserved fields are not allowed (already defined via "project name", "run_name")
  group: "mygroup"
  wandb_kwargs: 
      group: "repeated"   # <- Duplicate top-level and nested fields are not allowed.
```

### DIFF
--- a/deeplabcut/pose_estimation_pytorch/config/logger.py
+++ b/deeplabcut/pose_estimation_pytorch/config/logger.py
@@ -98,8 +98,9 @@ class WandbLoggerConfig(LoggerConfig):  #
         if not v:
             return v
         if set(v) & cls._FORBIDDEN_WANDB_KWARGS:
-            raise ValueError("wandb_kwargs can not include any of the reserved options: {cls._FORBIDDEN_WANDB_KWARGS}")
-        return v
+            raise ValueError(
+                f"wandb_kwargs cannot include any of the reserved options: {sorted(cls._FORBIDDEN_WANDB_KWARGS)}"
+            )
 
 
 class CSVLoggerConfig(LoggerConfig):  #
--- a/deeplabcut/pose_estimation_pytorch/config/logger.py
+++ b/deeplabcut/pose_estimation_pytorch/config/logger.py
@@ -13,9 +13,9 @@
 from __future__ import annotations
 
 from enum import Enum
-from typing import Any, Literal
+from typing import Any, ClassVar, Literal
 
-from pydantic import model_validator
+from pydantic import field_validator, model_validator
 
 from deeplabcut.core.config import DLCBaseConfig
 
@@ -56,6 +56,9 @@ class WandbLoggerConfig(LoggerConfig):  #
         wandb_kwargs: Additional keyword arguments to pass to wandb.init
     """
 
+    # These options are reserved since they are explicitly passed to wandb.init (from `project_name` and `run_name`)
+    _FORBIDDEN_WANDB_KWARGS: ClassVar[frozenset[str]] = frozenset({"project", "name"})
+
     type: Literal[LoggerType.WandbLogger]
     project_name: str = "deeplabcut"
     run_name: str = "tmp"
@@ -87,6 +90,16 @@ class WandbLoggerConfig(LoggerConfig):  #
 
         data["wandb_kwargs"] = {**nested, **extras}
         return data
+
+    @field_validator("wandb_kwargs")
+    @classmethod
+    def validate_wandb_kwargs(cls, v: dict | None) -> dict | None:
+        """Reject nested keys that collide with reserved wandb.init options."""
+        if not v:
+            return v
+        if set(v) & cls._FORBIDDEN_WANDB_KWARGS:
+            raise ValueError("wandb_kwargs can not include any of the reserved options: {cls._FORBIDDEN_WANDB_KWARGS}")
+        return v
 
 
 class CSVLoggerConfig(LoggerConfig):  #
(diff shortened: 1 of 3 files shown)


deeplabcut/pose_estimation_pytorch/runners/logger.py
```diff
@@ -291,6 +293,9 @@ def __init__(
         if wandb.run is not None:
             wandb.finish()
 
+        # A nested ``wandb_kwargs``mapping must be flattened
+        wandb_kwargs.update(wandb_kwargs.pop("wandb_kwargs", None) or {})
```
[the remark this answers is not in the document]
deruyter92: I would rather do the validation inside pydantic. Plus this comment is outdated anyway.

deeplabcut/pose_estimation_pytorch/config/logger.py
```diff
@@ -56,6 +67,40 @@ class WandbLoggerConfig(LoggerConfig):  #
...
+    @model_validator(mode="before")
+    @classmethod
+    def scoop_extra_wandb_kwargs(cls, data: Any) -> Any:
+        """Move unknown top-level keys into ``wandb_kwargs``"""
+        if not isinstance(data, dict):
+            return data
+
+        known = set(cls.model_fields)
+        provided = set(data)
+        extras = {key: data.pop(key) for key in provided - known}
```
C-Achard, line 80: It may be best to copy beforehand

deeplabcut/pose_estimation_pytorch/config/logger.py
```diff
@@ -56,6 +67,40 @@ class WandbLoggerConfig(LoggerConfig):  #
...
+        overlap = set(extras) & set(nested)
+        if overlap:
+            raise ValueError(f"Duplicate wandb.init options at top-level and in wandb_kwargs: {sorted(overlap)}")
+
+        data["wandb_kwargs"] = {**nested, **extras}
+        return data
+
+    @field_validator("wandb_kwargs")
+    @classmethod
+    def validate_wandb_kwargs(cls, v: dict | None) -> dict | None:
```
C-Achard, line 97: This will forward broken kwargs and fail later in the init call, which may not be ideal

deeplabcut/pose_estimation_pytorch/config/logger.py
```diff
@@ -56,6 +67,40 @@ class WandbLoggerConfig(LoggerConfig):  #
...
     wandb_kwargs: dict | None = None
 
+    @model_validator(mode="before")
+    @classmethod
+    def scoop_extra_wandb_kwargs(cls, data: Any) -> Any:
+        """Move unknown top-level keys into ``wandb_kwargs``"""
+        if not isinstance(data, dict):
+            return data
+
+        known = set(cls.model_fields)
```
C-Achard: DLCBaseConfig.resolve_aliases_before_validate is also mode="before", and pydantic runs those for a subclass first, so the scoop may see raw input and swallow a deprecated alias into wandb_kwargs before it can be mapped to the correct field.

```suggestion
        known = set(cls.model_fields) | set(cls._alias_map())
```

deeplabcut/pose_estimation_pytorch/config/logger.py
C-Achard: I don't think we can fix it in this file though, I'll open a follow-up PR with improved error handling

C-Achard (commented):
Just a few things worth checking, all good otherwise

deeplabcut/pose_estimation_pytorch/config/logger.py
C-Achard: See #3481

deruyter92: Yeah this follows my exact own preference: a follow-up PR with fixes in the logger.__init__  rather than the pydantic model. E.g. we cannot keep syncing with different wandb options (they actually changed regularly). Also don't want to make the config validation more heavy than needed. Your PR is the perfect middle ground IMO

deeplabcut/pose_estimation_pytorch/config/logger.py
deruyter92: Thanks. Good catch.

deeplabcut/pose_estimation_pytorch/config/logger.py
deruyter92: Done, thanks!

C-Achard (approved):

deruyter92:
@C-Achard, addressed your comments. Let me know if I missed anything. Thanks for the follow-up PR as well. Let's merge it separately

C-Achard:
Sounds good, thanks!

Issue #2642, opened 2024-06-28
FasterRCNN: `TypeError: target labels must of int64 type, instead got torch.int32`


Kpraetori:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Bug description

I'm trying to do transfer learning with the Superanimal Quadruped model but I keep getting a couple errors and it won't train. Maybe it's an install issue but I'm not sure and don't know how to resolve it.

1. it throws a notice that I set a batch size of 1 or freeze bn stats = false, neither of which is accurate according to both config files. The config files are attached as text files (won't let me upload .yaml).
2. When it begins start object detector training, a series of traceback errors are thrown and it does not train. I'm attaching the log output.

I have tried on two different servers and rebuilt the environment several times. I have verified nvidia-smi shows the GPU and that torch.cuda.is_available shows true in the environment. I have tried with the latest version of DLC. I have also tried with Pytorch 12.1 and 12.4 in case a newer version was needed. I have followed the regular installation guide in the docs and also the instructions at https://github.com/DeepLabCut/DeepLabCut/pull/2613 to no avail.

I am coming up to a submission deadline so help is really appreciated!

[Complete_Log_DLC_Failing.txt](https://github.com/user-attachments/files/16034320/Complete_Log_DLC_Failing.txt)
[config.txt](https://github.com/user-attachments/files/16034384/config.txt)
[pose_cfg.txt](https://github.com/user-attachments/files/16034385/pose_cfg.txt)
[pytorch_config.txt](https://github.com/user-attachments/files/16034386/pytorch_config.txt)

### Operating System

Windows 2022 Server

### DeepLabCut version

deeplabcut-3.0.0rc1

### DeepLabCut mode

single animal

### Device type

gpu, NVIDIA RTX A4000. I also tried a different server with NVIDIA A10.

```
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 537.42                 Driver Version: 537.42       CUDA Version: 12.2     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                     TCC/WDDM  | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  NVIDIA RTX A4000             WDDM  | 00000000:21:00.0 Off |                  Off |
[...]

PR #3470, opened 2026-08-30
Fix Faster R-CNN detector inference crashing on box_score_thresh: null

### LINKED ISSUE #3505: issue with tensor types

### Operating System

linux

### DeepLabCut version

3.0.1

pytorch

### DeepLabCut mode

single animal

### Device type

RTX A4000

### Bug description 🐛

Hello - we downloaded/updated to the newest version of deeplabcut and keep getting this error "'>' not supported between instances of 'Tensor' and 'NoneType'" that we didnt get before/don't get with a back-up of our older deeplabcut version. We are running into this issue whenever we try to train a new model or run a model on new videos.

### Steps To Reproduce

_No response_

### Relevant log output

```shell

```

### Anything else?

_No response_

### Code of Conduct

rysk-t:
Hello! This is my first PR here — thank you for DeepLabCut and its important contributions to the scientific community.

Faster R-CNN detectors trained through DeepLabCut cannot run inference.
Generated `pytorch_config.yaml` files store `box_score_thresh: null` in the
detector model section (`DetectorModelConfig.box_score_thresh` defaults to
`None` since the structured-config migration in #3198, first released in
v3.0.1), and `DETECTORS.build` forwards that `None` verbatim, overriding the
constructor default. torchvision's Faster R-CNN then keeps
`score_thresh=None` and every inference call crashes in
`postprocess_detections`:

    TypeError: '>' not supported between instances of 'Tensor' and 'NoneType'

The crash hits every eval-mode forward, on every device and OS: inference
with a trained detector, and also the validation pass that detector
training runs every `eval_interval` epochs (10 by default) — so a
default-length Faster R-CNN detector training run crashes at its first
validation epoch. Training-mode forwards are unaffected (`score_thresh` is
only used in `postprocess_detections`). The default `ssdlite` detector is
not affected (torchvision's SSD ignores this kwarg entirely; its own
parameter is `score_thresh`), and the SuperAnimal model configs pin an
explicit `box_score_thresh: 0.6` — which is why the bug only hits projects
that explicitly select a Faster R-CNN `detector_type`, and went unnoticed
on the default paths.

Reproduction:

    from deeplabcut import Engine
    deeplabcut.create_training_dataset(config, net_type="top_down_resnet_50",
                                       detector_type="fasterrcnn_resnet50_fpn_v2",
                                       engine=Engine.PYTORCH)
    deeplabcut.train_network(config, detector_epochs=1, epochs=1)
    deeplabcut.analyze_videos(config, [video])  # -> TypeError

Fix: treat `box_score_thresh=None` as "use the default" (0.01) in the shared
`TorchvisionDetectorAdaptor`, which also heals existing on-disk configs that
already contain `null` — no config regeneration needed. Behavior for any
non-`None` value is unchanged, and SSDLite construction is unchanged (we
verified torchvision's SSD resolves to the same `score_thresh` either way).
The code that reads the config value for plotting cutoffs
(`evaluate_snapshot`, `create_labeled_video`, via `... or 0.6`) reads the
YAML, not the model instance, and is unaffected. Subclass signatures are
updated to `float | None` to match.

A unit test covers the `None` and explicit-value cases
(`tests/pose_estimation_pytorch/models/test_detectors.py`).

Related: #2642 is a sibling dtype/config issue in the same detector target
path (`int32` labels where torchvision expects `int64`).

## Testing

- New unit test: `tests/pose_estimation_pytorch/models/test_detectors.py`
- `pytest tests/pose_estimation_pytorch`: 1011 passed, 29 skipped
  (macOS arm64, Python 3.12, torch 2.12.1 / torchvision 0.27.1)
- End-to-end: a Faster R-CNN detector trained on a real multi-animal project
  reproduces the crash on CPU with the generated config, and runs correctly
  with this fix
- `pre-commit run` (manual stage) and `ruff check` / `ruff format --check`: clean
- The change was independently double-checked by two AI code reviewers
  (Anthropic Claude Fable 5 and OpenAI Codex 5.6) in addition to the test
  suite


C-Achard:
Hello @rysk-t,

Thanks a lot for the fix! This is indeed an issue with the current handling.
I think we can make the solution slightly easier to maintain by centralizing the default value, and ensuring the config schema itself validates None upon loading.

I've pushed these additions in 6db842dbaa4f5497aa18e864cebe041aa334a61e, please let me know if you agree.
Otherwise everything looks good, thanks for the additional regression coverage as well.

Best,
Cyril

rysk-t:
Thanks, Cyril (@C-Achard)!

I agree with centralizing the default and normalizing `None` at the schema level — that's cleaner than the adaptor-only guard, and it also heals existing configs on load.

One note: the original PR applied `0.01` only when constructing the model, so the config value itself stayed `None`. Moving the default into the schema changes that value, which causes two problems when running the branch locally (macOS arm64, torch 2.12.1):

1. **A test case fails.** `tests/pose_estimation_pytorch/config/test_pose_config_creation.py::test_pose_config_creation[multi_resnet_50_td_ssdlite]` asserts that the generated ssdlite config has `box_score_thresh is None`, and the schema default is now `0.01`. The assertion needs updating to the new default (everything else passes: 1011 passed).

2. **The bbox plotting cutoff changes.** `evaluate_snapshot` (`evaluation.py:625`) and `create_labeled_video` (`make_labeled_video.py:608`, `:1196`) read the config value with `... or 0.6` for the bbox plotting cutoff. Previously `null` fell through to `0.6`; now they get `0.01`, so evaluation plots and labeled videos will draw many more low-confidence boxes. If that's not intended, those call sites could keep a dedicated plotting default instead of reusing the model threshold.

Happy to push either fix if you'd like — just let me know.

C-Achard:
No worries @rysk-t, thanks for the extra checks! I think a possibly cleaner option moving forward would be to have a dedicated cutoff config field, this broadens the PR scope a bit but should retain backwards comp and avoid having ad-hoc defaults spread everywhere.

Even if that particular change is reverted, we still intend to merge your fix with the next PR batch from dev. Thanks again!

rysk-t:
Thanks, Cyril.

That seems like the right decision. Decoupling the cutoff value for the plot from the model's threshold yields a cleaner result than what I originally proposed.

Thank you for the quick review — glad I could contribute to DeepLabCut.

deruyter92 (commented):
@rysk-t, excellent PR! Thanks a lot for spotting this real bug, providing a clear description and for the well-designed solution.

@C-Achard and @rysk-t, I like your proposed solutions. A few things worth discussing:

1. the model `box_score_threshold` is used internally in the torchvision detectors and is a completely separate concept from the `bboxes_pcutoff` which is used for plotting. I think you've both addressed that well, and I agree that `bboxes_pcutoff` deserves it's place in the `ProjectConfig`. Thanks for adding that @C-Achard. We might need to discuss the forward compatibility in older versions. (The current change means that migrating a 3.1 config to 3.0.1 will break).
1. There was another call site (`create_video_with_all_detections`) that wrongly used the old parameter from the detector config, I've pushed some extra changes to address this. 
1. The null normalization to default in the `DetectorModelConfig` is an elegant solution! I am opening a follow-up PR (#3475) to make this even available for other config models as well. This pattern may be needed in more places.
1. One solution I did not entirely agree with entirely is the shared `DEFAULT_BOX_SCORE_THRESH`: since null-normalization happens in the config, we do not need to support `box_score_thresh` as an optional argument, we can always require a float. I prefer keeping the null-normalization just in a single place (the config). Also, coupling the class defaults to the config model default feels a bit wrong, as I would prefer to always use the serialized value. Inside DeepLabCut it always goes via DetectorModelConfig anyway, which is authorative.  


Sorry for the elaborate review, but I think it is a good first example case, so I think it is worth some discussion. Happy to hear if both of you agree. Open for suggestions.

C-Achard:
> We might need to discuss the forward compatibility in older versions. (The current change means that migrating a 3.1 config to 3.0.1 will break).

Agreed, it does create the need for some minimal migration logic, let's make a follow-up PR.

deruyter92:
(Note to self and other devs: before the config refactors `None` was not valid for the torchvision detectors `box_score_thresh: float = 0.01`)

deruyter92:
@C-Achard, after our internal discussion I've updated `bboxes_pcutoff` to:
- be a property rather than a field in `ProjectConfig` to avoid breaking the schema right now (and get this PR merged first)
- be configurable in every caller, so that it does not rely on the config alone
- xfail until it's added as a field in `ProjectConfig`, as a reminder for ourselves that we can adjust this in the config schema update

deruyter92:
Let me know if you agree. I will mark you as a reviewer! I'm approving both of your changes

deruyter92 (approved):

rysk-t:
Thanks for the thorough review. I agree with all four points, and your reasoning on DEFAULT_BOX_SCORE_THRESH makes sense.
Everything looks good on my side, so let me know if there's anything else you need from me.

C-Achard:
Perfect, thanks @deruyter92!

C-Achard (approved):

PR #3471, opened 2026-08-31
Fix faulty str / Path conversions

C-Achard:
Minor fixes where str() conversion was missing


deruyter92:
Note to ourselves: this was a replacement backup PR for #3467, but not necessary anymore.

PR #3472, opened 2026-08-31
Update CI to target the dev branch

C-Achard:
Adds the dev branch to CI targets for testing, spell checks and uv lockfile checks.

### To consider when reviewing/merging

- [x] Ensure there are no side effects from adding dev
  - Aside from extra CI runs, there is no deploy logic that is specific to main besides docs
- [ ] [**ADMIN ONLY**] Ideally apply ruleset for main to dev, or a lightly updated one (if we could discuss together @MMathisLab @AlexEMG, this shouldn't take long to setup)
- [x] To save on CI runner congestion, avoid running on draft dev PRs

Fixes a workflow name consistency issue in uv lock check workflow.

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -2,7 +2,7 @@ name: Docs / Deploy docs (main + dev-docs main)
 
 on:
   push:
-    branches: [ main ]
+    branches: [ main ] # do not add dev/ here
 
 permissions:
   contents: write
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -2,7 +2,7 @@ name: Docs / Deploy docs (main + dev-docs main)
 
 on:
   push:
-    branches: [ main ] # do not add dev/ here
+    branches: [ main ] # do not add dev here
 
 permissions:
   contents: write
--- a/.github/workflows/intelligent-testing.yml
+++ b/.github/workflows/intelligent-testing.yml
@@ -2,10 +2,10 @@ name: Code / Intelligent Python Testing
 
 on:
   push:
-    branches: [ main ]
+    branches: [ main, dev ]
   pull_request:
     types: [opened, synchronize, reopened]
-    branches: [ main ]
+    branches: [ main, dev ]
 
 concurrency:
   group: intelligent-${{ github.event.pull_request.number && format('pr-{0}', github.event.pull_request.number) || format('run-{0}', github.run_id) }}
--- a/.github/workflows/intelligent-testing.yml
+++ b/.github/workflows/intelligent-testing.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches: [ main, dev ]
   pull_request:
-    types: [opened, synchronize, reopened]
+    types: [opened, synchronize, reopened, ready_for_review, converted_to_draft]
     branches: [ main, dev ]
 
 concurrency:
@@ -19,6 +19,7 @@ jobs:
     name: Select test plan
     runs-on: ubuntu-latest
     outputs:
+      is_dev_draft: ${{ steps.is_dev_draft.outputs.is_dev_draft }}
       run_skip: ${{ steps.selector.outputs.run_skip }}
       run_docs: ${{ steps.selector.outputs.run_docs }}
       run_fast: ${{ steps.selector.outputs.run_fast }}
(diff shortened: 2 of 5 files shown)


.github/workflows/intelligent-testing.yml
```diff
@@ -19,6 +19,7 @@ jobs:
     name: Select test plan
     runs-on: ubuntu-latest
     outputs:
+      is_dev_draft: ${{ steps.is_dev_draft.outputs.is_dev_draft }}
```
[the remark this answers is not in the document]
C-Achard: Not applicable since test selection takes <30s usually, no congestion will occur in a way that is comparable to a 40min long full test workflow.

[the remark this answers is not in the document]
C-Achard: This also keeps better granularity imo in re-enabling/disabling specific jobs on dev

deruyter92:
Great! This will make our life easier. 

As discussed: it would be great if the intelligent testing CI does *not* run on draft PRs, which could help avoid some unnecessary overflow.

C-Achard:
@deruyter92 I've made it so drafts targeting dev skip all heavier workflows for now. Tested on the fork, works well, including re-running check on draft status change.

See https://github.com/DeepLabCut/DeepLabCut/pull/3472/commits/9ee45b904af953940665297b52139caad998b6d7

.github/workflows/intelligent-testing.yml
```diff
@@ -31,6 +32,15 @@ jobs:
...
     steps:
+      - name: Evaluate if PR is a draft on dev/
+        id: is_dev_draft
+        run: |
+          echo "is_dev_draft=${{
+            github.event_name == 'pull_request' &&
+            github.base_ref == 'dev' &&
+            github.event.pull_request.draft == true
+          }}" >> "$GITHUB_OUTPUT"
+
```
deruyter92, line 43: is it intended that only the downstream lanes skip when `is_dev_draft != 'true'`? Wouldn't it be better if it skips the tests altogether?

deruyter92: especially since `converted_to_draft` is now also a trigger: changing the status to draft would preferably not run any CI right?

deruyter92: should we still run Codespell, docs/notebook checks, lockfile and pre-commit on drafts?

I can imagine some of these are still useful, but maybe just as a manual dispatch for drafts?

.github/workflows/deploy-docs.yml
```diff
@@ -2,7 +2,7 @@ name: Docs / Deploy docs (main + dev-docs main)
 
 on:
   push:
-    branches: [ main ]
+    branches: [ main ] # do not add dev/ here
```
deruyter92:
```suggestion
    branches: [ main ] # do not add dev here
```

.github/workflows/intelligent-testing.yml
C-Achard: > especially since converted_to_draft is now also a trigger: changing the status to draft would preferably not run any CI right?

It will create the concurrency guard to cancel the workflow though, which may be desirable if you explicitly revert to draft

C-Achard: > should we still run Codespell, docs/notebook checks, lockfile and pre-commit on drafts?

I would say yes which is why I kept them. Would you rather omit them entirely?

C-Achard: > is it intended that only the downstream lanes skip when is_dev_draft != 'true'? Wouldn't it be better if it skips the tests altogether?

It is indeed redundant right now, it was mostly to keep granularity in the future

deruyter92: only saw your reply to co-pilot after my review, sorry for the overlap. 

Agreed on all points: since they are fast we can keep them.

.github/workflows/intelligent-testing.yml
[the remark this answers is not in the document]
deruyter92: agreed!

deruyter92 (approved):

deruyter92:
Tested it as well and works nicely! Super helpful that intelligent testing now does not run on drafts. (We can always decide later if we want to also skip the smaller tests). 

Would be nice if we can merge ASAP!

AlexEMG (approved):

PR #3473, opened 2026-08-31
fix `export_model` dumping unnormalized objects in pickle

deruyter92:
Export model unintendedly missed the `normalize=True` flag, which makes sure that all objects in the config are normalized to primitive types: e.g. `Path` -> `str`,  `Enum` -> `str`, etc

This is addressed in the current PR by adding `normalize=True` in the `PoseConfig.to_dict()` call.

### DIFF
--- a/tests/pose_estimation_pytorch/apis/test_apis_export.py
+++ b/tests/pose_estimation_pytorch/apis/test_apis_export.py
@@ -224,10 +224,10 @@ def test_export_model(
 
 def _assert_plain_data(obj, path: str = "config") -> None:
     """Asserts that ``obj`` only contains primitive types, recursively."""
-    if isinstance(obj, dict):
+    if type(obj) is dict:
         for key, value in obj.items():
             _assert_plain_data(value, f"{path}[{key!r}]")
-    elif isinstance(obj, (list, tuple)):
+    elif type(obj) in (list, tuple):
         for index, value in enumerate(obj):
             _assert_plain_data(value, f"{path}[{index}]")
     else:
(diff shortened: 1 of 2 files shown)


tests/pose_estimation_pytorch/apis/test_apis_export.py
```diff
@@ -222,6 +222,39 @@ def get_mock_loader(*args, **kwargs):
             assert exported_data["detector"] == detector_data[detector_idx]["model"]
 
 
+def _assert_plain_data(obj, path: str = "config") -> None:
+    """Asserts that ``obj`` only contains primitive types, recursively."""
+    if isinstance(obj, dict):
+        for key, value in obj.items():
+            _assert_plain_data(value, f"{path}[{key!r}]")
+    elif isinstance(obj, (list, tuple)):
```
C-Achard: Is tuple intentional here?

C-Achard (approved):
Thanks for the fix!

tests/pose_estimation_pytorch/apis/test_apis_export.py
deruyter92: Well, good question. `tuple` is a pickle primitive type and also well handled by `torch.load`, so I think it deserves to stay here.

A tangential discussion point would be: should we adjust this in the `DLCBaseConfig` behavior? Currently `config.to_dict(normalize=True)` actually explicitly keeps tuples as tuples instead of lists. This works ok since our centralized yaml dumpers handle tuples well, but we can argue that converting them to list would be better. In that case I propose to open a separate PR for this.

deruyter92: BTW I've changed the instance checks to explicit type checks now, since raw unsupported subclasses (e.g. ruamel `CommentedMap`) would accidentally survive here.

C-Achard: > should we adjust this in the DLCBaseConfig behavior? 

I think it's fine, as long as we keep in mind the JSON case, where we would get lists rather than tuples. I don't think it's a big concern. Thanks!

PR #3475, opened 2026-09-01
Add config validator that maps null -> default

### LINKED ISSUE #3505: issue with tensor types

deruyter92:
Follow-up PR for #3470, introducing a central validator for cases when null needs to be mapped to default.

### DIFF
--- a/deeplabcut/core/config/validation.py
+++ b/deeplabcut/core/config/validation.py
@@ -1,9 +1,10 @@
 from collections.abc import Sequence
-from typing import Annotated, Any
+from typing import Annotated, Any, TypeAlias, TypeVar
 
 import numpy as np
 from numpy.typing import NDArray
 from pydantic import AfterValidator, BeforeValidator, Field, GetPydanticSchema, InstanceOf
+from pydantic_core import PydanticUseDefault
 
 
 def _describe(value: float, name: str | None = None) -> str:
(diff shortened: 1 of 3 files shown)


deeplabcut/core/config/validation.py
```diff
@@ -65,6 +66,19 @@ def _bodypart_pair(values: Sequence[Any]) -> list[str]:
     return list(unique_values(values))
 
 
+def _use_default_if_none(value: Any) -> Any:
```
C-Achard, line 69: Interesting!

tests/core/config/test_validation.py
```diff
@@ -0,0 +1,217 @@
...
+        with pytest.raises(ValidationError, match="must be unique"):
+            ListConfig(skeleton=[["nose", "nose"]])
+
+
+# ------------------------------------------------------------------
+# NDArrayInt
+# ------------------------------------------------------------------
+
+
+class TestNDArrayInt:
```
C-Achard, line 201: Out of scope for this PR but in hindsight I find the name to be slightly confusing. No concerns on functionality at all, just the int part.

C-Achard (approved):
Looks good, much cleaner now; thanks for the extra tests!

PR #3480, opened 2026-09-04
fix gui tests: use `config_path` in `MainWindow` config tests

deruyter92:
Tests still assigned `MainWindow.config` after the field was renamed to `config_path` in 30a8155d (#3350).

The `dev` branch currently fails CI on two GUI tests:
- _test_task_error.py_
- _test_main_window_config.py_

This PR fixes that by renaming `config` to `config_path` accordingly in the corresponding tests.


C-Achard (approved):
Looks good, thanks!

PR #3481, opened 2026-09-04
Harden wandb logger error handling

C-Achard:
- Validate model and train_folder before any wandb call, so unusable kwargs no longer open a W&B run before raising
- Guard wandb.init so an invalid kwarg surfaces as a config error naming the forwarded options and pointing at the logger: section of pytorch_config.yaml, instead of a bare TypeError.


deruyter92 (approved):

deruyter92:
thanks, good follow-up

C-Achard:
@deruyter92 Feel free to merge now or later in dev at your convenience!

PR #3494, opened 2026-09-08
Follow-up hardening for the LLM knowledge index

deruyter92:
#3453 introduced a new knowledge-indexing system that deploys indexes of the user docs and API references in JSONL files and a central llms.txt that points to the latest sources.

This follow-up PR implements some follow-up hardening based on additional review and testing (thanks to @C-Achard):
- **add dependency bounds to the knowledge-indexing group**
- **instead of a single link in the llms.txt, reference two pages for the dev API docs: the latest stable release (alias) + the rolling main.**"
- Reject `--skip-api` on non-`main` labels when that would leave nothing to do (was a silent success).
- Write empty JSONL as an empty file (no stray blank line).
- Check API provenance before creating a version directory, so failed skip-api runs do not leave empty dirs that `keep_files` deploys forever.
- Keep published-anchor occurrence counts aligned when a heading has no usable section id.
- Add focused regression tests for docs link resolution, frontmatter hiding, and package-root API re-export URLs.

Review points not implemented: 
- `::::{warning}` admonitions are still ignored in the summaries (they only exist in some hardly-used recipes)
- the `dev-docs` dependency extra is not re-implemented. I kept the dev-docs depencency group as a group not extra. I think we are the only ones using them so I believe it is fine to "break" the old installation contract of `--extra dev-docs`.


C-Achard:
> I kept the dev-docs depencency group as a group not extra. I think we are the only ones using them so I believe it is fine to "break" the old installation contract of --extra dev-docs

Agreed

C-Achard:
@deruyter92 See the follow-up fixes; thanks!

C-Achard (approved):

PR #3495, opened 2026-09-09
[Sync] merge to main for 3.0.2 release - bug fixes; knowledge indexing; future compatibility & dependencies  

deruyter92:
## Summary
Merge the current `dev` snapshot to `main` in preparation for 3.0.2 release.

> [!CAUTION]
> This **does not** bump the project version yet. Do not make a PyPI release before merging the version bump in #3496 

### Dependencies & Future compatibility
Adapt old usage patterns of libraries that changed their API, allowing us to release the upper bounds of our dependency stack and improve broad compatibility.
- ⚠️ #3499 — remove dependency upper bounds for numpy and matplotlib
- #3420 — Prepare NumPy 2 forward compatibility 
- #3508 — Follow-up with additional fixes
- #3445 — Guard pandas 3 CoW read-only `.to_numpy()` / `.values` mutations
- #3394 — Matplotlib v3.9 compatibility
- #3498 — fix broken numpy 2.0 compat in crossvalutils and predict_multianimal
- #3502 — NumPy 2 + numba: take real part of eig eigenvectors in EllipseFitter
- #3509 — Tighten strict canary CI handling (strict_cells for same-repo pushes/PRs)

### Agentic interaction
Improve the access to domain-specific knowledge for agents that interact with DeepLabCut.  
- ⚠️ #3483 — Add `AGENTS.md` for AI-assisted development 
- #3453 — Knowledge indexing for LLM-friendly API/docs index 
- #3494 — Follow-up hardening for the LLM knowledge index 

### Bug fixes
- ⚠️ #3441 — GUI config editor crash + shared deferred reload on `DefaultTab`
- ⚠️ #3435 — Fix `COCOLoader`: treat `model_cfg.metadata` / `max_individuals` as authoritative
- ⚠️ #3470 — Faster R-CNN inference crash when `box_score_thresh: null`
- #3514 — Fix hard-coded overwrite=False in PyTorch analyze_videos / auto-track path
- #3419 — Skip TF-backed tests cleanly when the `tf` extra is absent
- #3439 — Legacy ProjectConfig `None`/empty handling; remove `ignore_empty`
- #3447 — Evaluation visualization: scorer/individual mismatches, harden plotting + tests
- #3449 — Frame extraction failure with pathlib video selections (GUI/API)
- #3452 — Fix incorrect SuperAnimal HumanBody rename mapping on Hugging Face download
- #3467 — Faulty `str` / `Path` conversions
- #3468 — Latent bugs / misleading errors in GUI `create_videos`
[...]

PR #3499, opened 2026-09-09
remove dependency upper bounds for numpy and matplotlib

deruyter92:
numpy and matplotlib had upper bounds because we used legacy patterns that are now deprecated. These are addressed in the PRs below, allowing us to remove the upper bounds. This allows for broader compatibility. 

- https://github.com/DeepLabCut/DeepLabCut/pull/3420 
- https://github.com/DeepLabCut/DeepLabCut/pull/3445
- https://github.com/DeepLabCut/DeepLabCut/pull/3394
- https://github.com/DeepLabCut/DeepLabCut/pull/3498
- https://github.com/DeepLabCut/DeepLabCut/pull/3508


C-Achard (approved):
Let's keep pandas pinned for now, matplotlib is definitely fine.
numpy is inherently more risky but I also agree that we should unpin.

C-Achard:
@deruyter92 Some additional findings

### numpy lower bound and pickle

Once the `<2` cap is gone, np-1 and np-2 might see some exchange of project files. 
np2 can load pickle files created with np 1.26, and vice versa. For 1.25 and earlier loading np2 pickles will throw an exception.
See [NumPy 2.0 migration guide](https://numpy.org/doc/stable/numpy_2_0_migration_guide.html).

### We may want `numpy<2` only for `tf-cu11`

From PyPI metadata for each pinned TF version:

| extra | tensorflow | its own numpy specifier | needs a DLC cap? |
|---|---|---|---|
| `tf-cu11` | `==2.14` | `numpy>=1.23.5` — **no upper bound** | **yes**, the compile failure |
| `tf-cu12` | `==2.18` | `numpy<2.1.0,>=1.26.0` | **no** — supports numpy 2.0 already |
| `tf` | `2.12`–`2.17` | 2.15.1 `<2.0.0,>=1.23.5`; 2.16.1/2.17 `<2.0.0` | no |
| `apple_mchips` | `2.12`–`2.17` | same as above | no |

### CI changes

Every full-suite job installs the `[tf]` extra so after merge, CI resolves numpy 1 everywhere. We may have to add more jobs.
The downloaded fixtures (`tests/data/{dets,montblanc_*,trimouse_*}.pickle`, `montblanc_tracks.h5`) were written under numpy 1 and are readable by numpy 2, so they work.

### Removing both caps together is required

matplotlib 3.8.4 declares only `numpy>=1.21`, with **no** upper bound. So dropping `numpy<2` while leaving `matplotlib<3.9` would try to resolve matplotlib 3.8.4 + np2. Same risk for any compiled transitive dep that declares no numpy cap (`pycocotools`, `tables`, `h5py`, `scikit-image`), so an import smoke test under numpy 2 may be worth it.

### numba

numba 0.54 requires `numpy<1.21`, which no DLC-supported numpy version satisfies. Per numba's support matrix, 0.60.0 is the first release supporting numpy 2.0 and the latest (0.67) supports `2.0 <= numpy < 2.6`.
Should still play nice with TF opt deps I think.

### TF specific (low priority)

`np.lib.pad` is still used at [`visualize.py:93`](https://github.com/DeepLabCut/DeepLabCut/blob/dev/deeplabcut/pose_estimation_tensorflow/util/visualize.py#L93) and [`vis_dataset.py:70`](https://github.com/DeepLabCut/DeepLabCut/blob/dev/deeplabcut/pose_estimation_tensorflow/vis_dataset.py#L70). Unlikely to be a problem as we're not migrating TF to np2.

deruyter92:
> We may want numpy<2 only for tf-cu11

@C-Achard agreed, just arrived at the same conclusion

deruyter92:
> Every full-suite job installs the [tf] extra so after merge, CI resolves numpy 1 everywhere. We may have to add more jobs.

~Instead, we should just remove the upperbound from `tf` right?~

**Edit:** TF defines an upper bound themselves, which means that installing the `tf` extra naturally resolves to numpy 1.x for tensorflow versions up to 2.17

deruyter92:
Let's keep an eye on the canary lane CI. Errors should surface here (softly).

C-Achard:
Fixed numba + errors surfaced by the canary lane in #3508

C-Achard:
We should really decide whether this gets included now or later; I would include #3508 in 3.0.2 anyway but I think either:

- We ship 3.0.2 ASAP with no pins release
- Or we spend a bit more time on it while #3495 is being reviewed, and make sure everything is viable

but I don't think merging this as is right now is advisable.

**Missing**:

- [ ] napari release with no numpy pin
- [ ] #3508 
- [ ] #3509 / #3510

C-Achard:
Note: fmpose3d also still pins numpy. See https://github.com/AdaptiveMotorControlLab/FMPose3D/pull/36

deruyter92:
> I don't think merging this as is right now is advisable
 Of course, agreed. We won't merge this until CI is completely green. Assuming that our test-coverage is enough, I would say let's merge after our tests indicate we are ready. Any remaining compat issues from untested code can be easily addressed when they surface.

>  I think either:
> - We ship 3.0.2 ASAP with no pins release
> - Or we spend a bit more time on it while #3495 is being reviewed

Agreed. I would say **we ship 3.0.2 ASAP** in any case. If these PRs are fully resolved before the review + release is finished, we can include it. Otherwise we will ship in the next release.

**Review comments below**.  @C-Achard agreed? Shall I address these myself?

pyproject.toml
```diff
@@ -110,6 +111,7 @@ tf-cu11 = [
 tf-cu12 = [
   "imgaug>=0.4",
   "protobuf<7",
+  # No numpy cap needed: tensorflow 2.18 declares `numpy<2.1.0,>=1.26.0`
   "tensorflow==2.18",
   "tensorflow-metal==1.2; platform_system=='Darwin'",
   "tensorpack==0.11",
```
deruyter92, line 121: I think maybe we shouldn't try to over-manage our upstream compatibility: i.e. the dependency compat should be derived from the TensorFlow metadata, and not specified manually here, *unless they are not sufficient*.  In the case of `tf-cu11` we *need* the cap on `numpy` since `tensorflow` v2.14 does not specify an upper bound  themselves.

.github/workflows/intelligent-testing.yml
```diff
@@ -152,12 +152,16 @@ jobs:
       matrix_json: >-
         {
           "include": [
-            {"id":"numpy","os":"ubuntu-latest","python-version":"3.12","extras":"","pip_overrides":"numpy>=2"},
+            {"id":"resolve-py312","os":"ubuntu-latest","python-version":"3.12","extras":"","strict":"true"},
+            {"id":"resolve-py310","os":"ubuntu-latest","python-version":"3.10","extras":"","strict":"true"},
```
deruyter92, line 156: `"strict": "true"` seems to be logic from a different branch or not? Shall we remove it here?

C-Achard: Yes, sorry

pyproject.toml
C-Achard: If you mean keep the upper bound for tf-cu11 only then yes fully agree, does the suggested edit diff affect anything else? It's just the comment correct? (I agree it can go)

C-Achard: For tf the range does include versions without the correct upper bound right? For 2.12 to 2.14

.github/workflows/intelligent-testing.yml
C-Achard: Fixed now

PR #3508, opened 2026-09-11
Follow-up fixes for numpy2 compatibility

### LINKED ISSUE #3512: TypeError: can't assign a numpy.float32 to a torch.FloatTensor

### Operating System

ubuntu 24.04 desktop, vs code insiders, conda, numpy 2.4

### DeepLabCut version

# Name                    Version                   Build  Channel
deeplabcut                3.0.0rc14                pypi_0    pypi
napari-deeplabcut         0.3.1.1                  pypi_0    pypi

pytorch

### DeepLabCut mode

multi animal

### Device type

GeForce RTX 4070 Ti; cuda version 13.2, driver 595.84; 13th Gen Intel(R) Core(TM) i9-13900K

### Bug description 🐛

DLC hits corrupted frames (unavoidable across 600,000 videos with our hardware), attempts to assign NaN. pytorch catches it / it violates strict type checking. 

Fix:
/deeplabcut/pose_estimation_pytorch/models/predictors/paf_predictor.py
line 199: poses[i, j, :, 4] = assembly.affinity.item()

added .item(), extracts item directly. pose prediction then functions as expected.

### Steps To Reproduce

load video with corrupted frame --> it'll end up a NaN during deeplabcut.analyze_videos pass.

### Relevant log output

```shell
lost it when i fixed the bug (see above, fix is in description)
```

### Anything else?

_No response_

### Code of Conduct

C-Achard:
## Motivation

Preparation for [#3499](https://github.com/DeepLabCut/DeepLabCut/pull/3499), which removes the numpy<2 and matplotlib<3.9 upper bounds.

## Changes

- **Fixed Assembly.affinity type to be float consistently**
  Without, poses[i, j, :, 4] = assembly.affinity raises TypeError: can't assign a numpy.float32 to a torch.FloatTensor under NumPy 2, on the default multi-animal inference path.
  - Added regression tests
- **Updated _add_numpy_to_torch_safe_globals**
  numpy.core became numpy._core in NumPy 2, so the import failed. Now tries both module paths and logs a warning.
- **Dependencies**
  - numba>=0.60 (pyproject.toml)
  numba 0.54 requires numpy<1.21, below our own numpy>=1.22.4. 0.60 is also the first release supporting NumPy 2.
  - Declare opencv-python-headless>=4.10.0.84 (pyproject.toml)
cv2 is imported across deeplabcut/ but was not declared, it arrived only transitively via albumentations, imgaug and fmpose3d, so which build you got depended on which extras you installed.


deeplabcut/pose_estimation_pytorch/models/predictors/paf_predictor.py
```diff
@@ -191,7 +191,7 @@ def forward(self, stride: float, outputs: dict[str, torch.Tensor]) -> dict[str,
             if assemblies is not None:
                 for j, assembly in enumerate(assemblies):
                     poses[i, j, :, :4] = torch.from_numpy(assembly.data)
-                    poses[i, j, :, 4] = assembly.affinity
+                    poses[i, j, :, 4] = float(assembly.affinity)
```
deruyter92: Since the property now always returns a float this is not necessary here anymore right?

```suggestion
                    poses[i, j, :, 4] = assembly.affinity
```

pyproject.toml
```diff
@@ -39,8 +39,9 @@ dependencies = [
   "imageio-ffmpeg",
   "matplotlib>=3.3,<3.9,!=3.7,!=3.7.1",
   "networkx>=2.6",
-  "numba>=0.54",
+  "numba>=0.60",
   "numpy>=1.22.4,<2",
+  "opencv-python-headless>=4.10.0.84",
```
deruyter92, line 44: `opencv-python` is used by some other dependencies (e.g. in the extras `fmpose3d` or `tf`). Since both packages install cv2, I think it would be safer to use the same, right?
 
```suggestion
  "opencv-python>=4.10.0.84",
```

deruyter92 (approved):
Minor suggestions. Thanks for the fix!!

deeplabcut/pose_estimation_pytorch/models/predictors/paf_predictor.py
C-Achard: No, just kept it to be very explicit but not strictly required indeed

pyproject.toml
C-Achard: Good point, I'm not sure if they'd conflict or if one can install the full on top of headless. Let me check

C-Achard: I don't think headless is a good idea here, mainly because:

-  albumentations (core) and napari-deeplabcut (gui) require opencv-python-headless, while imgaug (tf*) and fmpose3d require the full opencv-python. A deeplabcut[tf] env installs both.
- Declaring opencv-python instead means albumentations would still pull headless, so a pip install deeplabcut with no extras would end up with both. Headless keeps the base and GUI installs headless and confines the overlap to the tf/fmpose3d extras.
- DLC never opens a cv2 window, and headless avoids the libGL.so.1 issues on CI/Colab.

**TL;DR Right now we rely on albumentations to install this so as is this is not a change technically, whereas removing headless would be.**

What do you think?

C-Achard: (given DLC does not need full cv2 I'm leaning towards leaving as is personally)

deruyter92: @C-Achard ok good analysis, then yes let's keep headless as is. It is the better default anyway, since we actually use it headlessly. 

I'll open an issue on FMPose3D

PR #3509, opened 2026-09-11
Tighten strict handling for canary CI

C-Achard:
Add a `strict_cells` workflow input and use it to make only `matrix.strict` canary lanes fail the job outright. 
The intelligent testing workflow now lets one define  strict canary checks for pushes and same-repo PRs while keeping fork PRs non-blocking, and the reporting step skips annotations for lanes that already fail on strict.
A global toggle is kept if needed.

Note that the "resolve" jobs I added in #3499 may not be needed anymore depending on how we change the full test CI to include tf (and therefore np1) or not and how. 
The goal is anyways to provide more flexibility in checks, which should be useful even if the exact jobs do not remain (esp. since #3499 and associated should make some of them irrelevant already).


C-Achard:
@deruyter92 I can also (and would like to) integrate the canary lane more closely to the test selector, so it runs only on deps changes. Do you think that would be interesting?

pyproject.toml
```diff
@@ -110,6 +110,7 @@ tf-cu11 = [
 ]
 tf-cu12 = [
   "imgaug>=0.4",
+  "numpy>=1.22.4,<2",
```
deruyter92: > No numpy cap needed: tensorflow 2.18 declares `numpy<2.1.0,>=1.26.0`

I think this was a stray remainder of some merging/basing maybe:

```suggestion
```

deruyter92 (approved):
Good PR thanks for adding this! No changes needed IMO, besides a stray numpy upper bound that I think is from an old base maybe

pyproject.toml
C-Achard: My bad, thanks!

PR #3510, opened 2026-09-11
Raise lower bounds to NumPy 2 capable releases

C-Achard:
### Motivation

#3499 removes the `numpy<2` upper bound, but several dependencies still have a np1-only lower bound.

| Dep | From | To | py3.10 resolves | py3.12 resolves |
|---|---|---|---|---|
| `matplotlib` | `>=3.3,!=3.7,!=3.7.1` | `>=3.9` | 3.10.9 | 3.11.1 |
| `pycocotools` | *(unpinned)* | `>=2.0.8` | 2.0.11 | 2.0.11 |
| `scikit-image` | `>=0.17` | `>=0.24` | 0.25.2 | 0.26.0 |
| `scipy` | `>=1.9` | `>=1.13` | 1.15.3 | 1.18.1 |
| `statsmodels` | `>=0.11` | `>=0.14.2` | 0.15.0 | 0.15.0 |
| `tables` | `>3.8` | `>=3.10.1` | **3.10.1** | 3.11.1 |

### Re: `tables`

`tables` 3.10.0 still declares `numpy<2,>=1.20.0`; the cap lifts in **3.10.1**. It cannot go higher either as 3.10.2+ requires Python ≥3.11.

### Not changed here

- `numba>=0.60` and `opencv-python-headless>=4.10.0.84` are in #3508

---

**Note:** If we want to narrow this as much as possible, we could limit to tables and scipy as they are the ones with numpy upper bounds.


deruyter92 (commented):
As a product decision (force a more modern stack) I see a good value of this PR. 

But to me this sounds entirely separate from the discussion on numpy in the associated PRs. The goal in those PRs is to _support_ numpy 2.0 for broad compatibility, not to _make sure_ that users don't end up with numpy 1.x. 

To me the benefits from enforcing a modern stack right now do not feel like they outweigh the restrictive downsides: if our users work with old other scientific software, it might be helpful to keep the lower-bounds low for flexibility, _unless we really can't support_. 

maybe we should add a test lane first with pip `resolution=lowest-direct` to test if our codebase contains unsupported usage. In that case we _need_ to update the lower-bounds. If not necessary, I wouldn't maybe for now. 

@C-Achard  do you agree?

C-Achard:
@deruyter92 Yes totally! Sorry if it sounds like it's tightly coupled to the other PRs, it isn't.

